PR #2, opened 2018-12-18
PSA tests: use a few common test macros

gilles-peskine-arm:
Declare a few macros in `helpers.function` and use them in the PSA tests. I haven't touched non-PSA tests.

* `TEST_EQUAL` (slightly more readable than `TEST_ASSERT(a==b)`, but the real payoff will come later when we instrument it to print the values on failure)
* `PSA_ASSERT` (`TEST_ASSERT(…==PSA_SUCCESS)`)
* `ARRAY_LENGTH` (with protection against usage on a pointer when compiling with GCC)
* `MIN`, `MAX`

There's some refactoring going on, so I recommend reviewing commit by commit.

### DIFF
--- a/tests/suites/helpers.function
+++ b/tests/suites/helpers.function
@@ -90,13 +90,23 @@ typedef struct data_tag
         }                                           \
     } while( 0 )
 
+/** Evaluate two expressions and fail the test case if they have different
+ * values.
+ *
+ * \param expr1     An expression to evaluate.
+ * \param expr2     The expected value of \p expr1. This can be any
+ *                  expression, but it is typically a constant.
+ */
+#define TEST_EQUAL( expr1, expr2 )              \
+    TEST_ASSERT( ( expr1 ) == ( expr2 ) )
+
 /** Evaluate an expression and fail the test case if it returns an error.
  *
  * \param expr      The expression to evaluate. This is typically a call
  *                  to a \c psa_xxx function that returns a value of type
  *                  #psa_status_t.
  */
-#define PSA_ASSERT( expr ) TEST_ASSERT( ( expr ) == PSA_SUCCESS )
+#define PSA_ASSERT( expr ) TEST_EQUAL( ( expr ), PSA_SUCCESS )
 
 /** Allocate memory dynamically and fail the test case if this fails.
  *
--- a/tests/suites/helpers.function
+++ b/tests/suites/helpers.function
@@ -202,6 +202,23 @@ typedef struct data_tag
     ( STATIC_ASSERT_THEN_RETURN( IS_ARRAY_NOT_POINTER( array ),         \
                                  ARRAY_LENGTH_UNSAFE( array ) ) )
 
+/** Return the smaller of two values.
+ *
+ * \param x         An integer-valued expression without side effects.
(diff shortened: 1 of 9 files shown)


tests/suites/helpers.function
```diff
@@ -150,9 +150,41 @@ typedef struct data_tag
...
+    ( ! __builtin_types_compatible_p( __typeof__( arg ),                \
+                                      __typeof__( &( arg )[0] ) ) )
+#else
+/* On platforms where we don't know how to implement this check,
+ * omit it. Oh well, a non-portable check is better than nothing. */
+#define STATIC_ASSERT_EXPR_IF_POSSIBLE( const_expr ) ( (void) 0 )
+#endif
+
+/* A compile-time constant with the value 0. If `const_expr` is not a
+   compile-time constant with a nonzero value, cause a compile-time error. */
```
Patater: Style: begin continued comment lines with a `*`.

tests/suites/helpers.function
```diff
@@ -150,9 +150,41 @@ typedef struct data_tag
...
+#if defined(__GNUC__)
+/* Test if arg and &(arg)[0] have the same type. This is true if arg is
+ * an array but not if it's a pointer. */
+#define IS_ARRAY_NOT_POINTER( arg )                                     \
+    ( ! __builtin_types_compatible_p( __typeof__( arg ),                \
+                                      __typeof__( &( arg )[0] ) ) )
+#else
+/* On platforms where we don't know how to implement this check,
+ * omit it. Oh well, a non-portable check is better than nothing. */
+#define STATIC_ASSERT_EXPR_IF_POSSIBLE( const_expr ) ( (void) 0 )
```
Patater: Where is `IS_ARRAY_NOT_POINTER` defined for non GNUC platforms? Should `STATIC_ASSERT_EXPR_IF_POSSIBLE` instead be `IS_ARRAY_NOT_POINTER`? Where is `STATIC_ASSERT_EXPR_IF_POSSIBLE` used?

gilles-peskine-arm: Leftover from a slightly different design that I abandoned without committing. I have amended the corresponding commit.

tests/suites/helpers.function
gilles-peskine-arm: Fixed in the amended commit.

gilles-peskine-arm:
I've amended the commit "Add a safety check to ARRAY_LENGTH" to address the issues that @Patater pointed out. Previous history at https://github.com/gilles-peskine-arm/mbed-crypto/tree/psa-test_macros-1

Patater:
CI failure is test_suite_timing on FreeBSD, a known issue.

Patater:
Other half of the release job: https://jenkins-internal.mbed.com/job/mbedtls-psa-release-new/202/

Patater:
Basic build and test passed. Mbed OS testing went badly because the test is broke: not this PR's fault.

gilles-peskine-arm:
And I also fixed the `\file` statement in `crypto_accel_driver.h` so `doxygen.sh` now passes.

Patater:
The test failure is the FreeBSD timing test, a known issue.

PR #3, opened 2018-12-21
Fix maybe-uninitialized warning

gilles-peskine-arm:
GCC 4.8 warns that some variables may be used without having been initialized. They aren't, but determining that takes nontrivial analysis, so initialize them at the point of definition.


Patater:

PR #7, opened 2019-01-05
[mbedcrypto] all.sh: make it possible to run a subset of the components

gilles-peskine-arm:
This is a side-port of https://github.com/ARMmbed/mbedtls/pull/2325 plus https://github.com/ARMmbed/mbedtls/pull/2332 to `mbed-crypto/development`.

Because this PR contains commits that are very tedious to replay and to review, I merged the previously-reviewed PR against mbedtls-2.14.0 rather than rebasing it on top of the current branch. The commits before the merge were previously reviewed in https://github.com/ARMmbed/mbedtls/pull/2235, therefore this PR needs a review of:

* The merge of `mbed-crypto/development` into my working branch.
* The mostly-uneventful cherry-picking of all the subsequent commits from https://github.com/ARMmbed/mbedtls/pull/2332. The only conflict was the removal of `run_all_components` where a slightly different list of components got deleted.


gilles-peskine-arm:
Run of `all.sh` on CI: https://jenkins-internal.mbed.com/job/mbedtls-psa-release-new/220/

gilles-peskine-arm:
New run of `all.sh`: https://jenkins-internal.mbed.com/job/mbedtls-psa-release-new/226/

Patater:
Should we instead merge mbedtls/development after your similar change goes in to that branch? I'd prefer to avoid unecessary merge conflicts when we merge mbedtls/development into mbed-crypto/development, and that way seems more likely to avoid conflicts than to merge this and that and then later merge in mbedtls/development.

gilles-peskine-arm:
Yes, we can merge `mbedtls/development` instead, if you prefer the strategy of #12 with a giant merge commit. I would prefer to avoid hairy merges and minimize conflicts between mbedcrypto and mbedtls, but I think it's too late for that.

Patater:
After #12, your PR to mbedtls/development should apply here as well, at least more easily, but would still need mbed-crypto-specific tweaking still for any all.sh tests we've added (or removed). At least the tooling updates would be single-sourced and we'd not have to resolve conflicts between two patch sets of tooling updates.

Patater:
#12 is too hairy to merge. So, it's not likely to be ready sooner than this PR. We'll deal with the conflicts and updating to the latest Mbed TLS crypto later, and enjoy a better all.sh today.

gilles-peskine-arm:
I thought there wouldn't be any merge conflicts on `all.sh` since this PR and the one on 2.16 make identical changes to the parts that were identical on both sides. I just tried a `git merge` and there is a merge conflict in `pre_parse_command_line` even though the two branches have identical content at that point. I think it's because so much has changed in the file (because most lines got reindented) that git is unable to recognize what the common ancestor of that part of the file is and so is trying to plug the pieces back at an offset. But for a human it's easy to analyze: the two versions of `pre_parse_command_line` are byte-for-byte identical, we meant them to be identical, so pick one and move on.

Patater:
Test failure is `test_suite_timing` on FreeBSD, a known issue.

PR #8, opened 2019-01-08
modify check-names.sh and list-macros.sh to work with PSA constants

NirSonnenschein:
fixed processing of PSA macros in check names script.
This required changes in:
    list-macros.sh to scan the PSA headers
    check-names to scan PSA files and allow PSA_* macro names

PR #9, opened 2019-01-08
Make number of key slots externally configurable

NirSonnenschein:
changes to make number of key slots externally configurable

changed define name from PSA_KEY_SLOT_COUNT to MBEDTLS_PSA_KEY_SLOT_COUNT
moved define to crypto_extra.h and made it possible to externally override

depends on fixes from (and is rebased on) : https://github.com/ARMmbed/mbed-crypto/pull/8

### DIFF
--- a/include/psa/crypto_extra.h
+++ b/include/psa/crypto_extra.h
@@ -122,7 +122,9 @@ psa_status_t mbedtls_psa_inject_entropy(const unsigned char *seed,
 
 
 /* The Number of key slots (plus one because 0 is not used).
- * The value is a compile-time constant for now, for simplicity. */
+ * The value is a compile-time constant for now, for simplicity.
+ * Note: this configuration flag is an implementation detail that may change
+ * or be removed in future versions */
 #if !defined(MBEDTLS_PSA_KEY_SLOT_COUNT)
     #define MBEDTLS_PSA_KEY_SLOT_COUNT 32
 #endif /* !defined(MBEDTLS_PSA_KEY_SLOT_COUNT) */
--- a/include/psa/crypto_extra.h
+++ b/include/psa/crypto_extra.h
@@ -123,10 +123,9 @@ psa_status_t mbedtls_psa_inject_entropy(const unsigned char *seed,
 
 /* The Number of key slots (plus one because 0 is not used).
  * The value is a compile-time constant for now, for simplicity. */
-#if defined(MBEDTLS_PSA_KEY_SLOT_COUNT)
-#else
+#if !defined(MBEDTLS_PSA_KEY_SLOT_COUNT)
     #define MBEDTLS_PSA_KEY_SLOT_COUNT 32
-#endif /*MBEDTLS_PSA_KEY_SLOT_COUNT*/
+#endif /* !defined(MBEDTLS_PSA_KEY_SLOT_COUNT) */
 
 
 #ifdef __cplusplus
--- a/include/psa/crypto_extra.h
+++ b/include/psa/crypto_extra.h
@@ -121,6 +121,14 @@ psa_status_t mbedtls_psa_inject_entropy(const unsigned char *seed,
                                         size_t seed_size);
 
 
+/* The Number of key slots (plus one because 0 is not used).
+ * The value is a compile-time constant for now, for simplicity. */
+#if defined(MBEDTLS_PSA_KEY_SLOT_COUNT)
+#else
+    #define MBEDTLS_PSA_KEY_SLOT_COUNT 32
+#endif /*MBEDTLS_PSA_KEY_SLOT_COUNT*/
+
+
 #ifdef __cplusplus
 }
 #endif
--- a/library/psa_crypto_slot_management.c
+++ b/library/psa_crypto_slot_management.c
@@ -46,7 +46,7 @@
 
 typedef struct
 {
-    psa_key_slot_t key_slots[PSA_KEY_SLOT_COUNT];
+    psa_key_slot_t key_slots[MBEDTLS_PSA_KEY_SLOT_COUNT];
     unsigned key_slots_initialized : 1;
 } psa_global_data_t;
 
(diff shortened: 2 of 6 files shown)


include/psa/crypto_extra.h
```diff
@@ -121,6 +121,14 @@ psa_status_t mbedtls_psa_inject_entropy(const unsigned char *seed,
                                         size_t seed_size);
 
 
+/* The Number of key slots (plus one because 0 is not used).
+ * The value is a compile-time constant for now, for simplicity. */
+#if defined(MBEDTLS_PSA_KEY_SLOT_COUNT)
```
avolinski:
```suggestion
#if !defined(MBEDTLS_PSA_KEY_SLOT_COUNT)
```

include/psa/crypto_extra.h
```diff
@@ -121,6 +121,14 @@ psa_status_t mbedtls_psa_inject_entropy(const unsigned char *seed,
...
+#else
```
avolinski: not needed if above accepted

include/psa/crypto_extra.h
```diff
@@ -121,6 +121,14 @@ psa_status_t mbedtls_psa_inject_entropy(const unsigned char *seed,
...
+    #define MBEDTLS_PSA_KEY_SLOT_COUNT 32
```
avolinski: i understand this is the naming convention for external define.
but still mbedtls has nothing to do with psa key slots number.
can we not have the mbedtls prefix? have some other?
why not leave it PSA_KEY_SLOT_COUNT?

NirSonnenschein: I asked @gilles-peskine-arm and @Patater  on their inputs on this and this is the name @gilles-peskine-arm requested.

include/psa/crypto_extra.h
NirSonnenschein: I was told the preffernce was for positive ifdef checks as opposed to negative ones.
I agree that the negative route is shorter. I don't mind changing if this doesn't present a problem.

include/psa/crypto_extra.h
NirSonnenschein: see above

include/psa/crypto_extra.h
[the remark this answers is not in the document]
NirSonnenschein: Hi @gilles-peskine-arm, 
regarding adding this flag to config.h, you wrote in the email chain: 
"I'm not keen on adding it to mbedtls/config.h because it isn't necessarily something that we'll maintain backward compatibility with. "

This is the reason it was left out, I can add it now if you prefer

include/psa/crypto_extra.h
NirSonnenschein: great, thanks @gilles-peskine-arm,  will fix this to the simpler version

include/psa/crypto_extra.h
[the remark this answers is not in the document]
gilles-peskine-arm: Hmm, yes, that is a good point, we aren't going to promise backward compatibility for this, so it shouldn't be in `config.h`, or at least not until we revise (and document!) our backward compatibility rules. So please keep it in `crypto_extra.h`, and add a comment that states that this is an implementation detail that may change or be removed in future versions. By default we _do_ promise to maintain backward compatibility with things defined in `crypto_extra.h`.

include/psa/crypto_extra.h
NirSonnenschein: fixed

include/psa/crypto_extra.h
NirSonnenschein: fixed

include/psa/crypto_extra.h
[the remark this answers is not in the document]
NirSonnenschein: comment added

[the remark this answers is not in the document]
gilles-peskine-arm: We do have precedents for defining options in `config.h` that we don't intend to preserve backward compatibility for: `MBEDTLS_PSA_HAS_ITS_IO`, `MBEDTLS_PSA_CRYPTO_SPM`, the PSA storage modules. We intend to preserve the functionality, but not necessarily the configuration interface.

@Patater What's your opinion on this?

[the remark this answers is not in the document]
gilles-peskine-arm: More generally, in Mbed Crypto, we don't intend to preserve compatibility with `MBEDTLS_xxx` APIs in the long term. So I think my view of backward compatibility is behind the times and the configuration option should go in `config.h`.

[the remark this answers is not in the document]
NirSonnenschein: @gilles-peskine-arm and @Patater , what is the final ruling on this? (add to config.h or not)

avolinski: slot count is public define now, so it can be PSA_xxx ...

include/psa/crypto_extra.h
```diff
@@ -121,6 +121,15 @@ psa_status_t mbedtls_psa_inject_entropy(const unsigned char *seed,
                                         size_t seed_size);
 
 
+/* The Number of key slots (plus one because 0 is not used).
+ * The value is a compile-time constant for now, for simplicity.
+ * Note: this configuration flag is an implementation detail that may change
+ * or be removed in future versions */
+#if !defined(MBEDTLS_PSA_KEY_SLOT_COUNT)
+    #define MBEDTLS_PSA_KEY_SLOT_COUNT 32
```
avolinski: how about default number of 16 or 20?
cc @gilles-peskine-arm

include/psa/crypto_extra.h
NirSonnenschein: I think @gilles-peskine-arm said that PSA_XXX is only for public PSA API and this is not as it is a configuration for MbedCrypto that other implementations do not currently have to support.

include/psa/crypto_extra.h
NirSonnenschein: as was mentioned in the email discussion, we need to see some testing on real world use (e.g. mbed client usage) to decide on a new default value. So for now I've left the default as it is. If we can decide on a new value I will set it.

Patater:
Closing as stale. The check-names updates have been merged already. The benefit of moving the number of slots configuration option from one impelentation-specifc header to another is of questionable benefit. Better to move into config.h if we want this user configurable.

PR #12, opened 2019-01-10
Update to Mbed TLS 2.16.0

Patater:
This brings our crypto implementation up to date with the latest Mbed TLS crypto. This primarily brings us the parameter validation feature and new ECDH alternative implementation interface, but also other security and bug fixes.

All code has been reviewed already, except the merge commit, which is pretty hairy. Please review the merge commit (probably best done with command line tools, not GitHub at least.)


gilles-peskine-arm:
The merge conflict required a series of nontrivial design decisions, such as all the places where the objectives of “parameter validation” conflict with changes made for the PSA crypto API or for TLS-over-PSA. I would strongly prefer to have separate commits on one or both sides to make these design changes, and have a merge conflict which only takes obvious decisions.

Patater:
I don't like giant merges  any more than the next guy. I would suggest we break this up instead, using multiple merges to get us to 2.16.0. As such, I'm closing this PR.

Regarding the review feedback:
- Changes to constant time code were unintentional. The intention was to take Mbed TLS's changes as-is.
- I'm not sure where you think rsa.h is wrong. The intention was to make rsa.c OK with NULL buffers of length zero and the parameter checks were modified to enable that. The documentation in rsa.h was updated back to before https://github.com/ARMmbed/mbed-crypto/commit/2f660d047dd60ac68ebe6c8d4da0d141403b6d71 in order to document this properly. I could have made a mistake, of course.

PR #14, opened 2019-01-14
Simplify or specify public key format for DH and DSA

Patater:
This PR builds on top of #13 (starting with 63455af9dcf7 "psa: Add get/set domain parameters") to simplify the DSA key format and to specify the DH key format.

TODO:
- Determine what we want the API to look like for setting domain parameters and update the PR to reflect that choice.
- Rebase on top of #13 after that's merged.


include/psa/crypto.h
```diff
@@ -369,6 +369,70 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
...
+ * \brief Set domain parameters for a key.
+ *
+ * Some key types require additional domain parameters to be set before import
+ * or generation of the key. The domain parameters can be set with this
+ * function or, for key generation, through the \c extra parameter of
+ * psa_generate_key().
+ *
+ * The format for the required domain parameters varies by the key type.
+ *
+ * \param[in] data    Buffer containing the key domain parameters. The content
```
dgreen-arm: Documentation for handle parameter is missing.

include/psa/crypto.h
```diff
@@ -369,6 +369,70 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
...
+                                           size_t data_length);
+
+/**
+ * \brief Get domain parameters for a key.
+ *
+ * Get the domain parameters for a key with this function, if any. The format
+ * of the domain parameters written to \p data is specified in the
+ * documentation for psa_set_key_domain_parameters().
+ *
+ * \param[out] data         On success, the key domain parameters.
```
dgreen-arm: Documentation for handle parameter is missing.

Patater:
Rebased to add Doxygen documentation for handle parameters.

include/psa/crypto.h
```diff
@@ -428,6 +509,10 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
  *   and `PSA_ECC_CURVE_BRAINPOOL_PXXX`).
  *   This is the content of the `privateKey` field of the `ECPrivateKey`
  *   format defined by RFC 5915.
+ * - For Diffie-Hellman key exchange key pairs (#PSA_KEY_TYPE_DH_KEYPAIR), the
```
gilles-peskine-arm: “key pairs” → “private keys” for consistency

include/psa/crypto.h
```diff
@@ -369,6 +369,96 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
...
+ * \brief Set domain parameters for a key.
+ *
+ * Some key types require additional domain parameters to be set before import
+ * or generation of the key. The domain parameters can be set with this
+ * function or, for key generation, through the \c extra parameter of
+ * psa_generate_key().
+ *
+ * The format for the required domain parameters varies by the key type.
+ * - For DSA public keys (#PSA_KEY_TYPE_DSA_PUBLIC_KEY),
+ *   the `Dss-Parms` format as defined by RFC 3279 &sect;2.3.2.
```
gilles-peskine-arm: “the DER representation of the `Dss-Parms` format”

include/psa/crypto.h
```diff
@@ -369,6 +369,96 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
...
+ *   ```
+ *   Dss-Parms ::= SEQUENCE  {
+ *      p       INTEGER,
+ *      q       INTEGER,
+ *      g       INTEGER
+ *   }
+ *   ```
+ * - For Diffie-Hellman key exchange keys (#PSA_KEY_TYPE_DH_PUBLIC_KEY), the
+ *   `DomainParameters` format as defined by RFC 3279 &sect;2.3.3.
```
gilles-peskine-arm: “the DER representation of the `DomainParameters` format”

include/psa/crypto.h
```diff
@@ -369,6 +369,96 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
...
+ *   ```
+ *   DomainParameters ::= SEQUENCE {
+ *      p               INTEGER,                    -- odd prime, p=jq +1
+ *      g               INTEGER,                    -- generator, g
+ *      q               INTEGER,                    -- factor of p-1
+ *      j               INTEGER OPTIONAL,           -- subgroup factor
+ *      validationParms ValidationParms OPTIONAL
```
gilles-peskine-arm: We try to avoid optional stuff in PSA. Do we allow, mandate or reject `j` and `validationParms`? If it's allowed, do we guarantee that the implementation will check it?

My gut feeling:
* Allow them because it's convenient to accept whatever has been package, and it's very easy to skip.
* Allow an implementation to ignore them altogether, because verifying them can be costly and the API is meant to be suitable for microcontrollers.

include/psa/crypto.h
```diff
@@ -369,6 +369,96 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
...
+ *   }
+ *   ValidationParms ::= SEQUENCE {
+ *      seed            BIT STRING,
+ *      pgenCounter     INTEGER
+ *   }
+ *   ```
+ *
+ * \param handle      Handle to the key to set domain parameters for.
+ * \param[in] data    Buffer containing the key domain parameters. The content
+ *                    of this buffer is interpreted according to \p type. of
```
gilles-peskine-arm: Missing `\p type` parameter in the documentation and in the prototype.

gilles-peskine-arm: Something's missing before “of `psa_export_key()` or …”

include/psa/crypto.h
```diff
@@ -369,6 +369,96 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
...
+ * \retval #PSA_SUCCESS
+ * \retval #PSA_ERROR_INVALID_HANDLE
+ * \retval #PSA_ERROR_OCCUPIED_SLOT
+ *         There is already a key in the specified slot.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_HARDWARE_FAILURE
+ * \retval #PSA_ERROR_TAMPERING_DETECTED
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
```
gilles-peskine-arm: Not relevant for this function (you won't get a handle in the first place)

include/psa/crypto.h
```diff
@@ -369,6 +369,96 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
...
+ * \param handle            Handle to the key to get domain parameters from.
+ * \param[out] data         On success, the key domain parameters.
+ * \param data_size         Size of the \p data buffer in bytes.
+ * \param[out] data_length  On success, the number of bytes
+ *                          that make up the key domain parameters data.
+ *
+ * \retval #PSA_SUCCESS
+ * \retval #PSA_ERROR_INVALID_HANDLE
+ * \retval #PSA_ERROR_EMPTY_SLOT
+ *         There is no key in the specified slot.
```
gilles-peskine-arm: That's not an error: you can read domain parameters that have been set even before importing a key. The possible errors here are that no domain parameters have been set, or that the slot contains a key which does not have any domain parameters.

include/psa/crypto.h
```diff
@@ -369,6 +369,96 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
...
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ * \retval #PSA_ERROR_NOT_SUPPORTED
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_HARDWARE_FAILURE
+ * \retval #PSA_ERROR_TAMPERING_DETECTED
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
```
gilles-peskine-arm: Not relevant for this function (you won't get a handle in the first place)

include/psa/crypto.h
```diff
@@ -369,6 +369,96 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
...
+ * \param[out] data         On success, the key domain parameters.
+ * \param data_size         Size of the \p data buffer in bytes.
+ * \param[out] data_length  On success, the number of bytes
+ *                          that make up the key domain parameters data.
+ *
+ * \retval #PSA_SUCCESS
+ * \retval #PSA_ERROR_INVALID_HANDLE
+ * \retval #PSA_ERROR_EMPTY_SLOT
+ *         There is no key in the specified slot.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
```
gilles-peskine-arm: \+ `NOT_SUPPORTED` if the implementation doesn't have code to format domain parameters

include/psa/crypto.h
```diff
@@ -2253,6 +2319,18 @@ typedef struct {
  *                            specifying the public exponent. The
  *                            default public exponent used when \p extra
  *                            is \c NULL is 65537.
+ *                          - For an DSA key (\p type is
+ *                            #PSA_KEY_TYPE_DSA_KEYPAIR), \p extra is an
+ *                            optional structure specifying the key domain
```
gilles-peskine-arm: It isn't a structure, it's a buffer that will be parsed.

gilles-peskine-arm: I find this description unclear: it looks like you can pass `extra`, use `set_key_domain_parameters`, or do neither. In fact you need to do at least one, and the last one applies.

> For a DSA key (`\p type` is `#PSA_KEY_TYPE_DSA_KEYPAIR`) or a finite-field Diffie-Hellman key (`p type` is `#PSA_KEY_TYPE_DH_KEYPAIR`), `extra` is a pointer to a buffer containing domain parameters in DER representation. The format of `buffer` is the same as the input to `psa_set_key_domain_parameters()`. If the application has previously called `psa_set_key_domain_parameters()` on `\p handle`, it may pass an empty buffer (`` `extra_length = 0` ``), in which case the key generation will use the previously set domain parameters.

include/psa/crypto.h
```diff
@@ -2253,6 +2319,18 @@ typedef struct {
...
+ *                            parameters. The key domain parameters can also be
+ *                            provided by psa_set_key_domain_parameters(),
+ *                            which documents the format of the structure.
+ *                          - For a DH key (\p type is
```
gilles-peskine-arm: The logic is exactly the same for DSA and DH, so you can combine their documentation.

include/psa/crypto.h
```diff
@@ -369,6 +369,96 @@ psa_status_t psa_get_key_information(psa_key_handle_t handle,
...
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ * \retval #PSA_ERROR_NOT_SUPPORTED
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_HARDWARE_FAILURE
+ * \retval #PSA_ERROR_TAMPERING_DETECTED
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
```
gilles-peskine-arm: Needs a macro in `crypto_sizes.h` to determine the buffer size from the key type and bit-size.

include/psa/crypto.h
yanesca: That sounds very good for usability.

Considering it from the security perspective the checks a specific implementation makes have significant impact and I think that we should be very explicit about it. How about allowing the `validationParams` to be present and allowing the implementations to ignore them, but adding a new function for parameter validation?

By the way can't we just drop custom parameter FFDH from 1.0 and use named groups instead. We could reintroduce custom parameters if and when customers ask for it.
(I know I have asked this before, but I don't remember what was the conclusion.)

I dislike custom parameter FFDH, because it can be backdoored and no matter how hard you validate your parameters, you won't notice:
https://tls.mbed.org/tech-updates/blog/dh-backdoors

(Of course you can argue that named groups have to be bigger because of logjam, but I wouldn't advise to use keys that short with custom parameters either.)

gilles-peskine-arm:
For the beta1 API release, I'm adding a `type` parameter to `psa_set_key_domain_parameters`. https://github.com/gilles-peskine-arm/mbed-crypto/tree/pubkey-format-dh-dsa-pass_type_to_set_domain_parameters

Patater:
Closing as very out of date. Needs to be redone in a PSA Crypto API 1.0b3 way.

PR #16, opened 2019-01-14
Support for slots to handles crypto API changes under SPM

itayzafrir:
Support for the API changes from slots to handles under SPM (code separation between SPE and NSPE). 
Same idea as in https://github.com/ARMmbed/mbed-crypto/blob/development/library/psa_crypto.c#L29-L42

### DIFF
--- a/library/psa_crypto_slot_management.c
+++ b/library/psa_crypto_slot_management.c
@@ -26,6 +26,21 @@
 #endif
 
 #if defined(MBEDTLS_PSA_CRYPTO_C)
+/*
+ * When MBEDTLS_PSA_CRYPTO_SPM is defined, the code is being built for SPM
+ * (Secure Partition Manager) integration which separates the code into two
+ * parts: NSPE (Non-Secure Processing Environment) and SPE (Secure Processing
+ * Environment). When building for the SPE, an additional header file should be
+ * included.
+ */
+#if defined(MBEDTLS_PSA_CRYPTO_SPM)
+/*
+ * PSA_CRYPTO_SECURE means that this file is compiled for the SPE.
+ * Some headers will be affected by this flag.
+ */
+#define PSA_CRYPTO_SECURE 1
+#include "crypto_spe.h"
+#endif
 
 #include "psa/crypto.h"
 
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -27,15 +27,16 @@
 
 #if defined(MBEDTLS_PSA_CRYPTO_C)
 /*
- * In case MBEDTLS_PSA_CRYPTO_SPM is defined the code is built for SPM (Secure
- * Partition Manager) integration which separate the code into two parts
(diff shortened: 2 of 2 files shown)


library/psa_crypto_slot_management.c
```diff
@@ -26,6 +26,20 @@
 #endif
 
 #if defined(MBEDTLS_PSA_CRYPTO_C)
+/*
+ * In case MBEDTLS_PSA_CRYPTO_SPM is defined the code is built for SPM (Secure
+ * Partition Manager) integration which separate the code into two parts
+ * NSPE (Non-Secure Process Environment) and SPE (Secure Process Environment).
+ * In this mode an additional header file should be included.
```
Patater: Could we change this to the following?

```
When MBEDTLS_PSA_CRYPTO_SPM is defined, the code is being build for SPM (Secure Partition Manager) integration which separates the code into two parts: NSPE (Non-Secure Processing Element) and SPE (Secure Processing Element). When building for the SPE, an additional header file should be included.
```

library/psa_crypto_slot_management.c
```diff
@@ -26,6 +26,20 @@
...
+ */
+#if defined(MBEDTLS_PSA_CRYPTO_SPM)
+/*
+ * PSA_CRYPTO_SECURE means that this file is compiled to the SPE side.
```
Patater:
```diff
- compiled to the SPE side.
- compiled for the SPE.
```

library/psa_crypto_slot_management.c
```diff
@@ -26,6 +26,20 @@
...
+ * some headers will be affected by this flag.
```
Patater:
```diff
- some headers will be affected by this flag.
+ Some headers will be affected by this flag.
```

Patater:
If you wouldn't mind, it'd help to have another commit to update the other comment block. Thanks.

library/psa_crypto_slot_management.c
itayzafrir: Changed

library/psa_crypto_slot_management.c
itayzafrir: Changed

library/psa_crypto_slot_management.c
itayzafrir: Changed

itayzafrir:
@Patater  I've updated the the code comments, commit message title and description, I also added a new commit to update existing SPM documentation in the code. 
Please take another look.

library/psa_crypto_slot_management.c
Patater: I see this still says "Environment" instead of "Element".

gilles-peskine-arm: I don't see where “environment” should be changed to “element”. (N)SPE is (Non-)Secure Processing Environment.

PR #17, opened 2019-01-14
Doc only: multipart AEAD, one-shot hash/MAC/cipher, abort operation on destroy_key

gilles-peskine-arm:
This PR combines several pending API changes related to symmetric cryptography. In this PR, I have only modified the spec, I have not implemented the changes. The code still builds but does not conform to the specification.

* When a key is closed or destroy, abort any active multipart operation using that key.
* Add multipart functions for AEAD.
* Add one-shot functions for hash, MAC and cipher.


Patater:
Will update the docs when we implement these features (https://github.com/ARMmbed/mbed-crypto/issues/35, https://github.com/ARMmbed/mbed-crypto/issues/36, and https://github.com/ARMmbed/mbed-crypto/issues/86). These API and documentation changes are already present in the API-only branch.

PR #18, opened 2019-01-19
New function psa_hash_clone

gilles-peskine-arm:
Clone a hash operation.

Test good cases as part as multipart tests. Add new test functions for
the state machine.

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -921,6 +921,24 @@ psa_status_t psa_hash_verify(psa_hash_operation_t *operation,
  */
 psa_status_t psa_hash_abort(psa_hash_operation_t *operation);
 
+/** Clone a hash operation.
+ *
+ * \param[in] source_operation      The active hash operation to clone.
+ * \param[in,out] target_operation  The operation object to set up.
+ *                                  It must be initialized but not active.
+ *
+ * \retval #PSA_SUCCESS
+ * \retval #PSA_ERROR_BAD_STATE
+ *         \p source_operation is not an active hash operation.
+ * \retval #PSA_ERROR_BAD_STATE
+ *         \p source_operation is active.
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_HARDWARE_FAILURE
+ * \retval #PSA_ERROR_TAMPERING_DETECTED
+ */
+psa_status_t psa_hash_clone(const psa_hash_operation_t *source_operation,
+                            psa_hash_operation_t *target_operation);
+
 /**@}*/
 
 /** \defgroup MAC Message authentication codes
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -923,6 +923,15 @@ psa_status_t psa_hash_abort(psa_hash_operation_t *operation);
 
 /** Clone a hash operation.
  *
+ * This function copies the state of an ongoing hash operation to
+ * a new operation object. In other words, this function is equivalent
+ * to calling psa_hash_setup() on \p target_operation with the same
+ * algorithm that \p source_operation was set up for, then
+ * psa_hash_update() on \p target_operation with the same input that
+ * that was passed to \p source_operation. After this function returns, the
+ * two objects are independent, i.e. subsequent calls involving one of
+ * the objects do not affect the other object.
+ *
  * \param[in] source_operation      The active hash operation to clone.
  * \param[in,out] target_operation  The operation object to set up.
  *                                  It must be initialized but not active.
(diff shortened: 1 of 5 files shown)


include/psa/crypto.h
```diff
@@ -921,6 +921,24 @@ psa_status_t psa_hash_verify(psa_hash_operation_t *operation,
...
+ *
+ * \param[in] source_operation      The active hash operation to clone.
+ * \param[in,out] target_operation  The operation object to set up.
+ *                                  It must be initialized but not active.
+ *
+ * \retval #PSA_SUCCESS
+ * \retval #PSA_ERROR_BAD_STATE
+ *         \p source_operation is not an active hash operation.
+ * \retval #PSA_ERROR_BAD_STATE
+ *         \p source_operation is active.
```
hanno-arm: These two return value documentations seem to contradict each other.

hanno-arm: Do you mean `\p target_operation is active` in the second `\retval`?

include/psa/crypto.h
```diff
@@ -921,6 +921,24 @@ psa_status_t psa_hash_verify(psa_hash_operation_t *operation,
  */
 psa_status_t psa_hash_abort(psa_hash_operation_t *operation);
 
+/** Clone a hash operation.
+ *
+ * \param[in] source_operation      The active hash operation to clone.
```
hanno-arm: It should be documented that this should be initialized and active.

library/psa_crypto.c
```diff
@@ -1421,6 +1421,67 @@ psa_status_t psa_hash_verify( psa_hash_operation_t *operation,
     return( PSA_SUCCESS );
 }
 
+psa_status_t psa_hash_clone(const psa_hash_operation_t *source_operation,
```
hanno-arm: Usually, the context acted upon is listed as the first argument; I'd therefore prefer the destination context to come first.

include/psa/crypto.h
```diff
@@ -921,6 +921,24 @@ psa_status_t psa_hash_verify(psa_hash_operation_t *operation,
  */
 psa_status_t psa_hash_abort(psa_hash_operation_t *operation);
 
+/** Clone a hash operation.
```
hanno-arm: Everything else wouldn't make sense, but we might still mention here that the target context is active and in the same internal state as the source after a successful call to this function.

include/psa/crypto.h
gilles-peskine-arm: Isn't the current wording sufficent? Active implies initialized.

library/psa_crypto.c
gilles-peskine-arm: There's no such convention in this API. There's a convention that input comes before output, and a convention that the context comes before other parameters. With an input context and an output context, the input context comes first.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -1924,6 +1924,92 @@ exit:
...
+    PSA_ASSERT( psa_hash_setup( &op_aborted, alg ) );
+    PSA_ASSERT( psa_hash_abort( &op_aborted ) );
+
+    TEST_EQUAL( psa_hash_clone( &op_source, &op_setup ),
+                PSA_ERROR_BAD_STATE );
+
+    PSA_ASSERT( psa_hash_clone( &op_source, &op_init ) );
+    PSA_ASSERT( psa_hash_finish( &op_init,
+                                 hash, sizeof( hash ), &hash_len ) );
+    PSA_ASSERT( psa_hash_clone( &op_source, &op_finished ) );
```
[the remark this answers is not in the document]
gilles-peskine-arm: This isn't critical to understanding the current behavior, and other types of operation should use consistent terminology, so I'll do that in a later writing pass. Tracked internally at 
https://github.com/ARMmbed/psa-crypto/issues/132

library/psa_crypto.c
hanno-arm: Ok.

include/psa/crypto.h
hanno-arm: That was an oversight on my side, because I was expecting the wording to be "The XXX. Must be YYY." instead of "The YYY XXX". I'd prefer the former, but that's not a blocker.

tests/suites/test_suite_psa_crypto.function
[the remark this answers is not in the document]
hanno-arm: Agreed.

include/psa/crypto.h
gilles-peskine-arm: I added a paragraph of English text that defines the behavior of the function.

include/psa/crypto.h
gilles-peskine-arm: Yes. I fixed the copypasta.

library/psa_crypto.c
```diff
@@ -1421,6 +1421,67 @@ psa_status_t psa_hash_verify( psa_hash_operation_t *operation,
...
+                            psa_hash_operation_t *target_operation)
```
yanesca: Style: spaces before and after the parentheses.

yanesca:
Found a minor style issue, but other than that it looks good to me!

Patater:
CI was flaky and failed `DTLS proxy: 3d, gnutls server, fragmentation `, a known issue.

PR #19, opened 2019-01-19
Don't require a type and size when creating a key slot

gilles-peskine-arm:
Remove the type and bits arguments to psa_allocate_key() and
psa_create_key(). They can be useful if the implementation wants to
know exactly how much space to allocate for the slot, but many
implementations (including ours) don't care, and it's possible to work
around their lack by deferring size-dependent actions to the time when
the key material is created. They are a burden to applications and
make the API more complex, and the benefits aren't worth it.

Change the API and adapt the implementation, the units test and the
sample code accordingly.

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -124,14 +124,6 @@ psa_status_t psa_get_key_lifetime(psa_key_handle_t handle,
  * application calls psa_close_key() or psa_destroy_key() or until the
  * application terminates.
  *
- * This function takes a key type and maximum size as arguments so that
- * the implementation can reserve a corresponding amount of memory.
- * Implementations are not required to enforce this limit: if the application
- * later tries to create a larger key or a key of a different type, it
- * is implementation-defined whether this may succeed.
- *
- * \param type          The type of key that the slot will contain.
- * \param max_bits      The maximum key size that the slot will contain.
  * \param[out] handle   On success, a handle to a volatile key slot.
  *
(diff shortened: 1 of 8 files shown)


tests/suites/test_suite_psa_crypto_slot_management.function
```diff
@@ -350,16 +341,14 @@ void many_transient_handles( int max_handles_arg )
     psa_key_policy_t policy = PSA_KEY_POLICY_INIT;
     uint8_t exported[sizeof( size_t )];
     size_t exported_length;
-    size_t max_bits = PSA_BITS_TO_BYTES( sizeof( exported ) );
```
hanno-arm: This was a typo, but isn't present anywhere else in the test suite.

include/psa/crypto.h
```diff
@@ -124,14 +124,6 @@ psa_status_t psa_get_key_lifetime(psa_key_handle_t handle,
  * application calls psa_close_key() or psa_destroy_key() or until the
  * application terminates.
```
yanesca: How can we detect that the application has terminated and we can free the allocated key slot?

gilles-peskine-arm: That depends on the integration. For a library integration, when the application terminates, this frees all the runtime resources held by the crypto library. For a service integration, the access control layer in the service knows who each handle belongs to and should close a handle if it loses contact with the owner of that handle.

PR #28, opened 2019-01-25
Adapt ECDSA wrapper to new EC public key format

hanno-arm:
__Summary:__ Previously, PSA used `SubjectPublicKeyInfo` structures to serialize EC public keys. This has been changed in https://github.com/ARMmbed/mbed-crypto/pull/13 to using `ECPoint` structures instead, but the wrapper making PSA ECDSA verification available through Mbed TLS' PK API hasn't yet been adapted accordingly - which is what this PR does.

Luckily, Mbed TLS' PK API offers two functions `mbedtls_pk_write_pubkey()` and `mbedtls_pk_write_pubkey_der()`, the latter exporting a `SubjectPublicKeyInfo` structure and the former exporting an `ECPoint` structure in case of EC public keys. For the adaptation of the ECDSA wrapper `ecdsa_verify_wrap()` it is therefore sufficient to use `mbedtls_pk_write_pubkey()` instead of `mbedtls_pk_write_pubkey_der()`.

This PR is the sibling of https://github.com/ARMmbed/mbedtls/pull/2382 to the Mbed TLS repository.

### DIFF
--- a/library/pk_wrap.c
+++ b/library/pk_wrap.c
@@ -553,7 +553,7 @@ static int ecdsa_verify_wrap( void *ctx, mbedtls_md_type_t md_alg,
     int key_len;
     /* see ECP_PUB_DER_MAX_BYTES in pkwrite.c */
     unsigned char buf[30 + 2 * MBEDTLS_ECP_MAX_BYTES];
-    unsigned char *p = (unsigned char*) sig;
+    unsigned char *p;
     mbedtls_pk_info_t pk_info = mbedtls_eckey_info;
     psa_algorithm_t psa_sig_md, psa_md;
     psa_ecc_curve_t curve = mbedtls_psa_translate_ecc_group(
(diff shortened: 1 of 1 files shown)


hanno-arm:
Sibling ARMmbed/mbedtls#2382 passes `ssl-opt.sh` (apart from known 3d interop issue with GnuTLS) and `compat.sh`.

mpg:
Note: the only travis failures are in a known-fragile DTLS 3d interop test.

library/pk_wrap.c
```diff
@@ -567,7 +567,8 @@ static int ecdsa_verify_wrap( void *ctx, mbedtls_md_type_t md_alg,
      * re-construct one to make it happy */
     key.pk_info = &pk_info;
     key.pk_ctx = ctx;
-    key_len = mbedtls_pk_write_pubkey_der( &key, buf, sizeof( buf ) );
+    p = buf + sizeof( buf );
+    key_len = mbedtls_pk_write_pubkey( &p, buf, &key );
```
mpg: Minor: the above comment could be updated to reference `mbedtls_pk_write_pubkey()` without the `_der`.

hanno-arm:
@AndrzejKurek @mpg @Patater Fixed, please re-review.

library/pk_wrap.c
hanno-arm: Fixed.

mpg:
CI only failed in known-intermittent timing test on BSD, so is as good as a pass.

PR #29, opened 2019-01-28
Repo split sibling: Fix oid backwards deps

mpg:
As a preparation for deduplicating files, fix backwards include and config dependencies in the OID module.

This is a sibling PR to https://github.com/ARMmbed/mbedtls/pull/2380

Note: currently this is based on mbedcrypto-0.1.0-beta + ITS fix, in order to avoid pulling unrelated changes on the TLS side when referencing the top commit of this branch. 

**Note:** labeling as "do not merge" for now, as repo split has been postponed for a few weeks.


library/oid.c
```diff
@@ -152,7 +148,6 @@ int FN_NAME( ATTR1_TYPE ATTR1, ATTR2_TYPE ATTR2, const char **oid ,         \
     return( MBEDTLS_ERR_OID_NOT_FOUND );                                   \
 }
 
-#if defined(MBEDTLS_X509_USE_C) || defined(MBEDTLS_X509_CREATE_C)
```
gilles-peskine-arm: How much of a code size increase does this cause in a build with `MBEDTLS_OID_C` (for the sake of rsa or pkparse) but without X509?

gilles-peskine-arm:
What's the merge strategy here? Why not merge https://github.com/ARMmbed/mbedtls/pull/2380 and then update `mbed-crypto/development` to the latest `mbedtls/development`? Cc @Patater

mpg:
@gilles-peskine-arm Good question. I suggest we come back to this when we're ready to progress on that front; yesterday we decided to postpone the split to after Q1, so I consider this PR and associated discussion parked for now.

Patater:
Let's merge into Mbed TLS development first and then merge to Mbed Crypto development. We did sibling PRs in the last release because there wasn't anything on the Mbed TLS development branch to merge from, only development-psa. Sibling PRs also made it a bit annoying to reconcile the changes when we did merge from TLS development to Crypto development, so it's preferable to avoid them.

PR #31, opened 2019-01-29
Unified tests/scripts/depends.py

gilles-peskine-arm:
Unify `curves.pl`, `key-exchanges.pl`, `depends-pkalgs.pl`, `depends-hashes.pl`, and [`depends-ciphers.pl`](https://github.com/ARMmbed/mbedtls/pull/1919) into a single, newly-written script. This isn't meant to be an identical replacement, but to provide at least the same amount of coverage.

For curves, key exchanges, ciphers and hashes, in addition to testing all-but-one settings in the group like the old scripts, also run the tests with a single option in the group.

The new script is unified (so less code to maintain), runs more tests (only-one in addition to all-but-one), and offers better tools to investigate failures.

The new script always starts from the `full` config (minus unimportant memory management options).

This is work in progress. Still to do:

* Fix all the builds.
    * Include patches from https://github.com/ARMmbed/mbedtls/pull/2019 for the only-one-curve jobs.
    * Include patches from https://github.com/ARMmbed/mbedtls/pull/1919 for the all-but-one-cipher-id jobs.
* Tutorial-type documentation to complement the reference documentation.

Fix https://github.com/ARMmbed/mbedtls/issues/1892

Obsoletes https://github.com/ARMmbed/mbedtls/pull/1919 and https://github.com/ARMmbed/mbedtls/pull/2019 (when fully done). Needs sideport/backport to mbedtls/development, 2.16 and 2.7.

### DIFF
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -118,6 +118,10 @@
 #error "MBEDTLS_ECP_RESTARTABLE defined, but it cannot coexist with an alternative ECP implementation"
 #endif
 
+#if defined(MBEDTLS_ECP_RESTARTABLE) && !defined(MBEDTLS_ECP_C)
+#error "MBEDTLS_ECP_RESTARTABLE defined, but not all prerequisites"
+#endif
+
 #if defined(MBEDTLS_ECDSA_DETERMINISTIC) && !defined(MBEDTLS_HMAC_DRBG_C)
 #error "MBEDTLS_ECDSA_DETERMINISTIC defined, but not all prerequisites"
 #endif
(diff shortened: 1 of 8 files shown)


Patater:
@gilles-peskine-arm Is this going to be picked up again soon?

We should keep PRs open while they are being worked on, and otherwise close them. This helps us to see what's active in the PR view and where we can help out with reviews; the list is a bit too unwieldy otherwise. When we want to continue the work, we can re-open this PR if what's been done in this PR is relevant to the new work.

gilles-peskine-arm:
> Is this going to be picked up again soon?

You're my boss, you tell me…

Please don't close PRs just because they're stalled. If we're planning to pick them up again, they should be open, so that they're found in a search for open PRs.

Patater:
Keeping on top of PRs needing merging is much easier when the list is kept tidy and only to active PRs. If this isn't active, we should close it. Please re-open if you pick up work on it again and you feel that approach is better than opening a new PR. Leaving half done work around everywhere will just lead to an unmaintainable mess. You can keep the branch open in your repo so you can see what half done work you have in progress, but we shouldn't use the shared PR view for inactive PRs we aren't actively working on. Maybe useful to tag with "stale" when closing so you can search for them. Not a good idea to have them open, though.

PR #32, opened 2019-01-29
Prepare support for 64 bit key ids under SPM

itayzafrir:
Preparation for type separation between SPE and NSPE.

This PR is the setup required for moving from 32 bit key ids to 64 bit key ids in Mbed OS in a system where SPM is present. When compiling as a library (in a system where there's no SPM) the changes in this PR are irrelevant and should not affect anything.

The file `crypto_types.h` will be handled differently in Mbed OS when SPM is present. There will be 2 different implementation of the file, one for the client and one for the server. Much like `crypto_structs.h` is separated between client and server. Specifically, the type `psa_key_id_t` will be 32 bits on the client and 64 bits on the server. This work is being done as part of https://github.com/ARMmbed/mbed-os/pull/9575.

### DIFF
--- a/library/psa_crypto_storage.c
+++ b/library/psa_crypto_storage.c
@@ -29,6 +29,20 @@
 
 #include <stdlib.h>
 #include <string.h>
+/*
+ * When MBEDTLS_PSA_CRYPTO_SPM is defined, the code is being built for SPM
+ * (Secure Partition Manager) integration which separates the code into two
+ * parts: NSPE (Non-Secure Processing Environment) and SPE (Secure Processing
+ * Environment). When building for the SPE, an additional header file should be
+ * included.
+ */
+#if defined(MBEDTLS_PSA_CRYPTO_SPM)
+/*
+ * PSA_CRYPTO_SECURE means that this file is compiled for the SPE.
+ * Some headers will be affected by this flag.
+ */
+#define PSA_CRYPTO_SECURE 1
+#endif
 
 #include "psa/crypto.h"
 #include "psa_crypto_storage.h"
--- a/library/psa_crypto_storage.c
+++ b/library/psa_crypto_storage.c
@@ -29,21 +29,8 @@
 
 #include <stdlib.h>
 #include <string.h>
-/*
- * When MBEDTLS_PSA_CRYPTO_SPM is defined, the code is being built for SPM
- * (Secure Partition Manager) integration which separates the code into two
- * parts: NSPE (Non-Secure Processing Environment) and SPE (Secure Processing
- * Environment). When building for the SPE, an additional header file should be
- * included.
- */
-#if defined(MBEDTLS_PSA_CRYPTO_SPM)
-/*
- * PSA_CRYPTO_SECURE means that this file is compiled for the SPE.
- * Some headers will be affected by this flag.
- */
-#define PSA_CRYPTO_SECURE 1
-#endif
 
+#include "psa_crypto_service_integration.h"
 #include "psa/crypto.h"
 #include "psa_crypto_storage.h"
 #include "psa_crypto_storage_backend.h"
--- a/library/psa_crypto_storage_its.c
+++ b/library/psa_crypto_storage_its.c
@@ -26,22 +26,9 @@
 #endif
 
 #if defined(MBEDTLS_PSA_CRYPTO_STORAGE_ITS_C)
-/*
- * When MBEDTLS_PSA_CRYPTO_SPM is defined, the code is being built for SPM
- * (Secure Partition Manager) integration which separates the code into two
- * parts: NSPE (Non-Secure Processing Environment) and SPE (Secure Processing
- * Environment). When building for the SPE, an additional header file should be
- * included.
- */
-#if defined(MBEDTLS_PSA_CRYPTO_SPM)
-/*
- * PSA_CRYPTO_SECURE means that this file is compiled for the SPE.
- * Some headers will be affected by this flag.
- */
-#define PSA_CRYPTO_SECURE 1
-#endif
 
 #include "psa/error.h"
+#include "psa_crypto_service_integration.h"
 #include "psa/crypto.h"
 #include "psa_crypto_storage_backend.h"
 #include "psa/internal_trusted_storage.h"
--- a/library/psa_crypto_storage_its.c
+++ b/library/psa_crypto_storage_its.c
@@ -26,6 +26,20 @@
 #endif
 
 #if defined(MBEDTLS_PSA_CRYPTO_STORAGE_ITS_C)
+/*
+ * When MBEDTLS_PSA_CRYPTO_SPM is defined, the code is being built for SPM
+ * (Secure Partition Manager) integration which separates the code into two
+ * parts: NSPE (Non-Secure Processing Environment) and SPE (Secure Processing
+ * Environment). When building for the SPE, an additional header file should be
+ * included.
+ */
+#if defined(MBEDTLS_PSA_CRYPTO_SPM)
+/*
+ * PSA_CRYPTO_SECURE means that this file is compiled for the SPE.
+ * Some headers will be affected by this flag.
+ */
+#define PSA_CRYPTO_SECURE 1
+#endif
 
 #include "psa/error.h"
 #include "psa/crypto.h"
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -26,22 +26,8 @@
 #endif
 
 #if defined(MBEDTLS_PSA_CRYPTO_C)
-/*
- * When MBEDTLS_PSA_CRYPTO_SPM is defined, the code is being built for SPM
- * (Secure Partition Manager) integration which separates the code into two
- * parts: NSPE (Non-Secure Processing Environment) and SPE (Secure Processing
- * Environment). When building for the SPE, an additional header file should be
- * included.
- */
-#if defined(MBEDTLS_PSA_CRYPTO_SPM)
-/*
- * PSA_CRYPTO_SECURE means that this file is compiled for the SPE.
- * Some headers will be affected by this flag.
- */
-#define PSA_CRYPTO_SECURE 1
-#include "crypto_spe.h"
-#endif
 
+#include "psa_crypto_service_integration.h"
 #include "psa/crypto.h"
 
 #include "psa_crypto_core.h"
--- a/library/psa_crypto_service_integration.h
+++ b/library/psa_crypto_service_integration.h
@@ -1,0 +1,40 @@
+/*  Copyright (C) 2019, ARM Limited, All Rights Reserved
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *  not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  This file is part of mbed TLS (https://tls.mbed.org)
+ */
+
+#ifndef PSA_CRYPTO_SERVICE_INTEGRATION_H
+#define PSA_CRYPTO_SERVICE_INTEGRATION_H
+
+/*
+ * When MBEDTLS_PSA_CRYPTO_SPM is defined, the code is being built for SPM
+ * (Secure Partition Manager) integration which separates the code into two
+ * parts: NSPE (Non-Secure Processing Environment) and SPE (Secure Processing
+ * Environment). When building for the SPE, an additional header file should be
+ * included.
+ */
+#if defined(MBEDTLS_PSA_CRYPTO_SPM)
+/*
+ * PSA_CRYPTO_SECURE means that the file which included this file is being
+ * compiled for SPE. The files crypto_structs.h and crypto_types.h have
+ * different implementations for NSPE and SPE and are compiled according to this
+ * flag.
+ */
+#define PSA_CRYPTO_SECURE 1
+#include "crypto_spe.h"
+#endif // MBEDTLS_PSA_CRYPTO_SPM
+
+#endif // PSA_CRYPTO_SERVICE_INTEGRATION_H
--- a/library/psa_crypto_slot_management.c
+++ b/library/psa_crypto_slot_management.c
@@ -26,22 +26,8 @@
 #endif
 
 #if defined(MBEDTLS_PSA_CRYPTO_C)
-/*
- * When MBEDTLS_PSA_CRYPTO_SPM is defined, the code is being built for SPM
- * (Secure Partition Manager) integration which separates the code into two
- * parts: NSPE (Non-Secure Processing Environment) and SPE (Secure Processing
- * Environment). When building for the SPE, an additional header file should be
- * included.
- */
-#if defined(MBEDTLS_PSA_CRYPTO_SPM)
-/*
- * PSA_CRYPTO_SECURE means that this file is compiled for the SPE.
- * Some headers will be affected by this flag.
- */
-#define PSA_CRYPTO_SECURE 1
-#include "crypto_spe.h"
-#endif
 
+#include "psa_crypto_service_integration.h"
 #include "psa/crypto.h"
 
 #include "psa_crypto_core.h"
--- a/visualc/VS2010/mbedTLS.vcxproj
+++ b/visualc/VS2010/mbedTLS.vcxproj
@@ -238,6 +238,7 @@
     <ClInclude Include="..\..\include\psa\crypto_values.h" />
     <ClInclude Include="..\..\library/psa_crypto_core.h" />
     <ClInclude Include="..\..\library/psa_crypto_invasive.h" />
+    <ClInclude Include="..\..\library/psa_crypto_service_integration.h" />
     <ClInclude Include="..\..\library/psa_crypto_slot_management.h" />
     <ClInclude Include="..\..\library/psa_crypto_storage.h" />
     <ClInclude Include="..\..\library/psa_crypto_storage_backend.h" />


itayzafrir:
@Patater, @gilles-peskine-arm 
ITS unique identifiers are being changed 64 bits (as of https://github.com/ARMmbed/mbed-os/pull/9529). To support this, in 3a02e5376fb61e5926bfc6017bfe4aa6c9aeb654 I'v changed the value of `PSA_MAX_PERSISTENT_KEY_IDENTIFIER ` from `0xffff0000` to `0xffffffff00000000`.

Should the following also be updated?
https://github.com/ARMmbed/mbed-crypto/blob/3a02e5376fb61e5926bfc6017bfe4aa6c9aeb654/include/psa/crypto_extra.h#L38
If so, is there anyone external to crypto who is dependent on this value? I'v searched through Mbed OS and haven't seen anyone using this value directly outside of the crypto context. Can there be any backwards compatibility issues with changing this value?

Patater:
What's the logic behind changing `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` to 0xffff0000ffffffff? Could you put this in the commit message place?

The default random seed implementation uses the `PSA_CRYPTO_ITS_RANDOM_SEED_UID` define, but doesn't care about the value. The issue is with the upgrade path from a system that has an already injected random seed and we update the UID; the default random seed implementation would need to look in two places: the old and new location. A migration tool could be made for these devices to move anything present at the old UID to the new one, but there may not be a need yet.

@shelib01 Do we have any devices in the field using this entropy injection feature? Do we need to keep using the old UID to support those devices? If we haven't deployed anything yet, would you agree its okay to change the UID not before we get devices out in the field?

Patater:
Could you also have a look at the compilation failures? It superficially looks like there are some places that mishandle 64-bit numbers.

itayzafrir:
@Patater I've amended the commit and added the reasoning of changing `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` to `0xffff0000ffffffff` inside the commit message (919b89ec656c0693631547cf635774b8cd1ac0fc). Old branch can b found @ https://github.com/itayzafrir/mbed-crypto/tree/ipc-support-64-bit-key-ids-1.

About changing the value of `PSA_CRYPTO_ITS_RANDOM_SEED_UID`, I don't think it's necessary. Since `PSA_CRYPTO_ITS_RANDOM_SEED_UID` is defined as `0xFFFFFF52`, and the original value of `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` was set to `0xFFFF0000`, it wouldn't have passed the validation check at https://github.com/itayzafrir/mbed-crypto/blob/919b89ec656c0693631547cf635774b8cd1ac0fc/library/psa_crypto_slot_management.c#L218 anyways. A quick check revealed that the `PSA_CRYPTO_ITS_RANDOM_SEED_UID` doesn't go through the regular flow and is never tested against `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` in any case. Therefore IMO we can leave it as is.

itayzafrir:
Build failed on `test_suite_timing` - known issue not related to the changes in this PR.

library/psa_crypto_storage.h
```diff
@@ -59,7 +59,11 @@ extern "C" {
  * This limitation will probably become moot when we implement client
  * separation for key storage.
  */
+#if defined(MBEDTLS_PSA_CRYPTO_STORAGE_ITS_C)
+#define PSA_MAX_PERSISTENT_KEY_IDENTIFIER 0xffff0000ffffffff
+#else
 #define PSA_MAX_PERSISTENT_KEY_IDENTIFIER 0xffff0000
+#endif
```
Patater: Do we need to change `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` based on the presence of ITS or could we always use `0xffff0000ffffffff`?

itayzafrir: We should only change it for ITS. If we change it always, the following check will fail when compiling in a non-SPM system:
https://github.com/itayzafrir/mbed-crypto/blob/919b89ec656c0693631547cf635774b8cd1ac0fc/library/psa_crypto_slot_management.c#L205-L219
Actually clang failed this because the condition `if( id >= PSA_MAX_PERSISTENT_KEY_IDENTIFIER )` would never be true because id is 32 bits in a non-SPM system. In an SPM supported system, `psa_key_id_t` is 32 bits on the client and a 64 bit type on the service side.

Patater: Could it be 64-bits always?

itayzafrir: Can what be 64 bits? If you're asking about the `psa_key_id_t` then not really, it would then make ITS to be 96 bits and we'll be in the same loop again.

itayzafrir: @Patater I see you point, I overlooked the fact that ITS is a also used in a non-segmented system.

> In a segmented system:
ITS would be present on SPE, so MBEDTLS_PSA_CRYPTO_STORAGE_ITS_C would be used so we can use it. ITS would also be present on NSPE, but we wouldn't define ITS_C for the Mbed Crypto client code?

Not exactly, as the crypto client code is implemented just as a proxy, I think it's meaningless for the crypto client side whether ITS is enabled or not. Same thing goes for `PSA_MAX_PERSISTENT_KEY_IDENTIFIER`, it's not used by the client code and it isn't part of the PSA Crypto API.

> In a non-segmented system:
The server and client Mbed Crypto are the same. ITS is available, but we'd not define MBEDTLS_PSA_CRYPTO_STORAGE_ITS_C?

I guess we can't really do that...

---

So, since `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` is not part of the official PSA Crypto API it might be meaningless to other PSA Crypto implementations so I'm not sure if placing it in `crypto_types.h` makes sense...IMO it would make more sense placing it `crypto_extra.h` and keeping different copies of `crypto_extra.h` in Mbed OS, one for client and for server, same as is already being done for `crypto_structs.h` and is planned (https://github.com/ARMmbed/mbed-os/pull/9575) to be done with `crypto_types.h`. However, this forces us to maintain yet another file which is different for client/server. I think that a simpler solution is to just change (919b89ec656c0693631547cf635774b8cd1ac0fc):

From:
```c
#if defined(MBEDTLS_PSA_CRYPTO_STORAGE_ITS_C)
#define PSA_MAX_PERSISTENT_KEY_IDENTIFIER 0xffff0000ffffffff
#else
#define PSA_MAX_PERSISTENT_KEY_IDENTIFIER 0xffff0000	
#endif
```
To:
```c
#if defined(MBEDTLS_PSA_CRYPTO_SPM)
#define PSA_MAX_PERSISTENT_KEY_IDENTIFIER 0xffff0000ffffffff
#else
#define PSA_MAX_PERSISTENT_KEY_IDENTIFIER 0xffff0000	
#endif
```

`MBEDTLS_PSA_CRYPTO_SPM` is set only when compiling the _server side_, see https://github.com/ARMmbed/mbed-os/blob/57d8915a899a59999d7e2a9695e20a8a18cb54bd/targets/targets.json#L7905

Please let me know if this is an acceptable solution and I will update this PR.

itayzafrir: I changed it, also changed the `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` value from 0xffff0000ffffffff to 0xffff000000000000 when compiled for server side. I amended the commit, old branch https://github.com/itayzafrir/mbed-crypto/tree/ipc-support-64-bit-key-ids-2.

Patater: If a user calls `psa_create_key(PSA_KEY_LIFETIME_PERSISTENT, 0xFFFFFF52, ...)`, how do we know they won't ever get a persistent allocation with `PSA_CRYPTO_ITS_RANDOM_SEED_UID` as the storage UID for their key? `0xFFFFFF52` is certainly less than `PSA_MAX_PERSISTENT_KEY_IDENTIFIER`. Don't we need to explicitly block this UID when we select the storage ID? Previously, we depended on the `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` checks to block allocation of this storage location.

itayzafrir: There's no need for an explicit block on `PSA_CRYPTO_ITS_RANDOM_SEED_UID`.

In a system where SPM is present and as of https://github.com/ARMmbed/mbed-os/pull/9575, if a user were to call `psa_open_key(PSA_KEY_LIFETIME_PERSISTENT, PSA_CRYPTO_ITS_RANDOM_SEED_UID, &handle)` or `psa_create_key(PSA_KEY_LIFETIME_PERSISTENT, PSA_CRYPTO_ITS_RANDOM_SEED_UID, &handle)` the ID (`psa_key_id_t`) sent from the client (`PSA_CRYPTO_ITS_RANDOM_SEED_UID` in this case or any other ID for that matter) is translated by the server (crypto partition) into a different ID prior to calling the actual Mbed Crypto library. 

The ID (`psa_key_id_t`) on the server side is 64 bits and it translated from what the client sent (32 bits) in the following way:
Place the 32 bit ID received from the client in the upper 32 bits of the 64 bit `psa_key_id_t`, and the calling partition id (32 bits) in the lower 32 bits of the 64 bit `psa_key_id_t`.

| 32 bits PSA_CRYPTO_ITS_RANDOM_SEED_UID | 32 bits CALLING PARTITION ID |
|---|---|

The table shows what the actual ID sent to ITS will looks like.

Example: If a client were to call `psa_open_key` or `psa_create_key` with `PSA_CRYPTO_ITS_RANDOM_SEED_UID` as the ID, the actual ID value sent to ITS will be:
1. If calling from NSPE (partition id `-1`) - ID sent to ITS = `0xffffff52ffffffff`
2. If calling from different partition (for example partition id `3`) - ID sent to ITS = `0xffffff5200000003`

So there's no chance of overwriting the injected `PSA_CRYPTO_ITS_RANDOM_SEED_UID`.
Moreover, the results from the examples above would not pass the check against `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` as both  `0xffffff52ffffffff` and `0xffffff5200000003` are greater than `PSA_MAX_PERSISTENT_KEY_IDENTIFIER`.

All this magic is planned to happen in https://github.com/ARMmbed/mbed-os/pull/9575, this PR is about the required setup for 9575. You might want to take a look at it to see the complete picture.

Patater:
What testing has been done so far?

Did we test that a system (e.g. K64F) with injected entropy in Mbed OS 5.11 still keeps proper state after reflashing the K64F to use the latest Mbed OS development branch? Is the saved entropy still available and usable?

itayzafrir:
> What testing has been done so far?
> 
> Did we test that a system (e.g. K64F) with injected entropy in Mbed OS 5.11 still keeps proper state after reflashing the K64F to use the latest Mbed OS development branch? Is the saved entropy still available and usable?

Yes this has been tested as explained in mail.

library/psa_crypto_storage.h
Patater: I understand for client/server architectures that this would work. Thanks for your explanation.

How about for K64F or other non-segmented systems?

itayzafrir: For non-segmented systems this has no effect and everything remains as it is now. `MBEDTLS_PSA_CRYPTO_SPM` is currently only defined for the server side of psoc6 (`FUTURE_SEQUANA_M0_PSA`). See https://github.com/ARMmbed/mbed-os/blob/57d8915a899a59999d7e2a9695e20a8a18cb54bd/targets/targets.json#L7905

Patater:
@itayzafrir Could you please add a specification explaining what the promise between versions is exactly, and/or provide an automated regression test?

Thanks

Patater:
The specification is being provided through https://github.com/ARMmbed/mbed-crypto/pull/53

This PR is OK for merge alongside https://github.com/ARMmbed/mbed-crypto/pull/53

Patater:

library/psa_crypto_storage.c
```diff
@@ -29,6 +29,20 @@
...
 #include <stdlib.h>
 #include <string.h>
+/*
+ * When MBEDTLS_PSA_CRYPTO_SPM is defined, the code is being built for SPM
+ * (Secure Partition Manager) integration which separates the code into two
+ * parts: NSPE (Non-Secure Processing Environment) and SPE (Secure Processing
+ * Environment). When building for the SPE, an additional header file should be
+ * included.
+ */
+#if defined(MBEDTLS_PSA_CRYPTO_SPM)
```
gilles-peskine-arm: Why is this logic repeated three times? Please put it in a common internal header.

library/psa_crypto_storage.h
```diff
@@ -59,7 +59,11 @@ extern "C" {
  * This limitation will probably become moot when we implement client
  * separation for key storage.
  */
+#if defined(MBEDTLS_PSA_CRYPTO_SPM)
+#define PSA_MAX_PERSISTENT_KEY_IDENTIFIER 0xffff000000000000
```
gilles-peskine-arm: I don't understand this value. Why is 0xffffffff00001234 ok (as the key id 0xffffffff with the owner 0x00001234)? What is this macro supposed to be used for? At the moment, it's one plus the maximum key id value that an application may use.

library/psa_crypto_storage_its.c
```diff
@@ -26,6 +26,20 @@
...
+ * When MBEDTLS_PSA_CRYPTO_SPM is defined, the code is being built for SPM
+ * (Secure Partition Manager) integration which separates the code into two
+ * parts: NSPE (Non-Secure Processing Environment) and SPE (Secure Processing
+ * Environment). When building for the SPE, an additional header file should be
+ * included.
+ */
+#if defined(MBEDTLS_PSA_CRYPTO_SPM)
+/*
+ * PSA_CRYPTO_SECURE means that this file is compiled for the SPE.
+ * Some headers will be affected by this flag.
```
gilles-peskine-arm: What is “some headers”? Since none of the code in this repository uses `PSA_CRYPTO_SECURE`, please explain what clause of the contract between Mbed Crypto and [some PSA component] this is about.

(I should have made this comment on the PR that originally introduced this wording, but it's still relevant.)

library/psa_crypto_storage.h
itayzafrir: This is your macro, we're not using it externally. This has been explained in https://github.com/ARMmbed/mbed-crypto/pull/32#discussion_r254597745
Please take a look

gilles-peskine-arm: But on PSA service integration, `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` does not limit the key id values that applications can use. Applications can use any value between 0 and 0xffffffff inclusive, unless there's a check in the service code. So what's the point of setting `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` to this particular value?

itayzafrir: The service uses the library as a black box. The service is not communicating with neither ITS (directly) nor any internal crypto APIs. So the code flow remains the same from the librarys' point of view.

gilles-peskine-arm: In this case, nothing is enforcing a particular range of key ids for an application. Which may be ok (it's a design decision that we're still free to take), but then `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` is not useful in a PSA service integration.

itayzafrir: Why not? The following code runs when crypto is used as a library but also as a PSA service:
https://github.com/ARMmbed/mbed-crypto/blob/2d7e5fe31da55d47748d259d11c08800261a4fd1/library/psa_crypto_slot_management.c#L205-L219

Shouldn't the library (in both modes) reserve key ids for it's internal use as stated in documentation? > Reserve a whole range of key slots just in case something else comes up.

library/psa_crypto_storage.c
itayzafrir: Added a common internal header file `psa_crypto_service_integration.h` in d8a8074d6c16a1e1f017c139bb04e2162087b8be

library/psa_crypto_storage.h
gilles-peskine-arm: Ok, my bad, this _does_ enforce an upper limit of 0xfffeffff for key ids, since the key id is in the upper 32 bits. However, the checks as described are still not correct.

On an upgrade of ITS from 32-bit identifiers to 64-bit identifiers, how are identifiers converted?

* If the conversion is `uint64_t new_key_id = old_key_id`, then the entropy file is now `PSA_CRYPTO_ITS_RANDOM_SEED_UID` which is `0x00000000ffffff52`. This is key id 0 of owner 0xffffff52. Per https://github.com/ARMmbed/mbed-crypto/pull/32#discussion_r256532294 (no validation of key ids in the service) these are valid values. 0xffffff52 has the most significant bit set which makes it a non-secure partition. We don't have multiple non-secure partitions on Mbed OS (yet?), but we do on Linux.
* If the conversion is `uint64_t new_key_id = (uint64_t)old_key_id << 32`, then `PSA_CRYPTO_ITS_RANDOM_SEED_UID` needs to change to `0xffffff5200000000`, or we need an upgrade path.

itayzafrir: As far as I know the only key that must be preserved between Mbed OS versions 5.11 and 5.12 is the injected entropy key with id `PSA_CRYPTO_ITS_RANDOM_SEED_UID`. 

**_NON_-PSA Target: injected entropy preservation between versions 5.11 and 5.12:**
This has been tested and validated on a K64F (_non_-PSA target i.e. without the crypto service), see https://github.com/ARMmbed/mbed-crypto/pull/32#issuecomment-461457661.

**PSA Target: injected entropy preservation between versions 5.11 and 5.12:**
I don't see a reason to change the value of `PSA_CRYPTO_ITS_RANDOM_SEED_UID`, I'll try to explain my point of view...

The following is the entropy inject function implemented in mbed crypto:
https://github.com/ARMmbed/mbed-crypto/blob/0574e6a7bd0bec47e45bbd30db2e5236896b38ea/library/psa_crypto.c#L4427-L4455
As you can see entropy injection does _not_ go through the "normal" flow other crypto keys do (i.e. `psa_open_key` or `psa_create_key` etc.), it talks to the ITS directly. There's no intervention by the crypto service (or anything else) that changes this key id into anything else. This means that we can keep using the same `PSA_CRYPTO_ITS_RANDOM_SEED_UID` value both in a system where ITS ids are 32 or 64 bits. The ITS does not care about the _meaning_ of the id it recieves.

As to overwriting or obtaining a handle to a key with id `PSA_CRYPTO_ITS_RANDOM_SEED_UID`, the only way to access this ITS record is by talking to ITS directly, it will not be possible to do so via crypto API:

**In a PSA target**, if a user were to call `psa_open_key` or `psa_create_key` with id `PSA_CRYPTO_ITS_RANDOM_SEED_UID` the service would translate this value to:
* If calling from NSPE (partition id `-1`) - ID sent to ITS = `0xffffff52ffffffff`
* If calling from different partition (for example partition id `3`) - ID sent to ITS = `0xffffff5200000003`

So I don't see a possible way of obtaining a handle to a key with id `0xffffff52` when going through the crypto service.

Moreover, the results from these 2 examples  would not pass the check (in function `psa_internal_make_key_persistent`) against `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` (when it's defined to `0xffff000000000000`) as both `0xffffff52ffffffff` and `0xffffff5200000003` are greater than `0xffff000000000000` and the function would return a `PSA_ERROR_INVALID_ARGUMENT`.

**In a non-PSA system**, the above is also true, if `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` is defined as `0xffff0000` (as it is now), it will also fail the validation in `psa_internal_make_key_persistent` as `0xffffff52` is greater than `0xffff0000`.

One more point, the crypto service does (as part of https://github.com/ARMmbed/mbed-os/pull/9575) validate that key id is **not** 0 _prior_ to calling the library's open/create a key, see:
https://github.com/ARMmbed/mbed-os/blob/3bc5f9086c3bbf1471906d6a4753457cf46f19cf/components/TARGET_PSA/services/crypto/COMPONENT_SPE/psa_crypto_partition.c#L1179
and
https://github.com/ARMmbed/mbed-os/blob/3bc5f9086c3bbf1471906d6a4753457cf46f19cf/components/TARGET_PSA/services/crypto/COMPONENT_SPE/psa_crypto_partition.c#L1201

Clarification on:
> the only way to access this ITS record is by talking to ITS directly

It will only be possible to do so from the partition that created the ITS record. ITS has it's own validation/enforcement mechanism which prevents accessing ITS records from partitions which are not the record's owner. They do so by converting the id received into a _string_ and prepending the calling partition id to that string. So if ITS was called **directly** from NSPE (`0xffffffff`) with id `PSA_CRYPTO_ITS_RANDOM_SEED_UID` (`0xffffff52`), the actual ITS record id would be converted (by ITS) into `ffffffff_ffffff52`. In our case, in a PSA system (where the crypto service id is `0x23`), the injected entropy id that **ITS** eventually stores (as a string) is `23_ffffff52`. This ensures that the entropy key will only be accessible from the crypto partition and only via direct access to ITS (i.e. without going through open/create key APIs).

itayzafrir:
@gilles-peskine-arm please take a look at my answers in https://github.com/ARMmbed/mbed-crypto/pull/32#discussion_r257499554.

library/psa_crypto_storage.h
gilles-peskine-arm: > **In a PSA target**, if a user were to call `psa_open_key` or `psa_create_key` with id `PSA_CRYPTO_ITS_RANDOM_SEED_UID` the service would translate this value to:

> If calling from NSPE (partition id `-1`) - ID sent to ITS = `0xffffff52ffffffff`

On Linux, there are multiple partitions in the NSPE and they're numbered 0xffffffff, 0xfffffffe, etc. I don't know if it's already implemented, but if it isn't it's only a matter of weeks, and we need a solution that works in this time scale.

If partition 0xffffff52 wants to store key ID 0, this can only be blocked at the service level. The safety of the entropy seed file depends on this check in the service. I would strongly prefer a more robust solution where the entropy seed is in a domain that the PRoT already reserves for itself: owner_uid = 0.

itayzafrir:
@gilles-peskine-arm @Patater I have reverted commit _Change PSA_MAX_PERSISTENT_KEY_IDENTIFIER to 64 bits when using ITS_ (43b9f2d54b05f4bdb7ade03d002c8712fbb5f924) and also rebased to solve merge conflicts in file `psa_crypto_storage_its.c`.

Old branch https://github.com/itayzafrir/mbed-crypto/tree/ipc-support-64-bit-key-ids-3

Patater:
@itayzafrir Could you please delete the commit instead of add a revert commit?

itayzafrir:
@Patater, I removed the commit and also squashed some commits to make history clearer.

gilles-peskine-arm:
Note that the main goal of this PR, namely support for 64-bit key file IDs encoding the key owner, is now fulfilled by #59. The goal of this PR is to integrate #59 into the PSA crypto service.

itayzafrir:
@gilles-peskine-arm 
> we will remove it at some point, and use `psa_key_file_id_t` in `crypto.h`...

Understood, it shouldn't make too much of a difference in the service side as long as we'll still be able to have separation for the identifier between the crypto service client the crypto service itself.

Patater:

PR #38, opened 2019-01-31
API specifications at version 1.0 beta1

gilles-peskine-arm:
Upload the generated PDF and HTML for the PSA Crypto API 1.0 beta1.


gilles-peskine-arm:
Regarding the Doxygen issues: this will be fixed in the spec after migrating to the new format. For the implementation, I don't know how to fix it.

The text says that it's linking to the _specification_. That's why it's pointing to a different branch. Later we'll make a process to have the library documentation on tls.mbed.org or crypto.mbed.org or whatever.

Patater:
This is not quite passing CI... there are some failures in all.sh

gilles-peskine-arm:
Failures are expected since the code hasn't caught up with the specification. That's why this is for the `psa-crypto-api` branch and not for `development`.

Patater:
OK, didn't catch that this was for non-development branch. Mergin'.

PR #43, opened 2019-02-05
Update to a development version of Mbed TLS 2.16.0

Patater:
Update to a development version of Mbed TLS 2.16.0 without parameter validation as a step along the way to getting the full Mbed TLS 2.16.0 changes included in Mbed Crypto. The particular commit chosen as an intermediate point begins before parameter validation to avoid having to resolve parameter validation conflicts in both ARMmbed/mbedtls:development-psa and ARMmbed/mbed-crypto:development. The idea is to bring in as much into Mbed Crypto as soon as we can, and this is a good point in Mbed TLS 2.16.0 history from which to bring in those changes.

This brings in ECDH changes that are needed to support using Mbed Crypto as a submodule in Mbed TLS versions 2.16.0 or greater.


ChangeLog
```diff
@@ -6,6 +6,58 @@ Changes
    * Add unit tests for AES-GCM when called through mbedtls_cipher_auth_xxx()
      from the cipher abstraction layer. Fixes #2198.
 
+= mbed TLS 2.16.0 branch released xxxx-xx-xx
```
gilles-peskine-arm: This looks like it lists the changes in Mbed TLS 2.16.0 without a release date, but actually it only takes some of the changes of 2.16.0. Please don't use a version number here.

We don't have a precedent for splitting unreleased changes into two sections based on their origin. I think it would be fine to have a single `x.xx.xx` section here with the changes from both branches. But if you really think there should be separate sections, please use a non-misleading presentation, e.g. `mbed TLS 2.14+01b34fb316a5d8c37244719be3f11ccf2ae85e41 unreleased`

library/rsa.c
```diff
@@ -1423,64 +1530,109 @@ int mbedtls_rsa_rsaes_pkcs1_v15_decrypt( mbedtls_rsa_context *ctx,
...
+     * does not depend on the plaintext size. After this move, the
+     * starting location of the plaintext is no longer sensitive
+     * information. */
+    mem_move_to_left( buf + ilen - plaintext_max_size,
+                      plaintext_max_size,
+                      plaintext_max_size - plaintext_size );
+
+    /* Finally copy the decrypted plaintext plus trailing zeros
+     * into the output buffer. */
+    memcpy( output, buf + ilen - plaintext_max_size, plaintext_max_size );
```
gilles-peskine-arm: This breaks when `output_size` is 0 and `output` is not a valid pointer (e.g. null). [It's a known bug](https://github.com/ARMmbed/mbedtls/issues/1842) which came up when I was writing tests on the PSA branch. I thought CI would catch it, but it turns out that there's a test case with `max_output_len=0` only for OAEP, not for v15 decryption. So this passes CI, but it's still a functional regression.

I suggest the following change, which may be better done in a subsequent commit than in the merge itself since it touches highly sensitive code:

```c
    /* Finally copy the decrypted plaintext plus trailing zeros
     * into the output buffer. */
    memcpy( output, buf + ilen - plaintext_max_size, plaintext_max_size );
     * into the output buffer. If output_max_len is 0, then output
     * may be an invalid pointer and memcpy would be undefined;
     * check for this depending only on output_max_len which is public
     * information. */
    if( output_max_len != 0 )
        memcpy( output, buf + ilen - plaintext_max_size, plaintext_max_size );
```

ChangeLog
Patater: As explained in the commit message, we have two sections because Mbed Crypto has a ChangeLog entry already added to the x.xx.x version for a future release. 2.16.0 without a release date was used as the marker for a development version of 2.16.0 (pre-release). Seems like this explanation wasn't clear enough, was in a location not looked at, and the marker used was too confusing in isolation. Will use 2.14+01b34fb316a5 as the version marker.

gilles-peskine-arm: I know what you did, but I don't understand why you did it, and the `2.16.0` would have to stand on its own because it doesn't look like one would need to track down the commit that introduced it to realize that `2.16.0` does not in fact mean 2.16.0.

library/rsa.c
Patater: I've added a commit to fix this use case.

ChangeLog
Patater: I've added a commit to change the identifier to something hopefully more clear.

Patater:
Force pushed to fix typo in commit message.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -3258,6 +3258,7 @@ void asymmetric_decrypt_fail( int key_type_arg,
                               int alg_arg,
                               data_t *input_data,
                               data_t *label,
+                              int expected_output_size,
```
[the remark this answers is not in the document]
Patater: `expected_output_size` is assigned to `output_size`, a `size_t`, before use. I'll rename to match naming expectations and make the code less confusing.

Patater:
Rebased to address review comments. Modified how must-fail asymmetric decryption tests are updated to handle test-case-controlled output buffer sizes. Applied fixup commit to merge commit.

Old branch at https://github.com/Patater/mbed-crypto/tree/update-2.16-dev-1

mpg:
The Jenkins run on the merged PR only failed on a in a 3d+fragmentation test that is known to be flaky with the improper version of GnuTLS currently in our CI. That's as good as a pass, so I'm removing "needs: ci".

PR #44, opened 2019-02-05
Fix double return statement in cipher.c

mpg:
This was introduced in ce1ddee13a171

It was caught by the IAR compiler in the Mbed TLS CI in one of our PRs. Not sure why it wasn't caught by other compilers earlier.


mpg:
CI only failed in the known-fragile timing test on BSD (twice), so as good as a pass.

Patater:
In additional to the FreeBSD timing test failure, another CI failure is `DTLS proxy: 3d, gnutls server, fragmentation`, a known flaky test with the version of GnuTLS we currently use.

PR #46, opened 2019-02-07
psa: Test fresh contexts have default behavior

Patater:
Test that freshly-initialized contexts exhibit default behavior through
the API. Do this without depending on the internal representation of the
contexts. This provides better portability of our tests on compilers
like MSVC.

Testing done:
Builds and runs fine on Visual Studio 2017. Passes all tests run by building the CMake target `RUN_TESTS`.

### DIFF
--- a/tests/suites/test_suite_psa_crypto.function
+++ b/tests/suites/test_suite_psa_crypto.function
@@ -1441,15 +1441,15 @@ void key_policy_init( )
 
     memset( &zero, 0, sizeof( zero ) );
 
-    /* Although not technically guaranteed by the C standard nor the PSA Crypto
-     * specification, we test that all valid ways of initializing the object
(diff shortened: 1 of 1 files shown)


tests/scripts/all.sh
```diff
@@ -119,6 +119,7 @@ pre_initialize_variables () {
     MEMORY=0
     FORCE=0
     KEEP_GOING=0
+    CTEST_OUTPUT_ON_FAILURE=1
```
gilles-peskine-arm: That won't get you far unless you `export` it.

Patater: Correct, it doesn't do anything. Will add in separate PR.

Patater:
Force pushed to remove unrelated-to-this-PR CMake verbosity enhancement.

Patater:

PR #48, opened 2019-02-11
Allow NULL buffers in psa_copy_key_material when the key size is zero

dgreen-arm:
We should be able to copy a key with zero length. Currently this is prevented in `psa_copy_key_material`, however the other functions it calls, `psa_internal_export_key` and `psa_import_key` already handle zero-length keys. This change allows NULL buffers in the case when the key length is zero.

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -1227,7 +1227,7 @@ static psa_status_t psa_copy_key_material( const psa_key_slot_t *source,
     buffer_size = PSA_KEY_EXPORT_MAX_SIZE( source->type,
                                            psa_get_key_bits( source ) );
     buffer = mbedtls_calloc( 1, buffer_size );
-    if( buffer == NULL )
+    if( buffer == NULL && buffer_size != 0 )
         return( PSA_ERROR_INSUFFICIENT_MEMORY );
     status = psa_internal_export_key( source, buffer, buffer_size, &length, 0 );
     if( status != PSA_SUCCESS )
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -1235,7 +1235,8 @@ static psa_status_t psa_copy_key_material( const psa_key_slot_t *source,
     status = psa_import_key( target, source->type, buffer, length );
 
 exit:
-    mbedtls_platform_zeroize( buffer, buffer_size );
+    if( buffer_size != 0 )
+        mbedtls_platform_zeroize( buffer, buffer_size );
     mbedtls_free( buffer );
     return( status );
 }


library/psa_crypto.c
```diff
@@ -1227,7 +1227,7 @@ static psa_status_t psa_copy_key_material( const psa_key_slot_t *source,
     buffer_size = PSA_KEY_EXPORT_MAX_SIZE( source->type,
                                            psa_get_key_bits( source ) );
     buffer = mbedtls_calloc( 1, buffer_size );
-    if( buffer == NULL )
+    if( buffer == NULL && buffer_size != 0 )
```
Patater: `psa_internal_export_key()`, `psa_import_key()`, and `mbedtls_free()` are well defined when `buffer` is `NULL` and `buffer_size` is `0`, but how about `mbedtls_platform_zeroize()` which might just call libc's `memset()`?

dgreen-arm:
There is indeed a test case which caught it. However due to a mistake I made, the PR job wasn't running the components of all.sh properly. That's been fixed now, so this issue is showing up in other PR's.

library/psa_crypto.c
dgreen-arm: I've made the zeroization conditional on the buffer being of a non-zero length.

library/psa_crypto.c
```diff
@@ -1227,15 +1227,16 @@ static psa_status_t psa_copy_key_material( const psa_key_slot_t *source,
...
         return( PSA_ERROR_INSUFFICIENT_MEMORY );
     status = psa_internal_export_key( source, buffer, buffer_size, &length, 0 );
     if( status != PSA_SUCCESS )
         goto exit;
     status = psa_import_key( target, source->type, buffer, length );
 
 exit:
-    mbedtls_platform_zeroize( buffer, buffer_size );
+    if( buffer_size != 0 )
+        mbedtls_platform_zeroize( buffer, buffer_size );
```
gilles-peskine-arm: This isn't the only place where we call `mbedtls_platform_zeroize` with a pointer that might potentially be null if the size is 0. Should we change this function to guarantee that it works in this case?

library/psa_crypto.c
```diff
@@ -1227,15 +1227,16 @@ static psa_status_t psa_copy_key_material( const psa_key_slot_t *source,
     buffer_size = PSA_KEY_EXPORT_MAX_SIZE( source->type,
                                            psa_get_key_bits( source ) );
     buffer = mbedtls_calloc( 1, buffer_size );
-    if( buffer == NULL )
+    if( buffer == NULL && buffer_size != 0 )
```
gilles-peskine-arm: The same idiom is present in `psa_save_generated_persistent_key` just above. Is “import/export persistent raw key: 0 byte” failing? If not why?

library/psa_crypto.c
Patater: Sounds like a separate bug outside of scope of this PR. Let's at least avoid making things worse in new code.

Patater: Filed https://github.com/ARMmbed/mbed-crypto/issues/49

library/psa_crypto.c
dgreen-arm: “import/export persistent raw key: 0 byte” isn't failing because it doesn't use `psa_save_generated_persistent_key`. Looking at the tests that do though (PSA generate persistent key), there doesn't appear to be one testing a zero bit raw key.

Adding one causes a failure and will require a bit of change to the test to add it properly.

dgreen-arm: Filed #50 to track this.

Patater:
CI has known-issue failures on FreeBSD timing test and DTLS 3d proxy test.

PR #52, opened 2019-02-12
all.sh: Enable verbose failure messages for CMake

Patater:
Set the CMake-observed variable `CTEST_OUTPUT_ON_FAILURE`, so that when
a "make test" run by CMake fails, verbose test output about the detail
of failure is available.


Patater:
CI failure is `DTLS proxy: 3d, gnutls server, fragmentation`, a known issue.

PR #53, opened 2019-02-12
Specification of how keys and the random seed are stored

gilles-peskine-arm:
Describe the storage format for keys and random seed in Mbed Crypto 0.1.0 released with Mbed OS 5.11, over C stdio and over ITS with 32-bit file identifiers.
    
Describe the proposed storage format for keys and random seed in the future release of Mbed Crypto for Mbed OS 5.12, over C stdio and over ITS with 64-bit file identifiers.


docs/architecture/mbed-crypto-storage-specification.md
```diff
@@ -0,0 +1,153 @@
...
+
+Assumption: C stdio, allowing names containing lowercase letters, digits and underscores, of length up to 23.
+
+An undocumented build-time configuration value `CRYPTO_STORAGE_FILE_LOCATION` allows storing the key files in a directory other than the current directory. This value is simply prepended to the file name (so it must end with a directory separator to put the keys in a different directory).
+
+* `CRYPTO_STORAGE_FILE_LOCATION "psa_key_slot_0"`: used as a temporary file. Must be writable. May be overwritten or deleted if present.
+* `sprintf(CRYPTO_STORAGE_FILE_LOCATION "psa_key_slot_%lu", key_id)` [content](#key-file-format-for-0.1.0) of the [key whose identifier](#key-names-for-0.1.0) is `key_id`.
+* Other files: unused.
+
+Mbed Crypto 0.2.0
```
Patater: This should be 1.0.0, following [semantic versioning](https://semver.org/). We are breaking backwards compatibility with 0.x.x versions in this release.

docs/architecture/mbed-crypto-storage-specification.md
```diff
@@ -0,0 +1,153 @@
...
+* [Persistent transparent keys](#key-file-format-for-0.2.0) designated by a [key identifier and owner](#key-names-for-0.2.0).
+* [Nonvolatile random seed](#nonvolatile-random-seed-file-format-for-0.2.0) on ITS only.
+
+Backward compatibility commitments: TBD
+
+### Key names for 0.2.0
+
+Information about each key is stored in a dedicated file whose name is constructed from the 32-bit key identifier (`psa_key_id_t`) and, if applicable, the owner identifier. The way in which the file name is constructed depends on the storage backend. The content of the file is described [below](#key-file-format-for-0.2.0).
+
+The valid values for a key identifier are the range from 1 to 0xfffeffff. The range is not documented.
```
Patater: "The range is not documented." but doesn't the previous sentence document it?

docs/architecture/mbed-crypto-storage-specification.md
```diff
@@ -0,0 +1,153 @@
...
+
+* Library integration: the key file name is just the key identifer. This is a 32-bit value.
+* PSA service integration: the key file name is `key_id << 32 | owner_uid` where `key_id` is the key identifier specified by the application and `owner_uid` is the calling partition identifier provided to the serve by the partition manager. This is a 64-bit value.
+
+### Key file format for 0.2.0
+
+The layout is identical to [0.1.0](#key-file-format-for-0.1.0) so far. However note that the encoding of key types, algorithms and key material has changed, therefore the storage format is not compatible (despite using the same version so far).
```
Patater: "using the same version" of what? should this be "using the same layout"? Or should we update the VERSION tag in the format?

docs/architecture/mbed-crypto-storage-specification.md
gilles-peskine-arm: Uh, no, according to https://semver.org/ and common conventions, 0.x is not expected to be compatible to 0.y for x ≠ y.

I don't know how we'll call the version of Mbed Crypto shipped with Mbed OS 5.12, but it won't be compliant with version 1.0 of the PSA API, so I don't think we should call it 1.0.

docs/architecture/mbed-crypto-storage-specification.md
gilles-peskine-arm: This is an internal architecture document. We don't document it, as in, we haven't made any promise to users.

docs/architecture/mbed-crypto-storage-specification.md
gilles-peskine-arm: I meant the version tag in the format, I'll clarify.

Since so far this is an experimental format that we break often, we haven't been updating the version tag yet.

docs/architecture/mbed-crypto-storage-specification.md
Patater: OK, I see there are special provisions for version 0 and also for pre-release versions. Since we aren't yet ready to provide a stable API, and we want to align version with the PSA Crypto specification where possible, we should avoid going 1.0.0 until the PSA Crypto specification does and we've caught up with it.

This specification is specific to the implementation, not the PSA Crypto specification, so it should use the Mbed Crypto version. Ideally these are aligned at least on their MAJOR numbers.

docs/architecture/mbed-crypto-storage-specification.md
Patater: We can be more clear about "it's not documented" since it is technically documented here in this internal architecture document. We can say "The range is an internal implementation detail not guaranteed to be stable between versions," or something similar.

docs/architecture/mbed-crypto-storage-specification.md
Patater: Let's keep the version tag where it's at until 1.0.0 and API stability.

gilles-peskine-arm: We may well commit to a key format before we've fully achieve API stability. At that point we'd increment the storage version to 1.

docs/architecture/mbed-crypto-storage-specification.md
gilles-peskine-arm: Not so much an implementation detail as an undocumented limitation. At this point, the specification says that the key ID is a 32-bit value and doesn't say that implementations are allowed to declare some values as invalid. Should they be? To be clarified in version 1.0 of the API specification.

docs/architecture/mbed-crypto-storage-specification.md
Patater: Yes, we might do. Let's keep it where it's at until 1.0.0 unless we have a good reason not to.

docs/architecture/mbed-crypto-storage-specification.md
```diff
@@ -0,0 +1,153 @@
...
+
+Assumption: ITS provides a 64-bit file identifier namespace. The Crypto service can use arbitrary file identifiers and no other part of the system accesses the same file identifier namespace.
+
+* Files 0 through 0xffffff51, 0xffffff53 through 0xffffffff: unused.
+* File 0xffffff52 (`PSA_CRYPTO_ITS_RANDOM_SEED_UID`): [nonvolatile random seed](#nonvolatile-random-seed-file-format-for-0.1.0).
+* Files 0x100000000 through 0xffffffffffff: [content](#key-file-format-for-0.2.0) of the [key whose identifier is the file identifier](#key-names-for-0.2.0).
+
+### File namespace on ITS as a library for 0.2.0
+
+Assumption: ITS provides a 64-bit file identifier namespace. The Crypto service can use arbitrary file identifiers and no other part of the system accesses the same file identifier namespace.
```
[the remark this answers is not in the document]
gilles-peskine-arm: fixed

docs/architecture/mbed-crypto-storage-specification.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Nonvolatile random seed file format for 0.1.0
+
+The nonvolatile random seed file contains a seed for the random generator. If present, it is rewritten at each boot as part of the random generator initialization.
+
+The file format is just the seed as a byte string with no metadata or encoding of any kind.
+
+### File namespace on ITS for 0.1.0
+
+Assumption: ITS provides a 32-bit file identifier namespace. The Crypto service can use arbitrary file identifiers and no other part of the system accesses the same file identifier namespace.
```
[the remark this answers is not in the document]
gilles-peskine-arm: Since the header doesn't say that it's about a specific integration, why do you expect it not to apply to all integrations?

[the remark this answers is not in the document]
itayzafrir: OK

PR #56, opened 2019-02-13
psa_constant_names: status is signed

gilles-peskine-arm:
Actually support negative values in `psa_constant_names` and `test_psa_constant_names`.

Needed for #55


Patater:
CI failure is FreeBSD timing test

PR #57, opened 2019-02-15
psa: Check generator validity before read

Patater:
Check generator validity (i.e. that alg has been initialized) before
allowing reads from the generator or allowing reads of the generator's
capacity.

This aligns our implementation with the documented error code behavior
in our crypto.h and the PSA Crypto API.

Fixes https://github.com/ARMmbed/mbedtls-psa/issues/183

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -3621,6 +3621,12 @@ psa_status_t psa_generator_abort( psa_crypto_generator_t *generator )
 psa_status_t psa_get_generator_capacity(const psa_crypto_generator_t *generator,
                                         size_t *capacity)
 {
+    if( generator->alg == 0 )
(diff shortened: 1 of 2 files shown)


Patater:
Rebasing to address review comments and to resolve merge conflicts.

Patater:
Rebased. Unfortunately, the base change had to happen in order to get the merge conflicts resolved, and we've merged the latest Mbed TLS code in (so the base is quite different).

Probably easiest to diff between diffs on this one instead of using GitHub's "force-pushed" diff.

Old branch up at https://github.com/Patater/mbed-crypto/tree/check-generator-validity-1

Patater:
CI failure is the FreeBSD timing test, a known flaky test.

Issue #10, opened 2019-01-10
Implement and test BAD_STATE in multipart operation setup


gilles-peskine-arm:
### Description

Setup functions for multipart operations should return `PSA_ERROR_BAD_STATE` if the operation structure is already in use. https://github.com/ARMmbed/mbed-crypto/pull/11 updates the documentation. This issue is for implementing this and adding corresponding negative tests.

### Issue request type

    [ ] Question
    [ ] Enhancement
    [X] Bug

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTCRYPT-631*

PR #58, opened 2019-02-15
Disallow use of invalid contexts

Patater:
Disallow use of MAC, hash, and cipher contexts that haven't been set up (i.e. has a `*_setup()` function called on them).


tests/suites/test_suite_psa_crypto.function
```diff
@@ -2262,6 +2286,12 @@ void mac_sign( int key_type_arg,
                                      actual_mac, mac_buffer_size,
                                      &mac_length ) );
 
+    /* Ensure double sign fails properly. */
+    TEST_EQUAL( psa_mac_sign_finish( &operation,
```
gilles-peskine-arm: Please avoid mixing input validation tests with operation correctness tests. That makes it harder to understand what is tested, and makes it harder to run just the input validation tests without also running the more costly operation correctness tests. This function is for operation correctness.

This also goes for `mac_verify`.

Also, why test `sign_finish` followed by `sign_finish`, but not other invalid combinations such as `sign_finish` followed by `update`, `sign_setup` followed by `verify_finish`, etc.?

library/psa_crypto.c
```diff
@@ -1502,8 +1502,8 @@ psa_status_t psa_hash_update( psa_hash_operation_t *operation,
             break;
 #endif
         default:
-            ret = MBEDTLS_ERR_MD_BAD_INPUT_DATA;
-            break;
+            psa_hash_abort( operation );
+            return( PSA_ERROR_BAD_STATE );
```
gilles-peskine-arm: I'm not sure if it's a good idea to call `abort` on a `BAD_STATE` error. What is likely to happen?

If the operation object was uninitialized or corrupted, `abort` has undefined behavior and this could be a way for an attacker to leverage a memory corruption error into remote code execution.

If the object was in a well-formed state, but not ready for an update, then for a hash that means that the object is already in a non-setup state and `abort` will do nothing.

Patater: The same reasoning seems to apply to `psa_hash_finish()` as well. I'll update there, too.

If implementation functions like `mbedtls_sha1_finish_ret()` and can ever return an error equivalent to `PSA_ERROR_BAD_STATE`, then we may have a pre-existing bug in `psa_hash_finish()` as we will call `psa_hash_abort()` in such a case.

Patater:
Rebasing to address review feedback, extend tests, and fix #10.

Patater:
Rebased. Previous branch at https://github.com/Patater/mbed-crypto/tree/disallow-invalid-context-1

Patater:
I'm looking into some memory leaks.

In general, if you start a cipher operation, and some internal memory is allocated on your behalf (e.g. AES context created), and then you call finish twice and get a `PSA_ERROR_BAD_STATE` error, there is no way to free that lost memory anymore, is there? You can't call abort, since after you get an `PSA_ERROR_BAD_STATE`, all bets are off.

NirSonnenschein:
wouldn't the first call to finish de-allocate the resources?

gilles-peskine-arm:
> after you get an PSA_ERROR_BAD_STATE, all bets are off.

That's not true. All bets are off if you have a bad state because the object was corrupted or never initialized. But if you've correctly initialized the object and subsequently only ever modified it through API functions, then abort is guaranteed to succeed.

Patater:
@NirSonnenschein Yes, it likely would. Good point.

I'll change my example. How about if you call cipher finish without first setting an IV? Finish would fail with `PSA_ERROR_BAD_STATE`, but memory would have been allocated during cipher setup.

Patater:
Thanks for the clarification, Gilles. I'll switch from using memset to all 0 between bad order tests and use abort instead.

Also note that one isn't guaranteed to get BAD_STATE if the context is corrupt or never initialized, so I'm not sure why we should be concerned, as per your comment: https://github.com/ARMmbed/mbed-crypto/pull/58#discussion_r257349820, about calling abort if BAD_STATE is encountered in these sorts of "well-formed" situations.

Can we conclude that it is OK to call abort on BAD_STATE? Are there cases we shouldn't call abort given we've gotten a BAD_STATE?

gilles-peskine-arm:
https://github.com/ARMmbed/mbed-crypto/pull/58#discussion_r257349820 was about calling abort internally, not about calling it from the application. I distinguished two cases: well-formed situations, where it's debatable whether the right behavior is to abort or to not modify the object, and non-well-formed situations, where abort could make things worse.

In application code, you should always initialize your memory, and you should always call abort.

Patater:
Rebase incoming to switch from memset to abort between bad order tests.

Patater:
Rebased. Previous branch at https://github.com/Patater/mbed-crypto/tree/disallow-invalid-context-3

Patater:
Forgot to apply a fixup commit. Rebasing again.

Patater:
Rebased to apply fixup commit. Previous branch at https://github.com/Patater/mbed-crypto/tree/disallow-invalid-context-4

Patater:
All done rebasing. Ready to review.

Patater:
Added commits to address https://github.com/ARMmbed/mbed-crypto/pull/58#discussion_r257349820

Will apply as fixup with a rebase if review goes well

Patater:
This is getting a bit too messy. I'm going to have to rebase before review.

Patater:
Rebased to apply fixups and make some style improvements. Previous branch at https://github.com/Patater/mbed-crypto/tree/disallow-invalid-context-5

Patater:
Rebasing done.

Patater:
Will rebase to apply fixup commit now

Patater:
Rebased to apply fixup and fix CI issues with `depend_hashes.pl`.

Patater:
Rebasing and rework done for now.

Patater:
This has CI failures. I'll be doing some rework to address this.

library/psa_crypto.c
```diff
@@ -1379,7 +1379,13 @@ psa_status_t psa_hash_setup( psa_hash_operation_t *operation,
                              psa_algorithm_t alg )
 {
     int ret;
-    operation->alg = 0;
+
+    /* A context must be freshly initialized before it can be set up. */
+    if (operation->alg != 0)
```
gilles-peskine-arm: Style (here and in a few other places):
```suggestion
    if ( operation->alg != 0 )
```

gilles-peskine-arm:
I've given the PR a first glance and it looks like it's going in the right direction. I'll hold doing a detailed code review until you've finished the rework.

Patater:
We've uncovered pre-existing bugs where we didn't initialize the operation context before calling setup. Rebasing to add commits to fix these issues.

Patater:
Rebased to add commits to fix uninitialized context bugs and edit commits with bad style. Previous branch at https://github.com/Patater/mbed-crypto/tree/disallow-invalid-context-8

library/cipher.c
```diff
@@ -1236,7 +1236,7 @@ int mbedtls_cipher_crypt( mbedtls_cipher_context_t *ctx,
             (mbedtls_cipher_context_psa *) ctx->cipher_ctx;
 
         psa_status_t status;
-        psa_cipher_operation_t cipher_op;
+        psa_cipher_operation_t cipher_op = PSA_CIPHER_OPERATION_INIT;
```
gilles-peskine-arm: Why wasn't this caught by CI? Are we missing a test case or a test configuration?

library/psa_crypto.c
```diff
@@ -2238,6 +2248,11 @@ psa_status_t psa_mac_sign_finish( psa_mac_operation_t *operation,
 {
     psa_status_t status;
 
+    if( operation->alg == 0)
```
gilles-peskine-arm: Style: missing space before `)` (here and elsewhere)

Patater:
CI passed. Ready for review.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -1990,32 +1999,85 @@ exit:
...
 
-    /* psa_hash_update without calling psa_hash_setup beforehand */
-    memset( &operation, 0, sizeof( operation ) );
+    /* Call setup twice in a row. */
+    PSA_ASSERT( psa_hash_setup( &operation, alg ) );
+    TEST_EQUAL( psa_hash_setup( &operation, alg ),
+                PSA_ERROR_BAD_STATE );
+    PSA_ASSERT( psa_hash_abort( &operation ) );
+
+    /* Call update without calling setup beforehand. */
```
gilles-peskine-arm: Technically, you're calling update after abort. This is equivalent to update without setup only if abort behaves correctly.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -1960,6 +1961,14 @@ void hash_operation_init( )
 
     memset( &zero, 0, sizeof( zero ) );
 
+    /* A default hash operation should not be usable. */
```
gilles-peskine-arm: What does “default hash operation” mean? I understand from context, but this is the code explaining the comment whereas it should be the other way round. Suggestion: “freshly-initialized”

tests/suites/test_suite_psa_crypto.function
```diff
@@ -2309,6 +2508,12 @@ void mac_verify( int key_type_arg,
                                        expected_mac->x,
                                        expected_mac->len ) );
 
+    /* Ensure double verify fails properly. */
```
gilles-peskine-arm: Leftover from a previous iteration.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -1990,32 +1999,85 @@ exit:
...
         0xa4, 0x95, 0x99, 0x1b, 0x78, 0x52, 0xb8, 0x55 };
+    unsigned char hash[sizeof(valid_hash)] = { 0 };
     size_t hash_len;
     psa_hash_operation_t operation = PSA_HASH_OPERATION_INIT;
 
     PSA_ASSERT( psa_crypto_init( ) );
 
-    /* psa_hash_update without calling psa_hash_setup beforehand */
-    memset( &operation, 0, sizeof( operation ) );
+    /* Call setup twice in a row. */
```
gilles-peskine-arm: This is getting to be a long series of ad hoc tests. Have we reached the point where they would be easier to write and easier to read if we made the code data-driven? That would make it easier to add other sequences.
```
for(c in string) {
    switch(c) {
    case 'a': ret = abort();
    case 'u': ret = update();
    ...
    }
    TEST_EQUAL(ret, c is last ? PSA_SUCCESS : expected_last_status);
}
abort();
setup(); update(); finish(); // test that the operation object is usable after abort
```

library/cipher.c
Patater: This was caught by CI, but it wasn't a failure until we started checking that the context is initialized before allowing setup() (by fixing #10). Same with the X.509 issues. all.sh catches this when it runs "make test" when using the `MBEDTLS_USE_PSA_CRYPTO` option.

library/psa_crypto.c
Patater: Gah, I thought I fixed all of these...

tests/suites/test_suite_psa_crypto.function
Patater: Would you recommend we call both abort and memset the context to 0? I'm depending on abort setting the context to 0, and if it doesn't, the test will fail.

tests/suites/test_suite_psa_crypto.function
Patater: I've also seen us use "blank". I'll change to freshly-initialized.

tests/suites/test_suite_psa_crypto.function
Patater: Removed

tests/suites/test_suite_psa_crypto.function
gilles-peskine-arm: If abort works properly, it doesn't matter. It's only a matter of how hard it will be to investigate a test failure that is ultimately due to abort returning the expected status but not having done its job properly.

I would prefer to make each unit test independent. But it's also good to have a stress test where a single object is reused.

tests/suites/test_suite_psa_crypto.function
Patater: I did have this thought as well, but decided it wasn't worth the time yet.

Patater:
Rebasing to address review feedback.

tests/suites/test_suite_psa_crypto.function
Patater: I've raised #66 to add targeted abort tests.

Patater:
Rebased. Previous branch at https://github.com/Patater/mbed-crypto/tree/disallow-invalid-context-9

Patater:
Rebased to add a commit that returns bad state without aborting in more cases.

library/ssl_tls.c
```diff
@@ -6529,7 +6529,7 @@ static void ssl_calc_finished_tls_sha256(
     unsigned char padbuf[32];
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     size_t hash_size;
-    psa_hash_operation_t sha256_psa;
+    psa_hash_operation_t sha256_psa = PSA_HASH_OPERATION_INIT;
```
yanesca: Using `PSA_HASH_OPERATION_INIT`, `psa_hash_operation_init();` or simply `= 0` are all valid ways to do initialisation, right?

What are the advantages of different methods? In particular, why are we using on line 1449 and 1492 `psa_hash_operation_init();` and `PSA_HASH_OPERATION_INIT` here?

gilles-peskine-arm: `{0}` and `INIT` (macro) can be used to initialize global variables, but `init()` (function) can't. `{0}` is generic but some compilers warn that this implicitly initializes other fields to 0. `init()` is the only one that's typed, so you'll get an error if you use it in the wrong context. `memset` to 0 is also defined as equivalent because it allows applications to use e.g. `calloc`, and in practice it produces the same value as `{0}` on virtually all platforms.

In most cases, the choice is indifferent.

Patater: `PSA_HASH_OPERATION_INIT` is all static, so you can use it for static globals. The inline function can be used where functions can be used, which is a few less places, but it uses the static initializer under-the-hood. `= {0}` is usually ok, but some compilers will complain about missing field initializers.

I personally prefer using the macro everywhere, but previous authors inside library/ssl_tls.c liked the function. There is no functional difference in this case.

Patater:
CI failure is only FreeBSD timing test, a known-flaky test.

PR #59, opened 2019-02-18
Support key file IDs encoding the key owner

gilles-peskine-arm:
Add support for key file identifier that encode the key owner. Use the type `psa_key_file_id_t` to refer to key files, as opposed to `psa_key_id_t` which is the key identifier specified by the application. The two types are identical except when building the library for a multi-client service (currently: only the PSA crypto service, under the build option `MBEDTLS_PSA_CRYPTO_SPM`).

Because `psa/crypto.h` and `psa/crypto_types.h` are part of the source of the PSA Crypto API specification, use the standard types there. Define `psa_key_file_id_t` in `psa/crypto_platform.h` depending on whether `MBEDTLS_PSA_CRYPTO_SPM` is defined. We can make this cleaner by using `psa_key_file_id_t` in `psa/crypto.h` once the specification is separated from the implementation.

The file storage backend, and all of the tests, do not support `MBEDTLS_PSA_CRYPTO_SPM`. I have manually checked that the build with `MBEDTLS_PSA_CRYPTO_SPM` and `MBEDTLS_PSA_STORAGE_ITS_C` works with dummy `psa/error.h` and `psa/internal_trusted_storage.h`. Testing without a board would require updating and mergin  https://github.com/ARMmbed/mbedtls-psa/pull/204 .


library/psa_crypto_slot_management.c
```diff
@@ -236,7 +244,7 @@ static psa_status_t psa_internal_make_key_persistent( psa_key_handle_t handle,
 }
 
 static psa_status_t persistent_key_setup( psa_key_lifetime_t lifetime,
-                                          psa_key_id_t id,
+                                          psa_key_file_id_t id,
```
Patater: Is changing `psa_key_id_t` to `psa_key_file_id_t` everywhere really the best option? 

If we keep this PR as-is and then do the following, we wouldn't require changing to `psa_key_file_id_t` everywhere. Would this be a workable improvement?
- Remove `typedef uint32_t psa_key_id_t;` from crypto_types.h
- Add `typedef psa_app_key_id_t psa_key_id_t` to crypto_platform.h for non-SPM platforms

Then we'd have the following situation:
- Using `psa_key_file_id_t` as `psa_key_id_t` on SPM platforms (what we sort of have in this PR now, but don't use it)
- Using `psa_app_key_id_t `as `psa_key_id_t` on non-SPM platforms

Maybe there is a reason to keep defining psa_key_id_t in crypto_types.h that I'm not seeing. In which case, we can keep that as-is and actually use that type definition instead of `psa_key_file_id_t` everywhere.

gilles-peskine-arm: Currently, `crypto_types.h` and `crypto.h` are used to generate the specification, so they have to use the standard definitions. I wanted to minimize the incursion of MbedCrypto-specific stuff there.

Once the specification and the implementation are fully split (which should take a couple of weeks), I think we should define `psa_key_file_id_t` in `crypto_types.h` and use it in `crypto.h`.

include/psa/crypto_platform.h
```diff
@@ -49,4 +49,24 @@
...
+
+#if defined(MBEDTLS_PSA_CRYPTO_SPM)
+/* When the library is built as part of a PSA Cryptography service on a
+ * PSA platform, a key file ID encodes both the 32-bit key ID used by the
+ * application and the signed 32-bit partition ID of the key owner. */
+typedef struct
+{
+    int32_t owner;
+    uint32_t key_id;
+} psa_key_file_id_t;
```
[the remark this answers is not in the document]
gilles-peskine-arm: Indeed, putting the key id first saves a few instructions. From `make CC=arm-linux-gnueabi-gcc CFLAGS='-Os -mthumb'` (not exactly v7m but should be close enough):
```
-rw-rw-r-- 1 gilpes01 gilpes01 1948 Feb 18 20:43 library.spm/key_owner/psa_crypto_storage_its.o
-rw-rw-r-- 1 gilpes01 gilpes01 1964 Feb 18 20:44 library.spm/owner_key/psa_crypto_storage_its.o
```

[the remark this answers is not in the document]
Patater: I don't think it's so Arm specific. If the binary representation is the same, the program doesn't have to swap the order around before passing the 64-bit value on to ITS and may be able to use the binary representation as-is, assuming alignment works out and so forth.

include/psa/crypto_types.h
```diff
@@ -90,7 +90,15 @@ typedef uint32_t psa_key_lifetime_t;
 
 /** Encoding of identifiers of persistent keys.
  */
+/* Mbed Crypto particularity: in an SPM integration, psa_key_id_t is a larger
+ * type that encodes both the application-supplied key identifier and the
+ * identity of the owner of the key. This non-compliant implementation is only
+ * used when building Mbed Crypto as a library inside the PSA Cryptography
+ * service. Applications that call the service see the standard API with a
+ * 32-bit type. */
+#if !defined(MBEDTLS_PSA_CRYPTO_SPM)
```
itayzafrir: `MBEDTLS_PSA_CRYPTO_SPM` is defined for the entire secure side. This means that if another _secure_ partition will try to communicate with crypto, this client will see `psa_key_id_t` as the full struct (i.e. `psa_key_file_id_t`) as not as `uint32_t`.

itayzafrir: This could work if the ifdefs in the header files are tested for `PSA_CRYPTO_SECURE` instead of `MBEDTLS_PSA_CRYPTO_SPM`.

gilles-peskine-arm: Ok, I'll change that.

gilles-peskine-arm:
I'm going to update this PR with a new commit history that splits up the original large commit that both renames `psa_key_id_t` and adds owner-in-key-id support.

include/psa/crypto_platform.h
```diff
@@ -63,7 +63,6 @@ typedef struct
 } psa_key_file_id_t;
 typedef psa_key_file_id_t psa_key_id_t;
 #define PSA_KEY_FILE_GET_KEY_ID( file_id ) ( ( file_id ).key_id )
-#endif
```
gilles-peskine-arm: Stray leftover from a partial commit, I'll fix this in a rebase.

library/psa_crypto_storage_its.c
```diff
@@ -36,12 +36,26 @@
...
 
-static psa_storage_uid_t psa_its_identifier_of_slot( psa_key_id_t key )
+static psa_storage_uid_t psa_its_identifier_of_slot( psa_key_file_id_t file_id )
 {
-    return( key );
+#if defined(MBEDTLS_PSA_CRYPTO_SPM)
+    /* Encode the owner in the upper 32 bits. This means that if
+     * owner values are nonzero (as they are on a PSA platform),
+     * key files don't clash with file names such as the random seed
+       file which have an identifier in the range 0..0xffffffff. */
```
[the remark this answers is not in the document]
gilles-peskine-arm: Fixed in the new history.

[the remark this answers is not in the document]
gilles-peskine-arm: Reworded in the new history.

include/psa/crypto_platform.h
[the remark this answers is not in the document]
gilles-peskine-arm: I put `key_id` first in the rebased version. It's likely to be neutral or a tiny improvement on any little-endian platform. How much it saves depends on the ABI and of course on what the compiler makes of the code.

gilles-peskine-arm:
I have pushed an updated version with a new, more granular history. The differences:

* There is now a compilation option `MBEDTLS_PSA_CRYPTO_KEY_FILE_ID_ENCODES_OWNER` to gate the nonstandard definition of `psa_key_file_id_t`. This makes the code easier to follow than depending only on `PSA_CRYPTO_SECURE`. **The crypto service must enable this option**.
* Use `PSA_CRYPTO_SECURE` instead of `MBEDTLS_PSA_CRYPTO_SPM` to detect that the service is being built, as noted by Itay.
* Put the key ID first in `psa_key_file_id_t`.
* Some documentation improvements.

gilles-peskine-arm:
I made another, minor rebase to fix some issues reported by the CI:

* An option in `config.h` must not have Doxygen formatting if it isn't enabled by `scripts/config.pl all`.
* I'd forgotten to run `generate_version_features.pl` after adding an option to `config.h`.

Patater:
This needs CI to pass before we can merge it.

gilles-peskine-arm:
I fixed a typo in “Support encoding an owner in key file IDs” which should let the CI do its job.

gilles-peskine-arm:
> I still am not a fan of changing `psa_key_id_t` to `psa_key_file_id_t` in so many places.

Neither am I, but I'm even less a fan of writing `psa_key_id_t` where we mean something that isn't a key id.


> Let's revisit after the specification is split out from the implementation.

Actually, at that point, we'll change more occurrences of `psa_key_id_t` to `psa_key_file_id_t`, in `crypto.h`.

gilles-peskine-arm:
I fixed _another_ typo in “Support encoding an owner in key file IDs” which should let the CI do its job this time.

Patater:
Minor style: Now the alignment is off, since `key_id` and `file_id` don't have the name number of characters.

gilles-peskine-arm:
Since I've been rebasing anyway, I'll rebase again to fix the alignment.

gilles-peskine-arm:
Aligned.

PR #61, opened 2019-02-19
travis: Use seed 4 when running ssl-opt.sh

Patater:
Seed 4 has been shown to result in a DTLS proxy that works more often
than not. This should help reduce the flakiness we observe from Travis
CI runs.


Patater:
Trust our luck that we hit the FreeBSD timing test flakines on this PR which reduces flakiness elsewhere...

That's the only CI failure on this PR now.

PR #62, opened 2019-02-19
Improve multipart cipher tests: 2 blocks, CTR

gilles-peskine-arm:
The only cipher multipart tests we had were for CBC with exactly one block. Add tests with CBC with two blocks and tests with CTR.


Patater:

PR #64, opened 2019-02-20
Update the architecture document regarding key file identifiers

gilles-peskine-arm:
Switch to the terminology "key file identifier", as has been done in the code.

The owner uid is now in the upper 32 bits of the key file identifier, which facilitates namespacing.

This updates the architecture document to what we discussed in email and implemented in #32 and #59.


Patater:

PR #65, opened 2019-02-20
psa: Be compatible with deprecated constants

Patater:
In case the new constants aren't available yet in Mbed TLS, continue to
use the deprecated constants if they are available.

This breaks a cyclic dependency between Mbed Crypto needing new constants from Mbed TLS as well as Mbed TLS needing new PSA Crypto before either can ship into Mbed OS.


Patater:
Only CI failure on Jenkins CI is FreeBSD Timing test, a known-flaky test.

Jenkins CI is passing, which is more thorough than the Travis CI.

PR #67, opened 2019-02-21
psa: Add backwards compatible error codes

Patater:
Add deprecated error codes to help transition between the previous
version of the PSA Crypto specification and the current one.


Patater:
This is having issues with psa_constant_names right now, due to duplicate cases getting generated.

Patater:
OK to review.

CI is failing a TLS-using-PSA for chachapoly test. Unsure why TLS is trying to use PSA for chachapoly right now, since that's not supported in Mbed Crypto yet. Unsure why we didn't see this previously.

Patater:
Discussed with @hanno-arm. He's seen this before and it's a known flaky test, not due to use of chachapoly, but due to the alert and client hello messages getting coalesced. The server expects the alert and handles it, but never sees the subsequent client hello.

PR #70, opened 2019-02-25
psa: Explicitly include platform_util.h

Patater:
crypto_extra.h has a dependency on platform_util.h for
MBEDTLS_DEPRECATED_NUMERIC_CONSTANT. Make the dependency explicit by
including platform_util.h. Although in most use cases the header should
already be included by something else, it doesn't hurt to include it
again and helps to clarify dependencies.


Patater:
CI failure is only the FreeBSD timing test, a known flaky test.

PR #72, opened 2019-02-25
Fix missing cleanup in psa_cipher_setup

### LINKED ISSUE #68: psa_cipher_setup leaves operation in partially initialized state in case of error
### Description

in some of the setup functions the operation structure is not properly cleaned up in case of failure.
we specifically encountered this in the psa_cipher _setup. When passing in an invalid key handle (with other valid parameters) it will fail but leave the operation structure with the algorithm set.
This is OS and platform independent behavior.

Note: this likely also impacts other setup functions (e.g. psa_hmac_setup_internal, psa_mac_setup, etc..)

### Issue request type

    [ ] Question
    [ ] Enhancement
    [X] Bug

gilles-peskine-arm:
Fix #68.

Add non-regression tests: in `hash_setup`, `mac_setup` and `cipher_setup`, after a failed setup or a successful setup+abort, call setup again with valid parameters and check that it succeeds.

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -2902,8 +2902,8 @@ static psa_status_t psa_cipher_setup( psa_cipher_operation_t *operation,
                                       psa_algorithm_t alg,
                                       mbedtls_operation_t cipher_operation )
 {
-    int ret = MBEDTLS_ERR_CIPHER_FEATURE_UNAVAILABLE;
-    psa_status_t status;
+    int ret = 0;
+    psa_status_t status = PSA_ERROR_GENERIC_ERROR;
     psa_key_slot_t *slot;
     size_t key_bits;
     const mbedtls_cipher_info_t *cipher_info = NULL;
@@ -2923,19 +2923,19 @@ static psa_status_t psa_cipher_setup( psa_cipher_operation_t *operation,
 
     status = psa_get_key_from_slot( handle, &slot, usage, alg);
     if( status != PSA_SUCCESS )
-        return( status );
+        goto exit;
     key_bits = psa_get_key_bits( slot );
 
     cipher_info = mbedtls_cipher_info_from_psa( alg, slot->type, key_bits, NULL );
     if( cipher_info == NULL )
-        return( PSA_ERROR_NOT_SUPPORTED );
+    {
+        status = PSA_ERROR_NOT_SUPPORTED;
+        goto exit;
+    }
 
     ret = mbedtls_cipher_setup( &operation->ctx.cipher, cipher_info );
     if( ret != 0 )
-    {
-        psa_cipher_abort( operation );
-        return( mbedtls_to_psa_error( ret ) );
-    }
+        goto exit;
 
 #if defined(MBEDTLS_DES_C)
     if( slot->type == PSA_KEY_TYPE_DES && key_bits == 128 )
@@ -2956,10 +2956,7 @@ static psa_status_t psa_cipher_setup( psa_cipher_operation_t *operation,
                                      (int) key_bits, cipher_operation );
     }
     if( ret != 0 )
-    {
-        psa_cipher_abort( operation );
-        return( mbedtls_to_psa_error( ret ) );
-    }
+        goto exit;
 
 #if defined(MBEDTLS_CIPHER_MODE_WITH_PADDING)
     switch( alg )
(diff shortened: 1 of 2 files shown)


library/psa_crypto.c
```diff
@@ -2992,7 +2986,12 @@ static psa_status_t psa_cipher_setup( psa_cipher_operation_t *operation,
         operation->iv_size = PSA_BLOCK_CIPHER_BLOCK_SIZE( slot->type );
     }
 
-    return( PSA_SUCCESS );
+exit:
+    if( status == 0 )
+        status = mbedtls_to_psa_error( ret );
+    if( status != 0 )
+        psa_cipher_abort( operation );
+    return( status );
```
Patater: #68 suspected other setup functions might have issues with leaving around partially initialized contexts. Have our tests proven that they don't? It looks like the tests we've added in this PR just make sure the context is re-usable, not that we don't leave a partially initialized context around if setup fails.

NirSonnenschein: some things to look at: 
psa_cmac_setup - in this function if mbedtls_cipher_setup succeeds and mbedtls_cipher_cmac_starts fails
psa_generator_hkdf_setup - if psa_hmac_setup_internal succeeds but subsequesnt call to psa_hash_update fails.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -1988,9 +2116,16 @@ void hash_setup( int alg_arg,
     PSA_ASSERT( psa_crypto_init( ) );
 
     status = psa_hash_setup( &operation, alg );
-    psa_hash_abort( &operation );
     TEST_EQUAL( status, expected_status );
 
+    if( status == PSA_SUCCESS )
+        PSA_ASSERT( psa_hash_abort( &operation ) );
```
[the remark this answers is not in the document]
gilles-peskine-arm: Good point. I've changed the {failed-setup; successful-setup} sequences to {failed-setup; abort; failed-setup; successful-setup}.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -120,6 +203,64 @@ static int construct_fake_rsa_key( unsigned char *buffer,
...
+                        size_t key_length,
+                        psa_algorithm_t alg,
+                        psa_mac_operation_t *operation,
+                        psa_status_t *status )
+{
+    psa_key_handle_t handle = 0;
+    psa_key_policy_t policy = PSA_KEY_POLICY_INIT;
+
+    PSA_ASSERT( psa_allocate_key( &handle ) );
+    psa_key_policy_set_usage( &policy, PSA_KEY_USAGE_SIGN, alg );
```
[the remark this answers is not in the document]
gilles-peskine-arm: We have coverage with different policies in `mac_policy`.

Since this function only tests `psa_mac_sign_setup` and not `psa_mac_verify_setup`, it only needs a `USAGE_SIGN` policy, just like the corresponding cipher test calls `cipher_encrypt_setup` with a `USAGE_ENCRYPT` policy. I think having both `USAGE_SIGN` and `USAGE_VERIFY` here was a leftover from the old days when MAC functions didn't check policies correctly.

library/psa_crypto.c
gilles-peskine-arm: The tests `hash_setup`, `mac_setup` and `cipher_setup` check that the context is reusable. That's the only black-box way of testing that the context is properly reinitialized. There's no black-box way to test that a context isn't “partially initialized”: it's either sufficiently pristine for setup or not.

I don't think there's a way to test the case of `psa_cmac_setup` where `mbedtls_cipher_setup` succeeds and `mbedtls_cipher_cmac_starts` fails at the moment. An invalid key size is detected earlier, and an unsupported algorithm for CMAC can't be reached until we add support for Camellia and ARIA to `psa_crypto` (and at that point we may well allow them with CMAC).

I've deliberately not touched KDF code because the KDF API has upcoming changes and these bad-case tests will need to change accordingly.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -120,6 +203,74 @@ static int construct_fake_rsa_key( unsigned char *buffer,
...
+    psa_destroy_key( handle );
+    return( 1 );
+
+exit:
+    psa_destroy_key( handle );
+    return( 0 );
+}
+
+int exercise_cipher_setup( psa_key_type_t key_type,
+                        const unsigned char *key_bytes,
```
Patater: Style: funny alignment here

gilles-peskine-arm: I rewrote the history to fix the indentation in the offending commit.

PR #74, opened 2019-03-01
Break non-crypto dependencies

Patater:
This is a portion of #71 

The goal is only to break dependencies on TLS and X.509 and NET, without removing those modules yet.


tests/scripts/all.sh
```diff
@@ -207,13 +196,6 @@ General options:
 Tool path options:
      --armc5-bin-dir=<ARMC5_bin_dir_path>       ARM Compiler 5 bin directory.
      --armc6-bin-dir=<ARMC6_bin_dir_path>       ARM Compiler 6 bin directory.
-     --gnutls-cli=<GnuTLS_cli_path>             GnuTLS client executable to use for most tests.
```
gilles-peskine-arm: Should we recognize, but ignore these options for the time being, for the sake of facilitating CI scripts that apply to both before and after or that apply to both Mbed Crypto and Mbed TLS?

tests/scripts/all.sh
```diff
@@ -868,53 +703,16 @@ component_build_no_std_function () {
...
-
-component_build_no_ssl_cli () {
-    msg "build: full config except ssl_cli.c, make, gcc" # ~ 30s
-    scripts/config.pl full
-    scripts/config.pl unset MBEDTLS_SSL_CLI_C
-    make CC=gcc CFLAGS='-Werror -Wall -Wextra -O0'
-}
-
 component_build_no_sockets () {
+    # XXX Questionable we want to test this. May be redundant.
```
gilles-peskine-arm: I think this component should be removed. It's not as if there was a real risk that the crypto code would inadvertently start using sockets. This component was introduced by 009a2640f673447aedbad356337c0facc143b059 and at the time it was the only component with `-std=c99` (later `-pedantic`), but now we have others.

tests/scripts/all.sh
```diff
@@ -1216,20 +991,6 @@ component_test_valgrind () {
 
     msg "test: main suites valgrind (Release)"
     make memcheck
-
-    # Optional part(s)
-    # Currently broken, programs don't seem to receive signals
-    # under valgrind on OS X
-
-    if [ "$MEMORY" -gt 0 ]; then
```
gilles-peskine-arm: With this removal, the `MEMORY` variable and the `--memory` option are no longer used.

tests/scripts/check-generated-files.sh
```diff
@@ -65,6 +65,6 @@ check()
 }
 
 check scripts/generate_errors.pl library/error.c
-check scripts/generate_query_config.pl programs/ssl/query_config.c
+check scripts/generate_query_config.pl programs/test/query_config.c
```
gilles-peskine-arm: Feels more like a `util` program than a `test` program to me, like `strerror`.

programs/psa/crypto_examples.c
```diff
@@ -1,23 +1,15 @@
 #include "psa/crypto.h"
 #include <string.h>
-
-#if defined(MBEDTLS_PLATFORM_C)
-#include "mbedtls/platform.h"
```
gilles-peskine-arm: All example programs include `mbedtls/platform.h` and use indirections like `mbedtls_printf` and `MBEDTLS_EXIT_FAILURE`. Why did you remove these indirections, but only in `programs/psa/*.c` and not in other example/utility programs?

library/x509_crt.c
```diff
@@ -101,10 +101,6 @@ typedef struct {
  */
 const mbedtls_x509_crt_profile mbedtls_x509_crt_profile_default =
 {
-#if defined(MBEDTLS_TLS_DEFAULT_ALLOW_SHA1_IN_CERTIFICATES)
```
gilles-peskine-arm: It doesn't matter that much since this file is going to be removed in mbed-crypto, but this commit makes no sense. First, precisely because this file is going to be removed in mbed-crypto. Second, because if this file wasn't going to be removed, this is wrong: it breaks applications that use the (deprecated, but supported) option `MBEDTLS_TLS_DEFAULT_ALLOW_SHA1_IN_CERTIFICATES`, since they will no longer allow SHA-1 in certificates.

It may be time to remove this deprecated option, but that's a decision for the Mbed TLS project, not for the Mbed Crypto project, and it should go together with removing the option from `config.h`, not just breaking one crucial part of it.

programs/psa/crypto_examples.c
Patater: The plan is to make platform.h an internal header when we can. PSA programs can set a good example and avoid using platform.h. Other programs could remove platform.h as well, but they are still Mbed TLS examples and use Mbed TLS APIs which we also want to make internal when we can.

library/x509_crt.c
Patater: The commit is present, even if we are removing it, because X.509 and TLS are removed in separate commits, TLS being removed first. We still want to be able to build and have tests pass in intermediate commits, so ether X.509 has to stop depending on TLS before it is removed, or we remove TLS, X.509, and NET all in one commit.

tests/scripts/check-generated-files.sh
Patater: My thinking was the only program that uses this C file is in test, so it's placed there. Could move to util if you feel strongly about this, but it isn't built as a standalone util (it was previously part of ssl_client2 and ssl_server2 and the test program, but never a standalone util).

tests/scripts/all.sh
Patater: I left options like those in, but ignored, so that our CI can still use those options and test older versions of our software.

tests/scripts/all.sh
Patater: Yes, these need to be handled. They don't need to be documented.

programs/psa/crypto_examples.c
gilles-peskine-arm: I thought the point of these abstractions was that if you port the library to a system where they're needed, you also get the example programs. Why would this no longer be a desirable property?

Note that I can see why we'd want PSA demonstration programs to be free of `mbedtls` dependencies. But that's out of scope of this PR.

library/x509_crt.c
gilles-peskine-arm: I don't see anything wrong with removing TLS and X509 in the same commit. Or removing TLS except for (some of) its configuration, then removing X509, then removing stale configuration options. But I'm against having a major functionality break in the history, as in, a clearly advertised feature of the library still compiles but no longer has the intended effect.

tests/scripts/check-generated-files.sh
gilles-peskine-arm: I don't feel strongly enough to block the PR on it, but your reasoning doesn't make sense to me. `query_config` itself isn't used for testing anything in Mbed Crypto, so why would it be in `test`? If the only reason it's there is that it's being tested, that's not actually a reason to have it and it should be removed. But it's also useful for other purposes, namely to report build information as part of a test campaign or a platform integration. That makes it a `util`.

Patater:
Rebased to address review feedback in original commits. More rebasing expected before this is ready to review again.

Previous branch at break-non-crypto-dependencies-1

programs/psa/crypto_examples.c
Patater: The property is desirable, but the cost of maintaining the abstraction is not. There are ways to use the linker to replace e.g. malloc with a custom replacement (gcc has `-Wl,-wrap,malloc`, armcc, msvc, and IAR have other mechanisms). We don't need to maintain any macros to support this.

Patater:
Rebased to ignore `--memory` type options. Still work outstanding on removing TLS dependency from X.509.

programs/psa/crypto_examples.c
Patater: I agree that removing the platform module and our use of it externally is outside the scope of this PR. We can do that in a separate PR.

library/x509_crt.c
Patater: OK, we'll move to removing TLS, X509, and NET in one commit since that means less noise like this.

Patater:
I'm removing the oid and X.509 detangling commits as well, since we can get those from Mbed TLS development once merged there.

tests/scripts/all.sh
Patater: Removed

tests/scripts/check-generated-files.sh
Patater: If it built on its own, I'd agree, it'd be a `util`. As-is, it's just a C source file and doesn't have a `main()` function.

gilles-peskine-arm: I still think it's a `util` but that's not important. Ok.

tests/scripts/all.sh
```diff
@@ -310,29 +291,32 @@ pre_parse_command_line () {
...
-            --no-memory) MEMORY=0;;
-            --openssl) shift; OPENSSL="$1";;
-            --openssl-legacy) shift; OPENSSL_LEGACY="$1";;
-            --openssl-next) shift; OPENSSL_NEXT="$1";;
+            --no-memory) ;;
+            --openssl) shift;;
+            --openssl-legacy) shift;;
+            --openssl-next) shift;;
             --out-of-source-dir) shift; OUT_OF_SOURCE_DIR="$1";;
             --random-seed) unset SEED;;
```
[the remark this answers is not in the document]
Patater: Removed

tests/scripts/all.sh
```diff
@@ -619,36 +565,6 @@ component_test_ref_configs () {
     record_status tests/scripts/test-ref-configs.pl
 }
 
-component_test_sslv3 () {
```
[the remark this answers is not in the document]
Patater: Removed

include/mbedtls/asn1.h
```diff
@@ -66,7 +66,6 @@
  * - 0x02 -- tag indicating INTEGER
  * - 0x01 -- length in octets
  * - 0x05 -- value
- * Such sequences are typically read into \c ::mbedtls_x509_buf.
```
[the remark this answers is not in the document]
Patater: Re-added with "Mbed TLS's"

tests/scripts/all.sh
```diff
@@ -605,12 +557,6 @@ component_test_default_cmake_gcc_asan () {
 
     msg "test: main suites (inc. selftests) (ASan build)" # ~ 50s
     make test
-
-    msg "test: ssl-opt.sh (ASan build)" # ~ 1 min
```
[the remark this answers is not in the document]
Patater: I had done this in my commit that removed the files, which I've cherry-picked into this PR.

include/mbedtls/ecp.h
```diff
@@ -365,19 +365,19 @@ mbedtls_ecp_keypair;
...
  *
- *                  This only applies to functions whose documentation
- *                  mentions they may return #MBEDTLS_ERR_ECP_IN_PROGRESS (or
- *                  #MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS for functions in the
- *                  SSL module). For functions that accept a "restart context"
- *                  argument, passing NULL disables restart and makes the
- *                  function equivalent to the function with the same name
+ *                  This only applies to functions whose documentation mentions
+ *                  they may return #MBEDTLS_ERR_ECP_IN_PROGRESS (or
+ *                  `MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS` for functions in the
```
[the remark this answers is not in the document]
Patater: The commit is only to remove the dependency in the ecp module. A later commit that removes SSL options from config.h will take care to update other references.

Patater:
Rebased to address feedback and hopefully fix the Windows build.

Patater:
Force push to update visualc files...

Patater:
This needs a rebase to resolve conflicts, plus CI and re-review.

Patater:
Rebased to latest crypto/development to resolve conflicts. Previous branch at https://github.com/Patater/mbed-crypto/tree/break-non-crypto-dependencies-12.5. No changes other than conflict resolution.

gilles-peskine-arm:
The CI failures are due to a test script that calls `compat.sh` and `ssl-opt.sh` and needs to be updated (I believe that'll be fixed when https://github.com/ARMmbed/mbedtls-test/pull/56 is merged) and one failure of the FreeBSD timing test which is a known issue. So this PR is good to merge once the offending test script is updated.

Patater:
Retested after merging https://github.com/ARMmbed/mbedtls-test/pull/56 and all is good.

PR #80, opened 2019-03-05
Remove Diffie-Hellman examples

yanesca:
These examples rely on the NET module, which we want to remove. In
theory we could remove just the dependency, but we decided to remove the
whole example because:

 - They showcase some bad crypto: custom, undocumented protocol (not
   obviously broken though, apart from authenticating only one side);
   hard-coded limit of 512-bit size for the DH modulus (2048 is the
   recommended minimum these days); direct use of the shared secret as a
   key (instead of applying a KDF); encryption with ECB, custom
   parameters and the client not having the ability to verify them.
 - The programs use the DH API in the same way that TLS does, so they
   have limited demonstration value.
 - The programs only show finite-field DH, which is not used all that
   much these days. What people want to see is mostly ECDH.


Patater:
Are there any CMakeLists or Makefile or visual studio files to update?

gilles-peskine-arm:
All of these files probably need updating:
```
$ git grep -l dh_client
ChangeLog
programs/.gitignore
programs/Makefile
programs/README.md
programs/pkey/CMakeLists.txt
programs/pkey/dh_client.c
visualc/VS2010/dh_client.vcxproj
visualc/VS2010/mbedTLS.sln
```

yanesca:
I remember talking about it, but I don't remember how we decided, do we want to keep updating ChangeLog in mbed-crypto?

gilles-peskine-arm:
We've talked about it but not decided anything. In the meantime, I propose to keep updating `ChangeLog` with an “Mbed Crypto” entry for anything that impacts the non-PSA API (but not for PSA API stuff because we have a huge backlog and no strong promise of backward compatibility yet).

Patater:
Please don't add any Mbed Crypto entries to the ChangeLog. Feel free to start a wiki page for the next version's release notes, though. Until we have a solution to ChangeLog maintenance and automatic merging, it's not worth the effort. We can bring it back later if it makes sense.

yanesca:
The PR is ready for review again.

ChangeLog
```diff
@@ -64,6 +64,9 @@ Features
    * Add a new function mbedtls_asn1_write_named_bitstring() to write ASN.1
      named bitstring in DER as required by RFC 5280 Appendix B.
 
+Mbed Crypto
+   * Removed two Diffie-Hellman examples which were relying on the Net module.
```
gilles-peskine-arm: Changelog entries are primarily for users of the library. Most users of the library don't care much about module dependencies, but they might care about the example programs. So we should tell them what to look at instead. Proposal:

> Remove the Diffie-Hellman examples which implemented a toy protocol inspired by TLS DH key exchange. For an example of how to use the DHM module, see the code that calls `mbedtls_dhm_xxx` in `ssl_tls.c` and `ssl_cli.c` in Mbed TLS.

ChangeLog
```diff
@@ -64,6 +64,9 @@ Features
    * Add a new function mbedtls_asn1_write_named_bitstring() to write ASN.1
      named bitstring in DER as required by RFC 5280 Appendix B.
 
+Mbed Crypto
```
Patater: Please add release note items to [our upcoming release's release notes page](https://confluence.arm.com/display/IoTBU/Mbed+Crypto+1.1.0+Release+Notes). We can edit them and get them into shape before publishing separately from code changes.

ChangeLog
yanesca: Fixing it. (I had removed it before the review, just forgot to push.)

programs/Makefile
```diff
@@ -49,8 +49,7 @@ endif
 
 APPS =	aes/aescrypt2$(EXEXT)		aes/crypt_and_hash$(EXEXT)	\
 	hash/hello$(EXEXT)		hash/generic_sum$(EXEXT)	\
-					pkey/dh_client$(EXEXT)		\
-	pkey/dh_genprime$(EXEXT)	pkey/dh_server$(EXEXT)		\
+	pkey/dh_genprime$(EXEXT)		\
```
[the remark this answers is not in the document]
yanesca: Fixed in ff43390.

ChangeLog
yanesca: I have added the text suggested by @gilles-peskine-arm to the page.

yanesca:
The PR is ready for review again.

programs/Makefile
```diff
@@ -48,33 +48,33 @@ LOCAL_LDFLAGS += -lz
...
-	pkey/ecdsa$(EXEXT)		pkey/gen_key$(EXEXT)		\
-	pkey/key_app$(EXEXT)		pkey/key_app_writer$(EXEXT)	\
-	pkey/mpi_demo$(EXEXT)		pkey/pk_decrypt$(EXEXT)		\
-	pkey/pk_encrypt$(EXEXT)		pkey/pk_sign$(EXEXT)		\
-	pkey/pk_verify$(EXEXT)		pkey/rsa_genkey$(EXEXT)		\
-	pkey/rsa_decrypt$(EXEXT)	pkey/rsa_encrypt$(EXEXT)	\
-	pkey/rsa_sign$(EXEXT)		pkey/rsa_verify$(EXEXT)		\
-	pkey/rsa_sign_pss$(EXEXT)	pkey/rsa_verify_pss$(EXEXT)	\
-	psa/crypto_examples$(EXEXT)					\
+	hash/hello$(EXEXT)		hash/generic_sum$(EXEXT)			\
```
Patater: Could you align to tab-width 8 please?

yanesca: Done.

PR #83, opened 2019-03-11
API: minor corrections after beta 2

gilles-peskine-arm:
* Minor corrections noticed after beta 2 was finalized.
* Specify `psa_generator_import_key` for most key types.

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -2941,10 +2941,10 @@ psa_status_t psa_generator_read(psa_crypto_generator_t *generator,
  *       in big-endian order. Discard it if it is not in the range
  *       [0, *N* - 2] where *N* is the boundary of the private key domain
  *       (the prime *p* for Diffie-Hellman, the subprime *q* for DSA,
- *       or the order of the curve's coordinate field for ECC).
+ *       or the order of the curve's base point for ECC).
  *       Add 1 to the resulting integer and use this as the private key *x*.
- *       This is the method described as
- *       "key-pair generation by testing candidates"
+ *       This method allows compliance to NIST standards, specifically
+ *       the methods titled "key-pair generation by testing candidates"
  *       in NIST SP 800-56A &sect;5.6.1.1.4 for Diffie-Hellman,
  *       in FIPS 186-4 &sect;B.1.2 for DSA, and
  *       in NIST SP 800-56A &sect;5.6.1.2.2 or
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -2509,8 +2509,6 @@ psa_status_t psa_aead_finish(psa_aead_operation_t *operation,
  * \retval #PSA_ERROR_BAD_STATE
  *         The operation state is not valid (not set up, nonce not set,
  *         encryption, or already completed).
- * \retval #PSA_ERROR_BUFFER_TOO_SMALL
- *         The size of the \p output buffer is too small.
  * \retval #PSA_ERROR_INVALID_ARGUMENT
  *         The total length of input to psa_aead_update_ad() so far is
  *         less than the additional data length that was previously
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -2896,10 +2896,8 @@ psa_status_t psa_generator_read(psa_crypto_generator_t *generator,
  *   The following key types defined in this specification follow this scheme:
  *
  *     - #PSA_KEY_TYPE_AES;
- *     - #PSA_KEY_TYPE_ARIA;
  *     - #PSA_KEY_TYPE_ARC4;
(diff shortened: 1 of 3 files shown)


include/psa/crypto.h
```diff
@@ -2904,7 +2904,7 @@ psa_status_t psa_generator_read(psa_crypto_generator_t *generator,
  *         If the key is persistent, the key material and the key's metadata
  *         have been saved to persistent storage.
  * \retval #PSA_ERROR_INSUFFICIENT_CAPACITY
- *                          There were fewer than \p output_length bytes
+ *                          There were fewer than \p bits * 8 bytes
```
yanesca: It looks like this should be `(bits +7) / 8` bytes.

gilles-peskine-arm: Pff, `bits * 8`, `bits / 8`…

It's actually `bits / 8` and not `(bits + 7) / 8` since `bits` is a multiple of 8 for symmetric keys.

But this function should also be specified for other key types. Implementations don't have to support every key type, but if they do then they should follow a standard. I've rewritten the specification of this function to define the behavior for every standard key type other than RSA.

gilles-peskine-arm: I pushed another commit to make the specification unambiguous for DH, DSA and ECC (Weierstrass). I chose a method that is codified by NIST, so that it's possible to be FIPS-compliant when implementing this function.

include/psa/crypto.h
```diff
@@ -2924,12 +2924,31 @@ psa_status_t psa_generator_read(psa_crypto_generator_t *generator,
...
+ *       ECC keys on a Weierstrass elliptic curve
+ *       (#PSA_KEY_TYPE_ECC_KEYPAIR(\c curve) where \c curve designates a
+ *       Weierstrass curve).
+ *       For these key types, interpret the byte string as integer
+ *       in big-endian order. Discard it if it is not in the range
+ *       [0, *N* - 2] where *N* is the boundary of the private key domain
+ *       (the prime *p* for Diffie-Hellman, the subprime *q* for DSA,
+ *       or the order of the curve's coordinate field for ECC).
+ *       Add 1 to the resulting integer and use this as the private key *x*.
+ *       This is the method described as
```
yanesca: The method described in the standard is more relaxed in some cases (for example when generating DH keys the standard allows going for shorter private keys than the size of the prime). I think writing something like "This method allows for compliance with NIST and FIPS standards, see:" would be more accurate.

include/psa/crypto.h
```diff
@@ -2924,12 +2924,31 @@ psa_status_t psa_generator_read(psa_crypto_generator_t *generator,
...
+ *     - Finite-field Diffie-Hellman keys (#PSA_KEY_TYPE_DH_KEYPAIR),
+ *       DSA keys (#PSA_KEY_TYPE_DSA_KEYPAIR), and
+ *       ECC keys on a Weierstrass elliptic curve
+ *       (#PSA_KEY_TYPE_ECC_KEYPAIR(\c curve) where \c curve designates a
+ *       Weierstrass curve).
+ *       For these key types, interpret the byte string as integer
+ *       in big-endian order. Discard it if it is not in the range
+ *       [0, *N* - 2] where *N* is the boundary of the private key domain
+ *       (the prime *p* for Diffie-Hellman, the subprime *q* for DSA,
+ *       or the order of the curve's coordinate field for ECC).
```
yanesca: In case of ECC the boundary is the order of the curve's base point (here I use the term curve in the sense of the set of cryptographic domain parameters and not as a mathematical object, but I assume that is what we do throughout this documentation).

include/psa/crypto.h
gilles-peskine-arm: I tweaked the wording.

include/psa/crypto.h
gilles-peskine-arm: Indeed. Fixed, thanks.

include/psa/crypto.h
```diff
@@ -2941,10 +2941,10 @@ psa_status_t psa_generator_read(psa_crypto_generator_t *generator,
  *       in big-endian order. Discard it if it is not in the range
  *       [0, *N* - 2] where *N* is the boundary of the private key domain
  *       (the prime *p* for Diffie-Hellman, the subprime *q* for DSA,
- *       or the order of the curve's coordinate field for ECC).
+ *       or the order of the curve's base point for ECC).
  *       Add 1 to the resulting integer and use this as the private key *x*.
- *       This is the method described as
- *       "key-pair generation by testing candidates"
+ *       This method allow compliance to NIST standards, specifically
```
yanesca: Typo: allow -> allows

gilles-peskine-arm: Amended

PR #90, opened 2019-04-10
Use Windows-specific renaming function

dgreen-arm:
On Windows, rename() fails if the new filename already exists.
Use the Windows specific function MoveFileExA with the
MOVEFILE_REPLACE_EXISTING flag set instead to do renames.

### DIFF
--- a/library/psa_its_file.c
+++ b/library/psa_its_file.c
@@ -33,6 +33,10 @@
 #define mbedtls_snprintf   snprintf
 #endif
 
+#if defined(_WIN32)
+#include <windows.h>
+#endif
+
 #include "psa_crypto_its.h"
 
 #include <limits.h>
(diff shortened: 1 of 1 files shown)


library/psa_its_file.c
```diff
@@ -209,7 +213,12 @@ psa_status_t psa_its_set( psa_storage_uid_t uid,
     }
     if( status == PSA_SUCCESS )
     {
+#if defined(_WIN32)
```
gilles-peskine-arm: Please avoid non-structural use of the preprocessor, with `#ifdef` around things that aren't a standalone C block. Here, I think the best solution would be to define a function-like macro `rename_replace_existing( oldpath, newpath )` with a platform-dependent definition.

library/psa_its_file.c
```diff
@@ -58,6 +62,13 @@
 #define PSA_ITS_MAGIC_STRING "PSA\0ITS\0"
 #define PSA_ITS_MAGIC_LENGTH 8
 
+#if defined(_WIN32)
+#define rename_replace_existing( oldpath, newpath ) \
+    (!MoveFileExA( oldpath, newpath, MOVEFILE_REPLACE_EXISTING ))
```
gilles-peskine-arm: Style:
```suggestion
    ( ! MoveFileExA( oldpath, newpath, MOVEFILE_REPLACE_EXISTING ) )
```

library/psa_its_file.c
```diff
@@ -58,6 +62,13 @@
 #define PSA_ITS_MAGIC_STRING "PSA\0ITS\0"
 #define PSA_ITS_MAGIC_LENGTH 8
 
+#if defined(_WIN32)
```
gilles-peskine-arm: Please add a comment that explains the purpose of this macro in a sentence or so, and documents the semantics of the return value (0=success, nonzero=error).

PR #97, opened 2019-04-26
Use generated source files from Mbed TLS

Patater:
There are a handful of generated source files that both Mbed Crypto and Mbed TLS can create. When created from Mbed Crypto, they contain only symbols, errors, options and other things visible within Mbed Crypto itself. When Mbed TLS generates these source files, it uses symbols, errors, options and other things visible from both Mbed TLS and Mbed Crypto. When building libmbedcrypto.a or libmbedcrypto.so as a submodule from Mbed TLS, use the Mbed-TLS-generated files as these are the most comprehensive and useful to the rest of Mbed TLS.


library/Makefile
```diff
@@ -88,8 +88,22 @@ OBJS_CRYPTO=	aes.o		aesni.o		arc4.o		\
...
 		ripemd160.o	rsa_internal.o	rsa.o  		\
 		sha1.o		sha256.o	sha512.o	\
-		threading.o	timing.o	version.o	\
-		version_features.o		xtea.o
+		threading.o	timing.o			\
+		xtea.o
+
+# For files generated by the parent project (Mbed TLS) when building Mbed
+# Crypto as a submodule, ensure that the parent project instance is used.
+ifdef USE_CRYPTO_SUBMODULE
```
AndrzejKurek: I'd recommend changing that to an `ifneq ($(USE_CRYPTO_SUBMODULE), 0)`, same as here:
https://github.com/ARMmbed/mbedtls/pull/2403/commits/fc5a7f53154057f25273eafaf58b7b3b2cd112ee
since USE_CRYPTO_SUBMODULE=0 will also have it defined.

Patater:
Rebased to use more robust option checking.

PR #101, opened 2019-05-03
Check unused attributes in import and copy

gilles-peskine-arm:
If the type/size attributes are specified during key creation but not required because they can be inferred from the actual data, check them against the actual data.

This PR adds tests for the type and size, but not for domain parameters which are currently mostly not implemented.

Internal ref: https://github.com/ARMmbed/psa-crypto/issues/140#issuecomment-489107823 — Validating unused attributes in key creation


library/psa_crypto.c
```diff
@@ -1480,20 +1480,79 @@ static void psa_fail_key_creation( psa_key_slot_t *slot )
...
+    }
+
+    if( attributes->domain_parameters_size != 0 )
+    {
+#if defined(MBEDTLS_RSA_C)
+        if( PSA_KEY_TYPE_IS_RSA( slot->type ) )
+        {
+            mbedtls_mpi actual, required;
+            int ret;
+            mbedtls_mpi_init( &actual );
```
Patater: In a real implementation, we might want to comment that this is comparing the public exponent (the only RSA domain parameter).

tests/suites/test_suite_psa_crypto.function
```diff
@@ -1959,14 +1957,8 @@ void copy_key( int source_usage_arg, int source_alg_arg,
...
-    TEST_EQUAL( psa_copy_key( source_handle,
-                              &target_attributes, &target_handle ),
-                expected_status_arg );
-    if( expected_status_arg != PSA_SUCCESS )
-    {
-        TEST_EQUAL( target_handle, 0 );
-        goto exit;
-    }
+    PSA_ASSERT( psa_copy_key( source_handle,
+                              &target_attributes, &target_handle ) );
```
Patater: For real implementation, would recommend asserting target_handle is non-zero upon return

PR #102, opened 2019-05-06
Multipart AEAD buffer output sizes

gilles-peskine-arm:
* Allow delayed output in AEAD for all algorithms. Although CCM and GCM allow the data to be encrypted or decrypted as a stream cipher, some hardware implementations can only process a full block at a time apart from the last block. Fix ARMmbed/psa-crypto#130
* Add output buffer parameter for `psa_aead_verify`. It needs it as much as `psa_aead_finish`.
* Add buffer size macros for multipart AEAD functions.
* Add a flag for built-from-block-cipher in AEAD algorithm encoding.
* Declare and implement ChaCha20 and ChaCha20-Poly1305 algorithms.

Out of scope: testing the new algorithms beyond smoke tests; testing the new macros; testing multipart AEAD.


include/psa/crypto_sizes.h
```diff
@@ -268,49 +268,102 @@
...
+ * \return                    The AEAD ciphertext size for the specified
+ *                            algorithm.
+ *                            If the AEAD algorithm is not recognized, return 0.
+ *                            An implementation may return either 0 or a
+ *                            correct size for an AEAD algorithm that it
+ *                            recognizes, but does not support.
+ */
+#define PSA_AEAD_DECRYPT_OUTPUT_SIZE(alg, ciphertext_length)      \
+    (PSA_AEAD_TAG_LENGTH(alg) != 0 ?                              \
+     (plaintext_length) - PSA_AEAD_TAG_LENGTH(alg) :              \
```
athoelke: copy paste bug? - `plaintext_length` -> `ciphertext_length`

include/psa/crypto_sizes.h
```diff
@@ -268,49 +268,102 @@
...
+ *                            correct size for an AEAD algorithm that it
+ *                            recognizes, but does not support.
+ */
+/* For all the AEAD modes defined in this specification, it is possible
+ * to emit output without delay. However, hardware may not always be
+ * capable of this. So for modes based on a block cipher, allow the
+ * implementation to delay the output until it has a full block. */
+#define PSA_AEAD_UPDATE_OUTPUT_SIZE(alg, input_length)                 \
+    (PSA_ALG_IS_AEAD_ON_BLOCK_CIPHER(alg) ?                             \
+     ((plaintext_length) + PSA_MAX_BLOCK_CIPHER_BLOCK_SIZE - 1) / PSA_MAX_BLOCK_CIPHER_BLOCK_SIZE : \
```
athoelke: 1. parameter mismatch: `plaintext_length` -> `input_length`
2. Result for AEAD on block cipher is not rounded, but scaled. Presume missing `( ... ) * PSA_MAX_BLOCK_CIPHER_BLOCK_SIZE`

include/psa/crypto.h
```diff
@@ -2495,6 +2495,14 @@ psa_status_t psa_aead_update_ad(psa_aead_operation_t *operation,
  *          - In particular, do not copy the output anywhere but to a
  *            memory or storage space that you have exclusive access to.
  *
+ * This function does not require the input to be aligned to any
+ * particular block boundary. If the implementation can only process
+ * a whole block at a time, it must store the last partial input block
+ * or adjust its internal state accordingly until the next call to
+ * psa_aead_update(), psa_aead_finish() or psa_aead_verify(), and produce
+ * the corresponding output when sufficient input is available or on the
+ * finish or verify call.
```
Patater: Is this asking too much of implementations (for accelerators or secure elements)?

gilles-peskine-arm:
> Do we need a new error code returned from any of the AEAD functions in order to indicate "The hardware is busy/delayed right now

No. The API is fully synchronous. If the hardware is busy, the call blocks.

All calls consume all the input (or fail). What's implementation-defined is only how whether the output at each stage is the maximum that can be mathematically calculated or not. Is this something that needs to be clarified in the documentation of specific functions?

include/psa/crypto.h
gilles-peskine-arm: The point of this PR is precisely to avoid asking too much. But the wording could use some improvements. What did you think this is asking?

gilles-peskine-arm: I've reworded this paragraph.

include/psa/crypto_sizes.h
gilles-peskine-arm: Fixed, and I modified the AEAD tests to use this macro.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -3272,11 +3272,16 @@ void aead_encrypt_decrypt( int key_type_arg, data_t *key_data,
...
     size_t output_length2 = 0;
-    size_t tag_length = 16;
+    size_t tag_length = PSA_AEAD_TAG_LENGTH( alg );
     psa_status_t expected_result = expected_result_arg;
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
 
     output_size = input_data->len + tag_length;
+    /* For all currently defined algorithms, PSA_AEAD_ENCRYPT_OUTPUT_SIZE
+     * should be exact. */
+    if( expected_result != PSA_ERROR_INVALID_ARGUMENT )
```
Patater: Do we expect the length to be correct for errors other than PSA_ERROR_INVALID_ARGUMENT? If not, should we check for expected_result == PSA_SUCCESS?

include/psa/crypto.h
```diff
@@ -2497,11 +2497,11 @@ psa_status_t psa_aead_update_ad(psa_aead_operation_t *operation,
...
- * a whole block at a time, it must store the last partial input block
- * or adjust its internal state accordingly until the next call to
- * psa_aead_update(), psa_aead_finish() or psa_aead_verify(), and produce
- * the corresponding output when sufficient input is available or on the
- * finish or verify call.
+ * a whole block at a time, it must consume all the input provided, but
+ * it may delay the end of the corresponding output until a subsequent
+ * call to psa_aead_update(), psa_aead_finish() or psa_aead_verify()
+ * provides sufficient input. The amount of data that can be delayed
+ * in this way is bounded by #PSA_AEAD_UPDATE_OUTPUT_SIZE.
```
Patater: OK, this makes it more clear that an implementation doesn't have to store and delay if it doesn't want to, but it has the option to. Good.

tests/suites/test_suite_psa_crypto.function
gilles-peskine-arm: The length is expected to be correct except if the algorithm is invalid. This includes cases where the operation might fail for other reasons, at least `INVALID_SIGNATURE`. It would be better to have properly separate tests for the macro and for the operation, but that's a much bigger task than this PR (it applies to all the other size macros).

Patater:

PR #104, opened 2019-05-06
Make key ids global and define their range

gilles-peskine-arm:
Make key ids global: you no longer need to specify a key's lifetime when you open it, the key id is enough. This is the “global model” of key ids, as opposed to the former “drive letter” model. Internal ref: https://github.com/ARMmbed/psa-crypto/issues/138

Carve the range of possible key ids into 3 sub-ranges: one for the application, one for the implementation and one reserved for future use. Internal ref: https://github.com/ARMmbed/psa-crypto/issues/136

### DIFF
--- a/include/psa/crypto_types.h
+++ b/include/psa/crypto_types.h
@@ -85,6 +85,19 @@ typedef uint32_t psa_algorithm_t;
  */
 
 /** Encoding of key lifetimes.
+ *
+ * The lifetime of a key indicates where it is stored and what system actions
+ * may create and destroy it.
+ *
+ * Keys with the lifetime #PSA_KEY_LIFETIME_VOLATILE are automatically
+ * destroyed when the application terminates or on a power reset.
+ *
+ * Keys with a lifetime other than #PSA_KEY_LIFETIME_VOLATILE are said
+ * to be _persistent_.
+ * Persistent keys are preserved if the application or the system restarts.
+ * Persistent keys have a key identifier of type #psa_key_id_t.
+ * The application can call psa_open_key() to open a persistent key that
+ * it created previously.
  */
 typedef uint32_t psa_key_lifetime_t;
 
--- a/include/psa/crypto_types.h
+++ b/include/psa/crypto_types.h
@@ -102,6 +102,13 @@ typedef uint32_t psa_algorithm_t;
 typedef uint32_t psa_key_lifetime_t;
 
 /** Encoding of identifiers of persistent keys.
+ *
+ * - Applications may freely choose key identifiers in the range
+ *   #PSA_KEY_ID_USER_MIN to #PSA_KEY_ID_USER_MAX.
+ * - Implementations may define additional key identifiers in the range
+ *   #PSA_KEY_ID_VENDOR_MIN to #PSA_KEY_ID_VENDOR_MAX.
+ * - Key identifiers outside these ranges are reserved for future use
+ *   in future versions of this specification.
  */
 /* Implementation-specific quirk: The Mbed Crypto library can be built as
  * part of a multi-client service that exposes the PSA Crypto API in each
--- a/include/psa/crypto_types.h
+++ b/include/psa/crypto_types.h
@@ -107,8 +107,8 @@ typedef uint32_t psa_key_lifetime_t;
  *   #PSA_KEY_ID_USER_MIN to #PSA_KEY_ID_USER_MAX.
  * - Implementations may define additional key identifiers in the range
  *   #PSA_KEY_ID_VENDOR_MIN to #PSA_KEY_ID_VENDOR_MAX.
- * - Key identifiers outside these ranges are reserved for future use
- *   in future versions of this specification.
+ * - 0 is reserved as an invalid key identifier.
+ * - Key identifiers outside these ranges are reserved for future use.
  */
 /* Implementation-specific quirk: The Mbed Crypto library can be built as
  * part of a multi-client service that exposes the PSA Crypto API in each
--- a/include/psa/crypto_values.h
+++ b/include/psa/crypto_values.h
@@ -1440,6 +1440,19 @@
  */
 #define PSA_KEY_LIFETIME_PERSISTENT             ((psa_key_lifetime_t)0x00000001)
 
+/** The minimum value for a key identifier chosen by the application.
+ */
+#define PSA_KEY_ID_USER_MIN                     ((psa_key_id_t)0x00000000)
+/** The minimum value for a key identifier chosen by the application.
+ */
+#define PSA_KEY_ID_USER_MAX                     ((psa_key_id_t)0x3fffffff)
+/** The minimum value for a key identifier chosen by the application.
+ */
+#define PSA_KEY_ID_VENDOR_MIN                   ((psa_key_id_t)0x40000000)
+/** The minimum value for a key identifier chosen by the application.
+ */
+#define PSA_KEY_ID_VENDOR_MAX                   ((psa_key_id_t)0x7fffffff)
+
 /**@}*/
 
 /** \defgroup policy Key policies
--- a/include/psa/crypto_values.h
+++ b/include/psa/crypto_values.h
@@ -1442,7 +1442,7 @@
 
 /** The minimum value for a key identifier chosen by the application.
  */
-#define PSA_KEY_ID_USER_MIN                     ((psa_key_id_t)0x00000000)
+#define PSA_KEY_ID_USER_MIN                     ((psa_key_id_t)0x00000001)
 /** The minimum value for a key identifier chosen by the application.
  */
 #define PSA_KEY_ID_USER_MAX                     ((psa_key_id_t)0x3fffffff)
--- a/include/psa/crypto_values.h
+++ b/include/psa/crypto_values.h
@@ -1443,13 +1443,13 @@
 /** The minimum value for a key identifier chosen by the application.
  */
 #define PSA_KEY_ID_USER_MIN                     ((psa_key_id_t)0x00000001)
-/** The minimum value for a key identifier chosen by the application.
+/** The maximum value for a key identifier chosen by the application.
  */
 #define PSA_KEY_ID_USER_MAX                     ((psa_key_id_t)0x3fffffff)
-/** The minimum value for a key identifier chosen by the application.
+/** The minimum value for a key identifier chosen by the implementation.
  */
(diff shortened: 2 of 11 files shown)


include/psa/crypto_types.h
```diff
@@ -85,10 +85,30 @@ typedef uint32_t psa_algorithm_t;
...
  */
 typedef uint32_t psa_key_lifetime_t;
 
 /** Encoding of identifiers of persistent keys.
+ *
+ * - Applications may freely choose key identifiers in the range
+ *   #PSA_KEY_ID_USER_MIN to #PSA_KEY_ID_USER_MAX.
+ * - Implementations may define additional key identifiers in the range
+ *   #PSA_KEY_ID_VENDOR_MIN to #PSA_KEY_ID_VENDOR_MAX.
+ * - Key identifiers outside these ranges are reserved for future use
```
athoelke: Suggest to remove the first "future" in the sentence.

include/psa/crypto_types.h
```diff
@@ -85,10 +85,30 @@ typedef uint32_t psa_algorithm_t;
...
+ *   in future versions of this specification.
```
athoelke: I cannot recall - did we decide that key id of zero was valid or always invalid? - or if we should have a key id that is always invalid/null.

Or is this now an application decision whether or not to allocate a key with id of zero? - is this worth mentioning at all?

include/psa/crypto_values.h
```diff
@@ -1440,6 +1440,19 @@
  */
 #define PSA_KEY_LIFETIME_PERSISTENT             ((psa_key_lifetime_t)0x00000001)
 
+/** The minimum value for a key identifier chosen by the application.
+ */
+#define PSA_KEY_ID_USER_MIN                     ((psa_key_id_t)0x00000000)
```
athoelke: Needs a `u` suffix. Same for the following three constants

library/psa_crypto_slot_management.c
```diff
@@ -278,11 +278,10 @@ static psa_status_t persistent_key_setup( psa_key_lifetime_t lifetime,
...
 }
 
-psa_status_t psa_open_key( psa_key_lifetime_t lifetime,
-                           psa_key_file_id_t id,
-                           psa_key_handle_t *handle )
+psa_status_t psa_open_key( psa_key_file_id_t id, psa_key_handle_t *handle )
 {
-    return( persistent_key_setup( lifetime, id, handle, PSA_SUCCESS ) );
+    return( persistent_key_setup( PSA_KEY_LIFETIME_PERSISTENT,
+                                  id, handle, PSA_SUCCESS ) );
```
Patater: Would nobody ever open a key that doesn't have a `#PSA_KEY_LIFETIME_PERSISTENT` lifetime? Is open not used for volatile keys? Is this obvious enough from the function documentation?

Would anybody ever try this, (perhaps from separate threads or processes)?
1. psa_create_key(PSA_KEY_LIFETIME_VOLATILE, someid, &handle1);
1. psa_open_key(someid, &handle2);

Two issues:
- They might not realize that the create does an implicit open (although this is documented already)
- Key ownership and sharing implications potentially come into scope, in a service model. e.g. if we allow opening of a volatile key, do we need to block opening a volatile key that is already open? do we share the key somehow?

include/psa/crypto_values.h
```diff
@@ -1440,6 +1440,19 @@
...
+/** The minimum value for a key identifier chosen by the application.
+ */
+#define PSA_KEY_ID_USER_MAX                     ((psa_key_id_t)0x3fffffff)
```
Patater: Are these reserved ranges implementation defined or part of PSA spec?

include/psa/crypto_types.h
gilles-peskine-arm: I'm in two minds about id=0. On the one hand, allowing it violates the general principle that 0 is a “safe” invalid value. On the other hand, forbidding it means that you can't recognize a valid key id by a simple bit mask, and you can calculate a key id hash by just taking an _n_-bit hash.

I'd rather keep `PSA_KEY_ID_VENDOR_MIN ≤ 1` to not gratuitously break existing code. Is 1..0x3fffffff a sensible range? Should it be 1..0x40000000? Or am I just imagining things and 1..0x3fffffff (or 1..0x7fffffff) is perfectly fine?

library/psa_crypto_slot_management.c
gilles-peskine-arm: Volatile keys don't have a key id. I think that's now clearly explained in the doc of `psa_key_id_t`. Maybe it needs to be explained elsewhere as well?

Since volatile keys don't have a key id, you can't open them.

include/psa/crypto_values.h
gilles-peskine-arm: Part of PSA spec. An application needs to know what key ids it can use, and so does an implementation.

library/psa_crypto_slot_management.c
gilles-peskine-arm: What's the connection with sharing? Why would sharing require a persistent id?

What's the current API for key sharing ? We discussed proposals in https://github.com/ARMmbed/psa-crypto/issues/72 , but that was done by the services team, and I can't find any design document now.

include/psa/crypto_types.h
athoelke: How many places does an implementation need to verify that a valid key id has been provided? - just during key creation and key opening it seems, so the need to check both ends of the range rather than just a one sided range isn't going to set off the code bloat alarm.

If we implement ARMmbed/psa-crypto#158 to unify handles and key ids then "zero is special" returns anyway to represent the NULL handle.

library/psa_crypto_slot_management.c
Patater: How can you know you are opening the same volatile key without having a persistent ID? Do we prevent opening multiple handles to the same key?

gilles-peskine-arm: You can't open a volatile key. Opening requires a persistent key id.

Patater: Yes, understood. Because you can't open a volatile key, I'm highlighting potential difficulty in obtaining multiple handles to a volatile key. It's easy to get multiple handles to a persistent key; I don't believe we prevent obtaining multiple handles to a persistent key.

> Why would sharing require a persistent id?
Because you can't open a volatile key.

gilles-peskine-arm: There's no way to obtain multiple handles to the same volatile key. Sharing creates a separate handle which (at least conceptually) points to a copy of the key material.

include/psa/crypto_types.h
gilles-peskine-arm: “Reserved for future use” is an idiom, so I'd rather keep it. Should I just stop the sentence there?

include/psa/crypto_values.h
gilles-peskine-arm: Does this actually matter, given that the constants are being cast? If it does we need to change all the literals in the specification, not just these.

tests/suites/test_suite_psa_crypto_slot_management.function
```diff
@@ -23,31 +23,47 @@ typedef enum
...
+static size_t num_key_ids_used;
+
+/* Record a key id as potentially used in a test case. */
+static int test_uses_key_id( psa_key_id_t key_id )
+{
+    size_t i;
+    if( key_id > PSA_MAX_PERSISTENT_KEY_IDENTIFIER )
+    {
+        /* Don't touch key id values that designate non-key files. */
+        return( 1 );
```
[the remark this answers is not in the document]
gilles-peskine-arm: It's an error to use such a key id, but not a _test error_: there's at least one test case that does it precisely to check that the implementation returns the expected error code. The cleanup code calls `psa_destroy_persistent_key` which is an internal function and may not enforce the limitation, so the code that sets up the information used by the cleanup code arranges not to call `psa_destroy_persistent_key` out of range.

include/psa/crypto_types.h
gilles-peskine-arm: I decided that being consistent with “0 is invalid”, and protecting against the application forgetting to set a value that it really should have set (which is part of the reason for “0 is invalid”), are more important than making the permitted range exactly describable with a power of 2. So I've changed `PSA_KEY_ID_VENDOR_MIN` to 0.

gilles-peskine-arm:
@athoelke @Patater The PR is now ready for re-review. I haven't addressed all your comments, but for some of them I'm waiting for clarifications since I'm not sure whether you're requesting a change.

library/psa_crypto_slot_management.c
Patater: Thanks for answering my questions. Sharing is out of scope for this PR, so resolving this conversation.

include/psa/crypto_values.h
```diff
@@ -1440,6 +1440,19 @@
  */
 #define PSA_KEY_LIFETIME_PERSISTENT             ((psa_key_lifetime_t)0x00000001)
 
+/** The minimum value for a key identifier chosen by the application.
+ */
+#define PSA_KEY_ID_USER_MIN                     ((psa_key_id_t)0x00000001)
+/** The minimum value for a key identifier chosen by the application.
```
AndrzejKurek: Maximum*

include/psa/crypto_values.h
```diff
@@ -1440,6 +1440,19 @@
...
+ */
+#define PSA_KEY_ID_USER_MAX                     ((psa_key_id_t)0x3fffffff)
+/** The minimum value for a key identifier chosen by the application.
+ */
+#define PSA_KEY_ID_VENDOR_MIN                   ((psa_key_id_t)0x40000000)
+/** The minimum value for a key identifier chosen by the application.
```
AndrzejKurek: Maximal*

include/psa/crypto_values.h
Patater: Maximum would be good. Minimal/Maximal; Minimum/Maximum.

gilles-peskine-arm: Thanks. Fixed, also application → implementation.

Patater:

include/psa/crypto_types.h
Patater: Repetition of "future" has been eliminated.

include/psa/crypto_values.h
Patater: Can be changed later (in a different PR), if deemed necessary.

PR #108, opened 2019-05-14
Add policy usage flag to copy a key

gilles-peskine-arm:
Require `PSA_KEY_USAGE_COPY` in `psa_copy_key`.


include/psa/crypto.h
```diff
@@ -850,6 +850,15 @@ psa_status_t psa_export_public_key(psa_key_handle_t handle,
  * this function may be used to share a key with a different party,
  * subject to implementation-defined restrictions on key sharing.
  *
+ * The policy on the source key must have the usage flag
+ * #PSA_KEY_USAGE_COPY set.
+ * In addition, some lifetimes also require the source key to have the
+ * usage flag #PSA_KEY_USAGE_EXPORT, because otherwise the source key
```
athoelke: The wording isn't completely clear, for example:
* who is doing the "locking inside"? - presume this is the SE designer, but better to avoid unnamed actors in the documentation.
* Is the use case here one where a key is being copied out of the SE into a volatile key (for example), and the SE design defaults to denying this operation unless the key is extractable? - presumably a copy within the SE would only require COPY policy in this set up?

include/psa/crypto_values.h
```diff
@@ -1459,6 +1459,20 @@
  */
 #define PSA_KEY_USAGE_EXPORT                    ((psa_key_usage_t)0x00000001)
 
+/** Whether the key may be copied.
+ *
+ * This flag allows the use of psa_crypto_copy() to make a copy of the key
```
athoelke: psa_copy_key()

include/psa/crypto_values.h
```diff
@@ -1459,6 +1459,20 @@
...
+ * with the same policy or a more restrictive policy.
+ *
+ * For some lifetimes, copying a key also requires the usage flag
+ * #PSA_KEY_USAGE_EXPORT, because otherwise the source key
+ * is locked inside a secure processing environment and cannot be
```
athoelke: Clarify use case? - as above

include/psa/crypto.h
gilles-peskine-arm: I updated the documentation.

Patater:

PR #109, opened 2019-05-15
Individual setters for persistent key attributes

gilles-peskine-arm:
Use individual setters for the id and lifetime fields of an attribute structure, like the other attributes.

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -208,14 +208,15 @@ typedef struct psa_key_attributes_s psa_key_attributes_t;
 
 /** Declare a key as persistent and set its key identifier.
  *
- * If the attribute structure declares the key as volatile (which is
- * the default content of an attribute structure), this function sets
+ * If the attribute structure currently declares the key as volatile (which
+ * is the default content of an attribute structure), this function sets
  * the lifetime attribute to #PSA_KEY_LIFETIME_PERSISTENT.
  *
- * This function does not access storage, it merely fills the attribute
- * structure with given value. The persistent key will be written to
- * storage when the attribute structure is passed to a key creation
- * function such as psa_import_key(), psa_generate_random_key(),
+ * This function does not access storage, it merely stores the given
+ * value in the structure.
+ * The persistent key will be written to storage when the attribute
+ * structure is passed to a key creation function such as
+ * psa_import_key(), psa_generate_random_key(),
  * psa_generate_derived_key() or psa_copy_key().
  *
  * This function may be declared as `static` (i.e. without external
@@ -231,12 +232,16 @@ static void psa_set_key_id(psa_key_attributes_t *attributes,
 /** Set the location of a persistent key.
  *
  * To make a key persistent, you must give it a persistent key identifier
- * with psa_set_key_id().
+ * with psa_set_key_id(). By default, a key that has a persistent identifier
+ * is stored in the default storage area identifier by
+ * #PSA_KEY_LIFETIME_PERSISTENT. Call this function to choose a storage
+ * area, or to explicitly declare the key as volatile.
  *
- * This function does not access storage, it merely fills the attribute
- * structure with given value. The persistent key will be written to
- * storage when the attribute structure is passed to a key creation
- * function such as psa_import_key(), psa_generate_random_key(),
+ * This function does not access storage, it merely stores the given
+ * value in the structure.
+ * The persistent key will be written to storage when the attribute
+ * structure is passed to a key creation function such as
+ * psa_import_key(), psa_generate_random_key(),
  * psa_generate_derived_key() or psa_copy_key().
  *
  * This function may be declared as `static` (i.e. without external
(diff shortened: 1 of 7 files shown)


include/psa/crypto_values.h
```diff
@@ -663,10 +663,8 @@
  * Then you may create and use a key as follows:
  * - Set the key usage field using #PSA_ALG_ANY_HASH, for example:
  *   ```
- *   psa_key_policy_set_usage(&policy,
- *                            PSA_KEY_USAGE_SIGN, //or PSA_KEY_USAGE_VERIFY
- *                            PSA_xxx_SIGNATURE(PSA_ALG_ANY_HASH));
- *   psa_set_key_policy(handle, &policy);
+ *   psa_set_key_usage_flags(&attributes, PSA_KEY_USAGE_SIGN); // or VERIFY
+ *   psa_set_key_algorithm(PSA_xxx_SIGNATURE(PSA_ALG_ANY_HASH));
```
Patater: Missing `&attributes` here?

```diff
- psa_set_key_algorithm(PSA_xxx_SIGNATURE(PSA_ALG_ANY_HASH));
+ psa_set_key_algorithm(&attributes, PSA_xxx_SIGNATURE(PSA_ALG_ANY_HASH));
```

include/psa/crypto_struct.h
```diff
@@ -287,12 +287,28 @@ static inline void psa_make_key_persistent(psa_key_attributes_t *attributes,
...
 {
     return( attributes->id );
 }
 
+static inline void psa_set_key_lifetime(psa_key_attributes_t *attributes,
+                                        psa_key_lifetime_t lifetime)
+{
+    attributes->lifetime = lifetime;
+    if( lifetime == PSA_KEY_LIFETIME_VOLATILE )
+        attributes->id = 0;
```
Patater: So, 0 is now officially an invalid id? What's the best place to document this?

include/psa/crypto.h
```diff
@@ -203,30 +206,51 @@ psa_status_t psa_crypto_init(void);
...
- *
  * This function may be declared as `static` (i.e. without external
  * linkage). This function may be provided as a function-like macro,
  * but in this case it must evaluate each of its arguments exactly once.
  *
  * \param[out] attributes       The attribute structure to write to.
  * \param id                    The persistent identifier for the key.
+ */
+static void psa_set_key_id(psa_key_attributes_t *attributes,
+                           psa_key_id_t id);
```
Patater: What should happen if the caller specifies `id` 0 here? Is this invalid? I think we need some more documentation here.

include/psa/crypto_struct.h
gilles-peskine-arm: Rework of https://github.com/ARMmbed/mbed-crypto/pull/104 coming up.

include/psa/crypto.h
gilles-peskine-arm: The functions have special cases to preserve the invariant that `id = 0` if `lifetime = VOLATILE`. But `id = 0` with `lifetime ≠ VOLATILE` is ok as the intermediate state if you call `set_key_lifetime` first, then `set_key_id`. Do you think there should be a special case for `id=0`?

include/psa/crypto_values.h
gilles-peskine-arm: I amended the corresponding commit.

include/psa/crypto.h
Patater: `id = 0` should be allowed as an intermediate state, and when the attributes are actually used, an error can be returned if the id is invalid.

include/psa/crypto.h
```diff
@@ -203,30 +206,51 @@ psa_status_t psa_crypto_init(void);
...
-/** Declare a key as persistent.
+/** Declare a key as persistent and set its key identifier.
+ *
+ * If the attribute structure declares the key as volatile (which is
+ * the default content of an attribute structure), this function sets
+ * the lifetime attribute to #PSA_KEY_LIFETIME_PERSISTENT.
  *
  * This function does not access storage, it merely fills the attribute
- * structure with given values. The persistent key will be written to
+ * structure with given value. The persistent key will be written to
```
Patater:
```diff
- structure with given value.
+ structure with the given value.
```

or, better

```diff
- structure with given value.
+ structure with the given \p id.
```

include/psa/crypto.h
```diff
@@ -203,30 +206,51 @@ psa_status_t psa_crypto_init(void);
...
+static void psa_set_key_id(psa_key_attributes_t *attributes,
+                           psa_key_id_t id);
+
+/** Set the location of a persistent key.
+ *
+ * To make a key persistent, you must give it a persistent key identifier
+ * with psa_set_key_id().
+ *
+ * This function does not access storage, it merely fills the attribute
+ * structure with given value. The persistent key will be written to
```
Patater:
```diff
- structure with given value.
+ structure with the given \p lifetime.
```

include/psa/crypto.h
gilles-peskine-arm: I don't think it's useful to specify what's going into the structure in this sentence. But I fixed the grammar and clarified a few things.

Patater:

PR #110, opened 2019-05-15
Print a command trace if the check-names.sh exits unexpectedly

gilles-peskine-arm:
Side-port of https://github.com/ARMmbed/mbedtls/pull/2647

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -536,7 +536,7 @@ component_check_files () {
 
 component_check_names () {
     msg "test/build: declared and exported names" # < 3s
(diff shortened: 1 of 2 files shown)


Patater:
We don't actually need to side-port this. We generally get changes from Mbed TLS shared code into Mbed Crypto by merging the Mbed TLS development branch to the Mbed Crypto development branch, like how we are doing with https://github.com/ARMmbed/mbedtls/pull/2636 and did with other PRs in the past.

gilles-peskine-arm:
Closing, since won't be side-porting this (and we've resolved the issue anyway).

PR #111, opened 2019-05-15
Pass handle parameter last on key creation

gilles-peskine-arm:
Change the order of parameters for `psa_import_key` and `psa_generate_derived_key` to pass the handle output parameter last. This makes the order consistent with the general convention in the library to put inputs before outputs. Other key handle creation functions (`psa_open_key`, `psa_generate_random_key`, `psa_copy_key`) already had the handle output parameter last.


gilles-peskine-arm:
`getting_started.md` is out of scope of the API document release. We'll update it in a batch later.

Patater:
CI failures on API branch are to do with building with crypto as a submodule and also TLS using PSA, both of which we don't care about at this point in development.

PR #112, opened 2019-05-16
Remove domain parameters from API 1.0

gilles-peskine-arm:
Domain parameters aren't quite working yet. So remove them from the API, but keep them as an experimental extension in our reference implementation. At this point, we plan to fix the domain parameter API and add them to the specification in a point release, possibly in a modified form.

This has the following impact for applications:

* There is no longer a way to generate an RSA key with e≠65537.
* Diffie-Hellman keys must use one of the groups from RFC 7919.
* DSA is no longer referenced and cannot be implemented without some API extension.

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -656,7 +656,8 @@ psa_status_t psa_destroy_key(psa_key_handle_t handle);
  *   and `PSA_ECC_CURVE_BRAINPOOL_PXXX`).
  *   This is the content of the `privateKey` field of the `ECPrivateKey`
  *   format defined by RFC 5915.
- * - For Diffie-Hellman key exchange key pairs (#PSA_KEY_TYPE_DH_KEYPAIR), the
+ * - For Diffie-Hellman key exchange key pairs (key types for which
+ *   #PSA_KEY_TYPE_IS_DH_KEYPAIR is true), the
  *   format is the representation of the private key `x` as a big-endian byte
  *   string. The length of the byte string is the private key size in bytes
  *   (leading zeroes are not stripped).
@@ -729,7 +730,8 @@ psa_status_t psa_export_key(psa_key_handle_t handle,
  *   representation of the public key `y = g^x mod p` as a big-endian byte
  *   string. The length of the byte string is the length of the base prime `p`
  *   in bytes.
- * - For Diffie-Hellman key exchange public keys (#PSA_KEY_TYPE_DH_PUBLIC_KEY),
+ * - For Diffie-Hellman key exchange public keys (key types for which
+ *   #PSA_KEY_TYPE_IS_DH_PUBLIC_KEY is true),
  *   the format is the representation of the public key `y = g^x mod p` as a
  *   big-endian byte string. The length of the byte string is the length of the
  *   base prime `p` in bytes.
@@ -3253,7 +3255,8 @@ psa_status_t psa_key_derivation_output_bytes(
  *       discard the first 8 bytes, use the next 8 bytes as the first key,
  *       and continue reading output from the operation to derive the other
  *       two keys).
- *     - Finite-field Diffie-Hellman keys (#PSA_KEY_TYPE_DH_KEYPAIR),
+ *     - Finite-field Diffie-Hellman keys (#PSA_KEY_TYPE_DH_KEYPAIR(\c group)
+ *       where \c group designates any Diffie-Hellman group),
  *       DSA keys (#PSA_KEY_TYPE_DSA_KEYPAIR), and
  *       ECC keys on a Weierstrass elliptic curve
  *       (#PSA_KEY_TYPE_ECC_KEYPAIR(\c curve) where \c curve designates a
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -641,10 +641,6 @@ psa_status_t psa_destroy_key(psa_key_handle_t handle);
  *       coefficient         INTEGER,  -- (inverse of q) mod p
  *   }
  *   ```
- * - For DSA private keys (#PSA_KEY_TYPE_DSA_KEYPAIR), the format is the
- *   representation of the private key `x` as a big-endian byte string. The
- *   length of the byte string is the private key size in bytes (leading zeroes
- *   are not stripped).
  * - For elliptic curve key pairs (key types for which
  *   #PSA_KEY_TYPE_IS_ECC_KEYPAIR is true), the format is
  *   a representation of the private value as a `ceiling(m/8)`-byte string
@@ -726,10 +722,6 @@ psa_status_t psa_export_key(psa_key_handle_t handle,
  *      - The byte 0x04;
  *      - `x_P` as a `ceiling(m/8)`-byte string, big-endian;
  *      - `y_P` as a `ceiling(m/8)`-byte string, big-endian.
- * - For DSA public keys (#PSA_KEY_TYPE_DSA_PUBLIC_KEY), the format is the
- *   representation of the public key `y = g^x mod p` as a big-endian byte
- *   string. The length of the byte string is the length of the base prime `p`
- *   in bytes.
  * - For Diffie-Hellman key exchange public keys (key types for which
  *   #PSA_KEY_TYPE_IS_DH_PUBLIC_KEY is true),
  *   the format is the representation of the public key `y = g^x mod p` as a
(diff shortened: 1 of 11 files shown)


include/psa/crypto.h
```diff
@@ -186,11 +184,11 @@ psa_status_t psa_crypto_init(void);
  *    psa_generate_derived_key() or psa_copy_key(). This function reads
  *    the attribute structure, creates a key with these attributes, and
  *    outputs a handle to the newly created key.
- * -# The attribute structure is now no longer necessary. If you called
- *    psa_set_key_domain_parameters() earlier, you must call
- *    psa_reset_key_attributes() to free any resources used by the
- *    domain parameters. Otherwise calling psa_reset_key_attributes()
- *    is optional.
+ * -# The attribute structure is now no longer necessary.
+ *    You may call psa_reset_key_attributes(), although this is optional
```
yanesca: Shouldn't we remove `psa_reset_key_attributes()` from the API too?

gilles-peskine-arm: No. It stays. You can get away without it now, but it's extremely likely that a future API version will require it. You are supposed to call it after `psa_get_key_attributes`.

tests/suites/test_suite_psa_crypto_metadata.function
```diff
@@ -457,3 +457,22 @@ void ecc_key_types( int curve_arg, int curve_bits_arg )
     TEST_ASSERT( curve_bits <= PSA_VENDOR_ECC_MAX_CURVE_BITS );
 }
 /* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_DHM_C */
+void dh_key_types( int group_arg, int group_bits_arg )
+{
+    psa_dh_group_t group = group_arg;
+    size_t group_bits = group_bits_arg;
+    psa_key_type_t public_type = PSA_KEY_TYPE_ECC_PUBLIC_KEY( group );
```
yanesca: Copypasta: this should be `PSA_KEY_TYPE_DH_PUBLIC_KEY`. Same goes for the macros in the lines below.

gilles-peskine-arm: I made a fixup commit to fix this and add `KEY_TYPE_IS_DH`.

include/psa/crypto.h
yanesca: Do you say that it is extremely likely, because of the reintroduction of `key_attributes` or do you have some other use case in mind as well?

gilles-peskine-arm: Domain parameters are the definite thing I have in mind. But also I'd be surprised if nothing else came up, especially when we consider more server and other rich platform use cases.

include/psa/crypto_extra.h
```diff
@@ -443,6 +443,248 @@ psa_status_t psa_generate_random_key_to_handle(psa_key_handle_t handle,
...
+ *
+ * \param key_type  A supported key type.
+ * \param key_bits  The size of the key in bits.
+ *
+ * \return If the parameters are valid and supported, return
+ *         a buffer size in bytes that guarantees that
+ *         psa_get_key_domain_parameters() will not fail with
+ *         #PSA_ERROR_BUFFER_TOO_SMALL.
+ *         If the parameters are a valid combination that is not supported
+ *         by the implementation, this macro either shall return either a
```
k-stachowiak: "Either" is duplicated in this sentence.

gilles-peskine-arm: Fixed here and in other places that had the same copypasted sentence.

gilles-peskine-arm:
Rebased on top of the target branch. The previous version is at https://github.com/gilles-peskine-arm/mbed-crypto/tree/psa-remove_domain_parameters-1

include/psa/crypto.h
```diff
@@ -189,11 +187,11 @@ psa_status_t psa_crypto_init(void);
...
+ *    with the workflow presented here because the attributes currently
+ *    defined in this specification do not require any additional resources
+ *    beyond the structure itself.
```
Patater: I'm not sure we should make this optional. Certain implementations might require `psa_reset_key_attributes()` to be called to free implementation-specific resources like domain parameters. It'd be nice if applications would continue to work when domain parameters become part of the official API.

include/psa/crypto_values.h
```diff
@@ -1187,7 +1134,7 @@
  */
 #define PSA_ALG_IS_HASH_AND_SIGN(alg)                                   \
     (PSA_ALG_IS_RSA_PSS(alg) || PSA_ALG_IS_RSA_PKCS1V15_SIGN(alg) ||    \
-     PSA_ALG_IS_DSA(alg) || PSA_ALG_IS_ECDSA(alg))
+     PSA_ALG_IS_ECDSA(alg))
```
Patater: Do we need to provide an implementation specific `PSA_ALG_IS_HASH_AND_SIGN()` that understands `PSA_ALG_IS_DSA()`?

gilles-peskine-arm: Yes, if we implement DSA, which we don't.

The current tooling doesn't allow us to document a macro without giving a definition for it. When we move away from Doxygen we'll be able to do it.

include/psa/crypto.h
gilles-peskine-arm: We debated this internally and concluded that we might as well allow it.

include/psa/crypto.h
```diff
@@ -105,8 +105,7 @@ psa_status_t psa_crypto_init(void);
  *   and its lifetime.
  * - The key's policy, comprising usage flags and a specification of
  *   the permitted algorithm(s).
- * - Information about the key itself: the key type, the key size, and
- *   for some key type additional domain parameters.
+ * - Information about the key itself: the key type and its size.
```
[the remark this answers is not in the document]
gilles-peskine-arm: Sure, everything is usable with implementation-specific extensions.

include/psa/crypto.h
Patater: I’m proposing stronger than allowing it, but to maintain requiring it.

include/psa/crypto_values.h
gilles-peskine-arm: But I actually did preserve the definition with DSA support in `crypto_extra.h`, anyway.

include/psa/crypto.h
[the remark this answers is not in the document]
Patater: The commit message is all lies!

include/psa/crypto.h
gilles-peskine-arm: See internal discussion https://github.com/ARMmbed/psa-crypto/issues/140#issuecomment-486767289 and https://github.com/ARMmbed/mbedtls-psa/pull/263#discussion_r279832614

Patater: I've read those threads.

My recommendation is to encourage future-proof use of the API by always requiring a reset.

```diff
-  You may call psa_reset_key_attributes(), although this is optional...
+ You must call psa_reset_key_attributes() to free any resources used by the attributes structure.
```

I can see your chosen approach is to only require calling reset when it might be actually necessary; and without key domain parameters, Mbed Crypto doesn't allocate resources when handling attributes. I think this assumes too much about the implementation, as some implementations could require the user calling reset for the attributes currently specified, although Mbed Crypto wouldn't. Applications written for an implementation that doesn't require calling reset will not be portable to implementations that require reset.

If we dont't want to always require it, I'd suggest removing reset from the API for now, so that no implementations that require it appear and cause application portability issues.

yanesca: I don't have strong feelings about this, but for me the removal seems the best solution. Mainly because I am secretly hoping that we can get away without adding domain parameters back:
- I think we have a reasonable compromise for RSA, it has less significance and it's popularity is decreasing.
- I don't think anybody will mourn DSA
- For FFDH named parameters are the safest (if you are worried about logjam, don't use small modulus) and we can use an ECC-like mechanism as we do it now.
- For PQ schemes to come: fiddling with the domain parameters has usually non-trivial impact on security strength and anything else than named parameters seems impractical to me. (The exception is hash based PQ, but there the parameters are as simple as with symmetric crypto.)

athoelke: I'm also feeling that this halfway house is not ideal. There's still the issue with how to extract a variable sized key attributes in an efficient way when we reintroduce them. I think the real options are:

1. Extended attributes are not part of the psa_key_attributes_t.
   * When we re-introduce domain parameters and other similar 'large' attributes, we create a different type to hold them.
   * If you need to extract them, you call an extra API to retrieve them and then release them later.
   * When creating keys, we either allow the extended attributes to be referenced by the key_attributes (but not owned), or add extra parameters to the creation APIs. (I prefer the former)
   * Retain `reset()` to zero the attributes if apps want to do this, is is not required to be called. Releasing memory is an operation on the extended attributes.

2. Make `reset()` mandatory for applications to use in this version of the spec. When domain params are added to the attributes, `reset()` starts checking if there is memory to be released. This adds cost to all apps as many won't touch the zero'd attributes after reset and many compilers will not elide the pointless memset().

3. Remove `reset()` from the API right now. Reintroduce it as "conditional mandatory" - **if you use domain parameters then you must call `reset()`**

I think I actually prefer approach 1. It moves the tricky memory management into a separate structure and allows `psa_key_attributes_t` to always be a simple plain-old-data struct.

Patater: The risk of some implementations requiring free to come up for the currentlt defined parameters is fairly low. Reset just zeros in this version, and is optional.

Patater:

include/psa/crypto.h
gilles-peskine-arm: With my gatekeeper hat for the API, I say: this stays as is in beta3. It's out of scope of this PR anyway. Let's discuss it further for the next beta: we can still make reset officially mandatory or completely optional. This commits us to keeping the function, but it's useful anyway to zero the structure.

Issue #37, opened 2019-01-31
Implement separate-inputs key derivation and key agreement


gilles-peskine-arm:
Implement the changes to the key derivation and key agreement API made in PSA Crypto API 1.0 beta1:

* Key derivation takes one input at a time, with new functions `psa_key_derivation_setup`, `psa_key_derivation_input_bytes`, `psa_key_derivation_input_key`, `psa_set_generator_capacity`.
* Key agreement uses separate functions to chain a key derivation (changed prototype of `psa_key_agreement`) or not (new function `psa_key_agreement_raw_shared_secret`).

Draft (internal link): https://github.com/ARMmbed/mbedtls-psa/pull/244

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTCRYPT-658*

gilles-peskine-arm:
This has been merged in the [psa-api-1.0-beta](https://github.com/ARMmbed/mbed-crypto/tree/psa-api-1.0-beta) branch for a while (https://github.com/ARMmbed/mbedtls-psa/commit/9dcc80e628f96c6ed64e3ace1df85866b0385030) and has been released in Mbed Crypto 2.0.0.

PR #113, opened 2019-05-16
Replace "generator" with "key derivation"

gilles-peskine-arm:
Generators are mostly about key derivation (currently: _only_ about key derivation). "Generator" is not a commonly used term in cryptography. So favor "derivation" as terminology. Call a generator a key derivation operation structure, since it behaves like other multipart operation structures. Furthermore, the function names were not fully consistent; this PR makes them more consistent.


gilles-peskine-arm:
`programs/psa/key_ladder_demo.c` also uses an obsolete function, which is a bigger problem. It'll have to be reworked as part of https://github.com/ARMmbed/mbed-crypto/issues/37 .

Patater:

PR #117, opened 2019-05-20
PSA return status coverage script

gilles-peskine-arm:
Add infrastructure to run unit tests and collect the return values for every PSA API function that returns psa_status_t.

    ./tests/scripts/psa_collect_statuses.py >statuses.txt

Status: runs ok but not very well documented and the build script is not very robust.

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -629,6 +629,16 @@ component_test_everest () {
     make test
 }
 
+component_test_psa_collect_statuses () {
+  msg "build+test: psa_collect_statuses" # ~30s
+  scripts/config.pl full
+  scripts/config.pl unset MBEDTLS_MEMORY_BUFFER_ALLOC_C # slow and irrelevant
+  record_status tests/scripts/psa_collect_statuses.py
+  # Check that psa_crypto_init() succeeded at least once
+  record_status grep -q '^0:psa_crypto_init:' tests/statuses.log
+  rm -f tests/statuses.log
+}
+
 component_test_full_cmake_clang () {
     msg "build: cmake, full config, clang" # ~ 50s
     scripts/config.pl full
--- a/tests/scripts/psa_collect_statuses.py
+++ b/tests/scripts/psa_collect_statuses.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+"""Describe the test coverage of PSA functions in terms of return statuses.
+
+1. Build Mbed Crypto with -DRECORD_PSA_STATUS_COVERAGE_LOG
+2. Run psa_collect_statuses.py
+
+The output is a series of line of the form "psa_foo PSA_ERROR_XXX". Each
+function/status combination appears only once.
+
+This script must be run from the top of an Mbed Crypto source tree.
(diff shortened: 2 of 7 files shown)


Patater:
This may conflict with https://github.com/ARMmbed/mbed-crypto/pull/213

I like the idea of checking that return codes are documented. I think the API is stable enough now that having that capability will be a help and not a hindrance.

Is it possible to continue work on this?

gilles-peskine-arm:
I'll rebase this on top of #213 if we decide to adopt it.

Are you ok with the current design? It was good enough for my goal at the time, which was to compare reachable errors with documented errors and is still relevant. But it isn't very well integrated into Mbed Crypto. The script I have for the other side, to collect documented errors, works from the PSA spec (private link: https://github.com/ARMmbed/psa-crypto/pull/227).

Patater:
> Are you ok with the current design?

It's a bit funky in that its a mix of shell in a Makefile and some Python, but it's beneficial as is. We can refactor into something better as needed.

gilles-peskine-arm:
I rebased the changes on top of current `development`. I added some documentation, but I did not change the design. I think this is good enough to include, and we'll see later whether to improve the design or to leave it as a minor hack depending on how useful we think this is.

tests/scripts/all.sh
```diff
@@ -629,6 +629,16 @@ component_test_everest () {
     make test
 }
 
+component_test_psa_collect_statuses () {
+  msg "build+test: psa_collect_statuses" # ~30s
+  scripts/config.pl full
+  scripts/config.pl unset MBEDTLS_MEMORY_BUFFER_ALLOC_C # slow and irrelevant
```
Patater: We may not need to add `unset MBEDTLS_MEMORY_BUFFER_ALLOC_C` if we sideport the removal of MBEDTLS_MEMORY_BUFFER_ALLOC_C from full config. Necessary for now.

tests/scripts/psa_collect_statuses.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    rebuilt = False
+    if not options.use_existing_log and os.path.exists(options.log_file):
+        os.remove(options.log_file)
+    if not os.path.exists(options.log_file):
+        if options.clean_before:
+            subprocess.check_call(['make', 'clean'],
+                                  cwd='tests',
+                                  stdout=sys.stderr)
+        with open(os.devnull, 'w') as devnull:
```
k-stachowiak: Why not store the build log in case of a failure?

tests/scripts/all.sh
gilles-peskine-arm: We definitely should sideport that change. I'd expect it to come around the next time we sync from mbedtls.

tests/scripts/psa_collect_statuses.py
gilles-peskine-arm: Good idea. I wrote this quickly and primarily for interactive use where I didn't want to scroll through build logs, but the log should definitely be stored in a file. https://github.com/ARMmbed/mbed-crypto/issues/252

PR #121, opened 2019-05-21
Fix unlinked macros in API spec

gilles-peskine-arm:
Get rid of ugly `#PSA_xxx` rendering.


gilles-peskine-arm:
Doc-only change. `doxygen.sh` passes locally, and I even checked that the default build works. So we can merge without waiting for CI.

PR #122, opened 2019-05-21
Update API document to beta3

gilles-peskine-arm:
1. Update the rendered PDF and HTML (for which this branch is the host, we link to it at https://github.com/ARMmbed/mbed-crypto) to 1.0 beta3.
1. Update the `psa-crypto-api` branch to the latest 1.0 beta sources.
1. Update `README.md` to link to `index.html` instead of `modules.html` which no longer exists.


gilles-peskine-arm:
I'd originally made the PR from the wrong starting point, which is why the history was a complete mess. I've pushed a whole new version with the intended history. I tweaked the commits a little: now there's the merge first, then rc4 docs, then rc5 docs. I also fixed the commit messages to indicate the proper rc numbers.

Patater:

PR #123, opened 2019-05-21
Update link to PSA Crypto API HTML

gilles-peskine-arm:
Now that we aren't using Doxygen directly any longer (#122), there is no longer a `modules.html`. Link to `index.html` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ There are currently a few deviations where the library does not yet implement th
 
 ### PSA Cryptography API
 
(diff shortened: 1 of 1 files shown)


Patater:
CI failure is ABI job, known to fail until without https://github.com/ARMmbed/mbedtls/pull/2636 present in the development branch. The merge tests here were run before https://github.com/ARMmbed/mbedtls/pull/2636 landed in the development branch. Tests pass on the actual merge result.

PR #125, opened 2019-05-23
Fix PSA Crypto API links in README

gilles-peskine-arm:
The PDF link was to a preview page. Link to the actual PDF instead.

The HTML link went to a third-party service that broke links. Link to Github Pages instead. **This requires the HTML to be in `/docs` on the `master` branch**. Alternatively we could change the repository settings to use a dedicated `gh-pages` branch — @Patater should we do this?

~There is still a problem with the HTML: [Sphinx puts resources (CSS, JS, images) in a directory with the hard-coded name `_static`](https://github.com/sphinx-doc/sphinx/issues/2202) but [Github Pages ignores files whose name starts with an underscore](https://help.github.com/en/articles/files-that-start-with-an-underscore-are-missing). We need to find a solution for this, but in the meantime, [unformatted HTML](https://armmbed.github.io/mbed-crypto/html/index.html) isn't as bad as missing content.~ https://github.com/ARMmbed/mbed-crypto/pull/127 makes `_static` work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ There are currently a few deviations where the library does not yet implement th
 
 ### PSA Cryptography API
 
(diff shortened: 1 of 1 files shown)


Patater:
I don't think we should use a gh-pages branch. I'd like to just merge once into our development branch, into its docs folder, and have github publish from there. This may mean we need to rename our development branch to master. That might not be a bad thing, since most other git-based projects use master as the name for the branch we use as our development branch.

Patater:
I've manually tested the links. They work!

PR #127, opened 2019-05-24
Allow publishing of folders starting with `_`

Patater:
GitHub pages uses Jekyll by default. Jekyll ignores all files and
folders beginning with `_`. Sphinx is hard-coded to always output its
theme to a folder named `_static`. In order to include `_static` in the
published site, add a `.nojekyll` file to disable Jekyll processing (we
don't need it, as we have html files and nothing needing Jekyll
processing like markdown).


Patater:
Tested manually that this enables the `_static` folder to work.

PR #128, opened 2019-05-24
Implementation-specific extension: Keys may allow a second algorithm

gilles-peskine-arm:
Add a second permitted algorithm to key policies, as an implementation-specific feature. This is an experimental feature which may or may not be added to the PSA Crypto API after 1.0.

This is a replay of https://github.com/ARMmbed/mbedtls-psa/pull/266 [private link] on the API branch. Because the development branch and the API branch have diverged (copy key policy, renamed identifiers, changed key creation interface), merging this feature from development to the API branch led to a lot of difficult-to-understand conflicts. Here, I take the commits from https://github.com/ARMmbed/mbedtls-psa/pull/266 one by one and modify them to suit the modern API.

Compared with https://github.com/ARMmbed/mbedtls-psa/pull/266:

* The order of commits is slightly different, to have preparation commits before feature commits.
* Commits related to `PSA_ECC_CURVE_BITS` have already been merged through https://github.com/ARMmbed/mbed-crypto/pull/107.
* I omitted commits to `docs/architecture/mbed-crypto-storage-specification.md` because this file has not been modified in the API branch.

Once this PR is merged, the next step is to merge `development` into `psa-api-1.0-beta`. This merge will have conflicts but they will be easy to resolve (the merge commit is at https://github.com/gilles-peskine-arm/mbed-crypto/tree/psa-api-1.0-beta-merge_development_20190524, I'll make a PR for it once this one is merged).

PR #130, opened 2019-05-27
Merge development into psa-api-1.0-beta

gilles-peskine-arm:
Merge `development` into `psa-api-1.0-beta`. The conflicts fall into two buckets:

* Most conflicts are due to the parallel development of the enrollment algorithm policy feature in the two branches, which was done this way because the feature in `development` could not be merged easily into `psa-api-1.0-beta`. I resolved these conflicts by taking the development done for the API branch in https://github.com/ARMmbed/mbed-crypto/pull/128.
* Some SSL files changed in the API branch and were removed in `development`. I resolved these conflicts by removing the files. The corresponding updates of PSA API usage will need to be made in `mbedtls`.


tests/scripts/test_psa_constant_names.py
```diff
@@ -23,6 +23,8 @@ def __init__(self, filename, line_number):
         self.line_number = line_number
 
 class read_file_lines:
+    # Dear Pylint, conventionally, a context manager class name is lowercase.
```
k-stachowiak: Did you consider using https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager here?

library/CMakeLists.txt
```diff
@@ -65,34 +64,26 @@ set(src_crypto
...
-    x509.c
-    x509_create.c
-    x509_crl.c
-    x509_crt.c
-    x509_csr.c
-    x509write_crt.c
-    x509write_csr.c
+# For files generated by the parent project (Mbed TLS) when building Mbed
+# Crypto as a submodule, ensure that the parent project instance is used.
+if(USE_CRYPTO_SUBMODULE)
```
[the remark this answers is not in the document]
gilles-peskine-arm: We've merged “always use crypto submodule” into Mbed TLS, but not yet dropped compatibility for previous Mbed TLS revisions in Mbed Crypto. In any case, that's out of scope here: the goal of this PR is solely to merge `mbed-crypto/development` into `psa-api-1.0-beta`, not to change anything that isn't affected by this merge.

tests/scripts/test_psa_constant_names.py
gilles-peskine-arm: Would it help? This context manager doesn't the acquire/release pattern that is the typical use case of `contextlib.contextmanager`.

scripts/generate_psa_constants.py
```diff
@@ -179,6 +186,12 @@
 '''
 
 class MacroCollector:
+    """Collect PSA crypto macro definitions from C header files.
```
AndrzejKurek: Strange indentation

AndrzejKurek:
@gilles-peskine-arm - would you be able to list the exact conflicts you've encountered, or is this data lost now? If you don't have the time for it, I can re-create it by running the merge myself in that case.

scripts/generate_psa_constants.py
gilles-peskine-arm: Is it? What's strange about it?

gilles-peskine-arm:
@AndrzejKurek I didn't keep a list of conflicts. I only checked that apart from the deletion of `ssl_*.c`, they were all related to the enrollment algorithm policy feature, and given #128 the resolution for that is always to take the API branch. The “merge follow-up” commit is actually a semantic conflict where enrollment algorithm policy code from the development branch was taken in when it shouldn't have been, because there was no syntactic conflict.

scripts/generate_psa_constants.py
AndrzejKurek: The ending and beginning quotes are misaligned, and the comments from lines below start at the beginning of the line.

gilles-peskine-arm: That's one of the valid indentation styles for docstrings. But I notice that other parts of the script use a different style, so I'll update this one to be consistent.

tests/scripts/test_psa_constant_names.py
k-stachowiak: I presumed that could have silence the code style error, but I have just check, and it doesn't, so no, unfortunately it would not help. Please disregard my above comment.

library/CMakeLists.txt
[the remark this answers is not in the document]
Patater: > so I would expect this option to no longer exist,

It'll continue to need to exist as long as Mbed Crypto can build both independently on its own, and as a submodule. Even if Mbed TLS always uses Mbed Crypto as a submodule, Mbed Crypto itself needs to support being built as a submodule or standalone.

Patater:
API compatibility test is failing. This is because we are changing the API on purpose and is OK.
- key attributes changed
- alg2 added
- new enums

Patater:
TLS tests with the API branch are expected to fail until TLS is updated to use the new API.

Patater:
I've performed the merge myself and ran into minimal differences: I had chose to keep using local variables to hold correctly typed parameters. I see your merge decided to inconsistently remove the typed local variables. Was this deliberate?

- In `test_suite_psa_crypto_slot_management.function`, why sometimes use a local variable for typing the function parameter and sometimes not? e.g. `target_alg` is used as a correctly typed variable, but `target_alg2_arg` is used directly, not through a correctly typed variable; `expected_usage_arg` turns into `expected_usage`, but `target_usage_arg` does not in `copy_success`.
- In ``test_suite_psa_crypto.function`, why remove use of local variables for typing function parameter names? e.g. `target_alg` used to be present, but now `target_alg_arg` is used directly.

Patater:
Just one question about typed argument locals, otherwise LGTM

gilles-peskine-arm:
These were not conscious decisions: for this PR I systematically took the API branch where there was a conflict. Maybe I ended up with slightly inferior code when I ported the feature in #128. I'll look at the specific cases.

gilles-peskine-arm:
As far as I can tell, `alg2` is always treated consistently with `alg`.

`_arg` parameters add complexity, so if I could, I wouldn't use them. I use them when it's a choice between that and casting the parameter upon use. Parameters need an `_arg` and a correctly-typed variable when they're used in comparisons, or in other contexts where a promotion might not do the right thing or might trigger a compiler warning. Some parameters get the `_arg` treatment just because they're close to another one that needs it. Some undoubtedly get it because they're the result of copy-paste of code that needed it. We don't have a policy on their usage and I'd rather not waste time on defining a policy: that time would be better used improving the test framework to support correctly-typed parameters.

So 1. I don't see any reason why this isn't a correct _merge_ (of two states that might each be flawed on their own); 2. I don't see any actual flaw in the result.

gilles-peskine-arm:
CI is ok except for TLS (expected) and ABI checking (the script reported a pass but failed due to a known issue for which the fix isn't merged yet). @Patater Since this has two approvals and your reservations are minor, I'm merging this PR to unblock the follow-ups.

PR #131, opened 2019-05-27
Remove old key creation functions and update cipher and pk

gilles-peskine-arm:
Remove the PSA key creation functions that used half-filled handles, which have been replaced by the attribute-based functions.

Now the crypto CI is expected to pass. The TLS CI is still expected to fail since TLS code hasn't been updated to the new key creation API and to the newish key derivation API.

Status: ready. ~~this PR follows https://github.com/ARMmbed/mbed-crypto/pull/130. The first new commit is “Remove spurious obsolete function call”.~~


library/psa_crypto_slot_management.h
```diff
@@ -55,6 +55,16 @@ psa_status_t psa_initialize_key_slots( void );
...
 
+/** Allocate a key slot.
+ *
+ * \param[out] handle   On success, a handle to a newly allocated key slot.
+ *                      0 if an error occurs.
+ *
+ * \retval #PSA_SUCCESS
+ * \retval #PSA_ERROR_INSUFFICIENT_MEMORY
+ */
+psa_status_t psa_allocate_key( psa_key_handle_t *handle );
```
AndrzejKurek: When exactly is this used now?

library/psa_crypto_slot_management.c
```diff
@@ -99,56 +99,25 @@ void psa_wipe_all_key_slots( void )
...
-    psa_status_t status;
-
-    status = psa_get_key_slot( handle, &slot );
-    if( status != PSA_SUCCESS )
-        return( status );
-
-    return( psa_wipe_key_slot( slot ) );
-}
-
-psa_status_t psa_allocate_key( psa_key_handle_t *handle )
```
AndrzejKurek: Ok, I can see now that it's removed later.

library/psa_crypto_slot_management.h
```diff
@@ -55,15 +55,18 @@ psa_status_t psa_initialize_key_slots( void );
  * This does not affect persistent storage. */
 void psa_wipe_all_key_slots( void );
 
-/** Allocate a key slot.
+/** Find a free key slot and mark it as in use.
  *
- * \param[out] handle   On success, a handle to a newly allocated key slot.
- *                      0 if an error occurs.
+ * \param[out] handle   On success, a slot number that is not in use. This
```
AndrzejKurek: This comments is a bit confusing. Isn't the slot in use after it's marked as in use?

gilles-peskine-arm:
I've rebased this PR on top of the target branch now that the preceding PR #130 is merged. The base is identical except for two Python scripts which this PR doesn't touch.

library/psa_crypto_slot_management.c
gilles-peskine-arm: I guess it could have been removed earlier, but I think it makes more sense to remove the function as part of the refactoring of the auxiliary function that it calls. Not a big deal anyway.

library/psa_crypto_slot_management.h
gilles-peskine-arm: I clarified the description.

gilles-peskine-arm:
CI passed except for TLS (which was expected) and the ABI checker (which was also expected since this PR removes obsolete functions, altough [the manner of the failure was not expected](https://github.com/ARMmbed/mbedtls/issues/2672)).

PR #132, opened 2019-05-28
Test that closing a handle reclaims its resources

gilles-peskine-arm:
We have infrastructure in place to test that `mbedtls_psa_crypto_free` frees all resources (run the tests with a memory leak detector such as ASan). But in the real world this function is called rarely if ever. What matters is that functions like `psa_close_key` and `psa_destroy_key` free the resources used by the key.

This PR adds a check that the key slot array is empty before closing `mbedtls_psa_crypto_free` at the end of each test. A few tests call `mbedtls_psa_crypto_free` with open handles and verify (when run under ASan) that that this does free the resources occupied by those keys.

As a byproduct, this PR adds a function that Mbed Crypto users can call to know how many key slots are available, in use, etc.

Status: ready ~~successor to https://github.com/ARMmbed/mbed-crypto/pull/131 . The first new commit is 1287739ce9a9748846deb6d01cd35ce7a439f9d5 “Create PSA-specific helper function file”.~~

### DIFF
--- a/include/psa/crypto_extra.h
+++ b/include/psa/crypto_extra.h
@@ -116,6 +116,43 @@ static inline psa_algorithm_t psa_get_key_enrollment_algorithm(
  */
 void mbedtls_psa_crypto_free( void );
 
+/** \brief Statistics about
+ * resource consumption related to the PSA keystore.
+ *
+ * \note The content of this structure is not part of the stable API and ABI
+ *       of Mbed Crypto and may change arbitrarily from version to version.
+ */
+typedef struct mbedtls_psa_stats_s
+{
+    /** Number of slots containing key material for a volatile key. */
+    size_t volatile_slots;
+    /** Number of slots containing key material for a key which is in
+     * internal persistent storage. */
+    size_t persistent_slots;
+    /** Number of slots containing a reference to a key in a
+     * secure element. */
+    size_t external_slots;
+    /** Number of slots which are occupied, but do not contain
+     * key material yet. */
+    size_t half_filled_slots;
+    /** Number of slots that contain cache data. */
+    size_t cache_slots;
+    /** Number of slots that are not used for anything. */
+    size_t empty_slots;
+    /** Largest key id value among open keys in internal persistent storage. */
+    psa_key_id_t max_open_internal_key_id;
+    /** Largest key id value among open keys in secure elements. */
+    psa_key_id_t max_open_external_key_id;
+} mbedtls_psa_stats_t;
+
+/** \brief Get statistics about
+ * resource consumption related to the PSA keystore.
+ *
+ * \note When Mbed Crypto is built as part of a service, with isolation
+ *       between the application and the keystore, the service may or
+ *       may not expose this function.
+ */
+void mbedtls_psa_get_stats( mbedtls_psa_stats_t *stats );
 
 /**
  * \brief Inject an initial entropy seed for the random generator into
--- a/library/psa_crypto_slot_management.c
+++ b/library/psa_crypto_slot_management.c
@@ -232,4 +232,36 @@ psa_status_t psa_close_key( psa_key_handle_t handle )
     return( psa_wipe_key_slot( slot ) );
 }
 
+void mbedtls_psa_get_stats( mbedtls_psa_stats_t *stats )
+{
+    psa_key_handle_t key;
+    memset( stats, 0, sizeof( *stats ) );
+    for( key = 1; key <= PSA_KEY_SLOT_COUNT; key++ )
+    {
+        psa_key_slot_t *slot = &global_data.key_slots[key - 1];
+        if( slot->type == PSA_KEY_TYPE_NONE )
+        {
+            if( slot->allocated )
+                ++stats->half_filled_slots;
+            else
+                ++stats->empty_slots;
+            continue;
+        }
+        if( slot->lifetime == PSA_KEY_LIFETIME_VOLATILE )
+            ++stats->volatile_slots;
+        else if( slot->lifetime == PSA_KEY_LIFETIME_PERSISTENT )
+        {
+            ++stats->persistent_slots;
+            if( slot->persistent_storage_id > stats->max_open_internal_key_id )
+                stats->max_open_internal_key_id = slot->persistent_storage_id;
+        }
+        else
+        {
+            ++stats->external_slots;
+            if( slot->persistent_storage_id > stats->max_open_external_key_id )
+                stats->max_open_external_key_id = slot->persistent_storage_id;
+        }
+    }
+}
+
 #endif /* MBEDTLS_PSA_CRYPTO_C */
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -104,6 +104,7 @@ $(BINARIES): %$(EXEXT): %.c $(DEP)
 	echo "  CC    $<"
 	$(CC) $(LOCAL_CFLAGS) $(CFLAGS) $<	$(LOCAL_LDFLAGS) $(LDFLAGS) -o $@
 
+$(filter test_suite_psa_crypto%, $(BINARIES)): psa_helpers.function
 
 clean:
 ifndef WINDOWS
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -104,8 +104,11 @@ $(BINARIES): %$(EXEXT): %.c $(DEP)
 	echo "  CC    $<"
 	$(CC) $(LOCAL_CFLAGS) $(CFLAGS) $<	$(LOCAL_LDFLAGS) $(LDFLAGS) -o $@
 
+# Some test suites require additional header files.
 $(filter test_suite_psa_crypto%, $(BINARIES)): psa_crypto_helpers.h
+$(addprefix embedded_,$(filter test_suite_psa_crypto%, $(APPS))): embedded_%: TESTS/mbedtls/%/psa_crypto_helpers.h
 $(filter test_suite_psa_%, $(BINARIES)): psa_helpers.h
+$(addprefix embedded_,$(filter test_suite_psa_%, $(APPS))): embedded_%: TESTS/mbedtls/%/psa_helpers.h
 
 clean:
 ifndef WINDOWS
@@ -143,3 +146,17 @@ $(EMBEDDED_TESTS): embedded_%: suites/$$(firstword $$(subst ., ,$$*)).function s
 
 generate-target-tests: $(EMBEDDED_TESTS)
 
+define copy_header_to_target
+TESTS/mbedtls/$(1)/$(2): $(2)
+	echo "  Copy ./$$@"
+ifndef WINDOWS
+	mkdir -p $$(@D)
+	cp $$< $$@
+else
+	mkdir $$(@D)
+	copy $$< $$@
+endif
+
+endef
+$(foreach app, $(APPS), $(foreach file, $(wildcard *.h), \
+	$(eval $(call copy_header_to_target,$(app),$(file)))))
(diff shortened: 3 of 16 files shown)


tests/Makefile
```diff
@@ -112,6 +104,7 @@ $(BINARIES): %$(EXEXT): %.c $(DEP)
 	echo "  CC    $<"
 	$(CC) $(LOCAL_CFLAGS) $(CFLAGS) $<	$(LOCAL_LDFLAGS) $(LDFLAGS) -o $@
 
+$(filter test_suite_psa_crypto%, $(BINARIES)): psa_helpers.function
```
gilles-peskine-arm: Do I need to do anything else so that the tests can find `psa_helpers.function` when building for on-target testing (`embedded_xxx` targets)?

tests/psa_helpers.function
```diff
@@ -0,0 +1,39 @@
...
+/** Evaluate an expression and fail the test case if it returns an error.
+ *
+ * \param expr      The expression to evaluate. This is typically a call
+ *                  to a \c psa_xxx function that returns a value of type
+ *                  #psa_status_t.
+ */
+#define PSA_ASSERT( expr ) TEST_EQUAL( ( expr ), PSA_SUCCESS )
+
+/*
+ * Local Variables:
```
AndrzejKurek: What is this comment for?

tests/suites/test_suite_psa_crypto.function
```diff
@@ -1,15 +1,11 @@
...
-#if defined(MBEDTLS_PSA_CRYPTO_SPM)
-#include "spm/psa_defs.h"
-#endif
-
 #include "mbedtls/asn1.h"
 #include "mbedtls/asn1write.h"
 #include "mbedtls/oid.h"
 
-#include "psa/crypto.h"
+#include "psa_helpers.function"
```
AndrzejKurek: I'm not sure if including such files is clean. If it's meant to be included, I'd recommend changing the extension to .h instead. If it's to be used as an input for the Python scripts, .function is fine.

include/psa/crypto_extra.h
```diff
@@ -116,6 +116,43 @@ static inline psa_algorithm_t psa_get_key_enrollment_algorithm(
  */
 void mbedtls_psa_crypto_free( void );
 
+/** \brief Statistics about
+ * resource consumption related to the PSA keystore.
```
AndrzejKurek: Why is this broken in two lines in such a strange manner?

tests/psa_helpers.function
```diff
@@ -32,6 +32,36 @@
...
 
+static void test_helper_psa_done( int line, const char *file )
+{
+    mbedtls_psa_stats_t stats;
+    const char *msg = NULL;
+
+    mbedtls_psa_get_stats( &stats );
+
+    if( stats.volatile_slots != 0 )
+        msg = "A volatile slot has not been closed properly.";
```
AndrzejKurek: Such checks cover each other in case of multiple failures at once.

tests/psa_helpers.function
gilles-peskine-arm: [Emacs](https://www.gnu.org/software/emacs/manual/html_node/emacs/Specifying-File-Variables.html) (and if someone wants to add the Vi(m) equivalent, be my guest).

include/psa/crypto_extra.h
gilles-peskine-arm: Probably the result of having edited an earlier draft. Does it matter?

tests/psa_helpers.function
gilles-peskine-arm: Indeed. The message has to be a string literal because the test infrastructure stores a pointer to it which it prints and never frees. Within this constraint, do you have a better suggestion? This is only a matter of reporting failures, and most tests will only have one failure to report in practice.

AndrzejKurek: I would suggest concatenating the errors, but since this is for test purpose only, it's not a blocker.

library/psa_crypto_slot_management.c
```diff
@@ -232,4 +232,36 @@ psa_status_t psa_close_key( psa_key_handle_t handle )
     return( psa_wipe_key_slot( slot ) );
 }
 
+void mbedtls_psa_get_stats( mbedtls_psa_stats_t *stats )
+{
+    psa_key_handle_t key;
+    memset( stats, 0, sizeof( *stats ) );
+    for( key = 1; key <= PSA_KEY_SLOT_COUNT; key++ )
+    {
+        psa_key_slot_t *slot = &global_data.key_slots[key - 1];
```
AndrzejKurek: cache_slots are not set here. Should they?

tests/psa_helpers.function
```diff
@@ -32,6 +32,36 @@
...
+     * break on it when debugging a failure. */
+    if( msg != NULL && test_info.failed == 0 )
+        test_fail( msg, line, file );
+
+    mbedtls_psa_crypto_free( );
+}
+
+/** Shut down the PSA subsystem. Expect a clean shutdown, with no slots
+ * in use.
+ */
```
AndrzejKurek: This comment could end in the previous line, to be compliant with the format used in other places.

tests/psa_helpers.function
```diff
@@ -48,12 +48,28 @@ static void test_helper_psa_done( int line, const char *file )
...
+     * information. Do keep the stats lookup above, because it can be
+     * convenient to break on it when debugging a failure. */
     if( msg != NULL && test_info.failed == 0 )
         test_fail( msg, line, file );
 
+    return( msg == NULL );
+}
+
+/** Check that no PSA slots are in use.
+ */
```
AndrzejKurek: This comment could also end in the previous line.

tests/suites/test_suite_psa_crypto_slot_management.function
```diff
@@ -62,6 +64,58 @@ static void psa_purge_key_storage( void )
 #define TEST_USES_KEY_ID( key_id ) ( (void) ( key_id ) )
 #endif /* MBEDTLS_PSA_CRYPTO_STORAGE_C */
 
+/** Apply \p close_method to invalidate the specified handles:
+ * close it, destroy it, or do nothing;
+ */
```
AndrzejKurek: This comment could also end in the previous line.

tests/suites/test_suite_psa_crypto_slot_management.function
```diff
@@ -62,6 +64,58 @@ static void psa_purge_key_storage( void )
...
+static int invalidate_handle( close_method_t close_method,
+                              psa_key_handle_t handle )
+{
+    switch( close_method )
+    {
+        case CLOSE_BY_CLOSE:
```
AndrzejKurek: I'm not sure if we shouldn't mark intentional fallthroughs somehow, I can recall that some static analysis required that, but maybe not for MbedTLS.

tests/psa_helpers.function
gilles-peskine-arm: Looking around, the house style is to put `*/` on a separate line for multiline doxygen comments. We aren't completely consistent about it, but this one is formatted like the majority.

tests/suites/test_suite_psa_crypto_slot_management.function
gilles-peskine-arm: There's no fallthrough here, just two cases following each other. A fallthrough does require a comment. I recall the compiler settings we use on CI make it very picky about what this comment must contain.

library/psa_crypto_slot_management.c
gilles-peskine-arm: I put `cache_slots` in for future use. We don't have them currently, so they stay at 0.

tests/psa_helpers.function
gilles-peskine-arm: Unfortunately, we can't concatenate the errors unless we provide a string literal for each possible combination of issues. It would be nice, yes, but I don't think it's worth it to make major changes to test failure reporting just for that.

gilles-peskine-arm:
I [rebased](https://github.com/ARMmbed/mbed-crypto/compare/c5b3bd6b6e2ef6db5a7dedc971b24f8eea199aa7..dd413d3c928e8941ea4d74ff89fd4e1632f4ed23) on top of the merge of the precursor PR #131. The only changes were in files that this PR doesn't change. The previous version is at https://github.com/gilles-peskine-arm/mbed-crypto/tree/mbedtls-stats-api-2 if needed.

gilles-peskine-arm:
The crypto CI passed. TLS testing failed but that was expected.

tests/psa_helpers.function
```diff
@@ -48,12 +48,28 @@ static void test_helper_psa_done( int line, const char *file )
...
+#define ASSERT_PSA_PRISTINE( )                                    \
+    do                                                            \
+    {                                                             \
+        if( ! test_helper_is_psa_pristine( __LINE__, __FILE__ ) ) \
```
k-stachowiak: The space between the negation operator and the function name should probably not be there.

tests/psa_helpers.function
gilles-peskine-arm: I have this for the other `.function` files:
```
(add-to-list 'auto-mode-alist '("/suites/[^/]+\\.function\\'" . c-mode))
```
I could do the same for `tests/*.function` but it's very generic, `suites` is a lot less common.

Ideally I'd prefer to do this from a `.dir-locals.el` file, but unfortunately you can't set the mode of a specific file in `.dir-locals.el`.

tests/suites/test_suite_psa_crypto.function
gilles-peskine-arm: I'm open to renaming this file to `.inc` or `.c`. It isn't used directly by anything that isn't in this PR yet so it's still very cheap to change.

There is a precedent for a file that is only meant to be #included: `programs/psa/psa_constant_names_generated.c`. AFAIR nobody remarked on the extension at that time. So go with `.c` here?

tests/psa_helpers.function
gilles-peskine-arm: It's not the first time this comes up. We put a space after `!` about half the time. I like it because it makes the negation operator more visible.

tests/suites/test_suite_psa_crypto.function
k-stachowiak: Since it's cheap, and there's a precedent, I'd go for it!

tests/psa_helpers.function
k-stachowiak: I don't intend to object to this personally, I just think it is not in line with how the library has been developed so far, hence I pointed it out.

tests/suites/test_suite_psa_its.function
```diff
@@ -1,6 +1,8 @@
 /* BEGIN_HEADER */
 #include "../library/psa_crypto_its.h"
 
+#define PSA_ASSERT( expr ) TEST_EQUAL( ( expr ), PSA_SUCCESS )
```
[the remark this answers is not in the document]
gilles-peskine-arm: I agree that the way I did it is not ideal, but I don't like your proposal better than what I did. I find it less clear because it works by general case and exception and it has negations of negations.

I think `psa_helpers` should be broken up into `psa_helpers` with only `#include "psa/error.h" and a definition of `PSA_ASSERT`, and `psa_crypto_helpers` with `#include "psa/crypto.h"` and all the rest. And perhaps even a `psa_crypto_se_driver_helpers` for secure element test drivers and so on. But this feels out of proportion with what we're doing now, and we don't even have `psa/error.h` yet.

tests/suites/test_suite_psa_crypto.function
gilles-peskine-arm: `.c` is not so cheap because it means changing makefiles not do delete `*.c` on cleanup.

@AndrzejKurek I'm warming up to `.h`. The file contains function definitions, but they're still pretty small functions. `.h` isn't the right thing, but it's the easy thing.

gilles-peskine-arm: I've gone with `.h`.

tests/Makefile
gilles-peskine-arm: Yes, I do. https://github.com/ARMmbed/mbed-crypto/pull/132/commits/1d10257d215fc2ea366e7b6f15b532e2f40504b9

PR #133, opened 2019-05-31
Allow DHM selftest to run if MBEDTLS_PEM_PARSE_C is unset

hanno-arm:
__Issue:__ If `MBEDTLS_PEM_PARSE_C` is unset, the DHM selftest fails because it uses PEM encoded test data.

This PR fixes this by providing the DER encoded test data instead in case `MBEDTLS_PEM_PARSE_C` is unset.


hanno-arm:
Yes, see https://github.com/ARMmbed/mbed-crypto/blob/development/library/dhm.c#L473

Patater:
> Will `mbedtls_dhm_parse_dhm()` accept non-PEM input? How does it know?

I see `mbedtls_dhm_parse_dhm()` will assume the input is raw if `MBEDTLS_PEM_PARSE_C` is unset.

PR #138, opened 2019-06-04
psa: Add NV seed as an entropy source when needed

Patater:
When MBEDTLS_PSA_INJECT_ENTROPY is used, we now require also defining
MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES. When
MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES is defined, we do not add entropy
sources by default. This includes the NV seed entropy source, which the
PSA entropy injection API is built upon.

The PSA entropy injection feature depends on using NV seed as an entropy
source. Add NV seed as an entropy source for PSA entropy injection.

Fixes e3dbdd8d9082 ("Gate entropy injection through a dedicated configuration option")

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -4639,6 +4639,15 @@ psa_status_t psa_crypto_init( void )
 
     /* Initialize the random generator. */
     global_data.entropy_init( &global_data.entropy );
+#if defined(MBEDTLS_PSA_INJECT_ENTROPY) && \
+    defined(MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES)
+    /* The PSA entropy injection feature depends on using NV seed as an entropy
+     * source. Add NV seed as an entropy source for PSA entropy injection. */
+    mbedtls_entropy_add_source( &global_data.entropy,
+                                mbedtls_nv_seed_poll, NULL,
+                                MBEDTLS_ENTROPY_BLOCK_SIZE,
+                                MBEDTLS_ENTROPY_SOURCE_STRONG );
+#endif
     mbedtls_ctr_drbg_init( &global_data.ctr_drbg );
     global_data.rng_state = RNG_INITIALIZED;
     status = mbedtls_to_psa_error(


library/psa_crypto.c
```diff
@@ -4639,6 +4639,14 @@ psa_status_t psa_crypto_init( void )
 
     /* Initialize the random generator. */
     global_data.entropy_init( &global_data.entropy );
+#if defined(MBEDTLS_PSA_INJECT_ENTROPY)
```
yanesca: I think we should make it conditional on the absence of `MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES`, because otherwise in some configurations we add the NVSEED as an entropy source twice. In the best case it means slower operation, extra cost of battery and extra wear on storage, in the worst it might have some unforeseen side effects as well.

Or alternatively we can make the requirement of having `MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES` as a prerequisite for `MBEDTLS_PSA_INJECT_ENTROPY` by adding it to `check-config.h`.

Also looking at the current condition in `check-config.h` I would expect either `MBEDTLS_ENTROPY_NV_SEED` or `MBEDTLS_PSA_CRYPTO_STORAGE_C` be sufficient to support `MBEDTLS_PSA_INJECT_ENTROPY` and that we don't need both. Is this the case?

gilles-peskine-arm: `check_config.h` does forbid `MBEDTLS_PSA_INJECT_ENTROPY` together with `MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES`. However, I agree with adding `&& !defined(MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES)` here, because 1. not everyone includes `check_config.h` and 2. it's more robust in case we do change `check_config.h`.

We need both `MBEDTLS_ENTROPY_NV_SEED` for the hook to read the seed file and `MBEDTLS_PSA_CRYPTO_STORAGE_C` to write the seed. This is messy and could be simplified, but I don't think it's simple to simplify it. I'd prefer to completely bypass `entropy.c` and use the PSA entropy module, but the PSA entropy module is yet to be written.

yanesca: Thanks, I can see it now, I was looking at the wrong branch.

Patater:
Rebased to add explicit dependency on `MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES`

k-stachowiak:
By the way of approving this PR I confirm that I checked this change on a K64F, and the `mbed-os-example-mbed-crypto` application works out of the box now. I tested it by manually applying the patch to the `mbed-os` copy cloned and build by `mbed-cli` in the example's directory.

PR #139, opened 2019-06-04
Make DES self-test faster, and fix a typo

Patater:


gilles-peskine-arm:
If you change what the test does, you need to change the expected result.

Patater:
Force pushed to update expected test results as well as to reduce iteration counts in the CBC test (previously only ECB test had iteration count reduced).

Patater:
Testing on head is failing because this PR is based on a very old version of the development branch. Merge result is OK, history after merge is clean, and that's good enough.

PR #145, opened 2019-06-10
psa: Use PSA as an entropy source on PSA NSPEs

Patater:
For NSPEs that still run Mbed TLS DRBG code, without any other source of
NS entropy, it's useful to use a PSA SPE as a source of entropy. For
example, a PSA SPE may have exclusive access to a hardware TRNG, leaving
the PSA NSPE with no source of entropy. In order to provide backwards
compatibility with applications that use Mbed TLS DRBG code, make the
PSA random generator be a source of entropy for the PSA NSPE.

### DIFF
--- a/include/mbedtls/entropy_poll.h
+++ b/include/mbedtls/entropy_poll.h
@@ -54,6 +54,15 @@ extern "C" {
                                 unsigned char *output, size_t len, size_t *olen );
 #endif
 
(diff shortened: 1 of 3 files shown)


Patater:
We can use platform entropy poll instead, like a Linux system would, if this isn't otherwise used on Mbed OS already.

PR #148, opened 2019-06-13
Make test suites compatible with #include <assert.h>

gilles-peskine-arm:
This is harmless but useless in itself, but needs to be applied to the crypto submodule before https://github.com/ARMmbed/mbedtls/pull/2697 can be merged.


Patater:
Merge testing failed with known flaky DTLS test `m->m dtls1_2,no TLS-DHE-PSK-WITH-AES-256-GCM-SHA384 .................... FAIL`. Head testing was OK.

PR #144, opened 2019-06-07
Fix full build including non-boolean with Asan: crypto part

gilles-peskine-arm:
This pull requests fixes some issues in test suites that we hadn't caught before because they aren't tested:
* ~Enabling `MBEDTLS_CHECK_PARAMS` as suggested in `config.h` doesn't work (https://github.com/ARMmbed/mbedtls/issues/2598)~ [split out in https://github.com/ARMmbed/mbed-crypto/pull/148]
* ~Fix undefined behavior in HAVEGE (https://github.com/ARMmbed/mbedtls/issues/2598)~ [split out in https://github.com/ARMmbed/mbed-crypto/pull/149]
* Fix a call of memset(NULL,0)
* Make some entropy tests more robust.
* Minor improvements in PSA tests.

This PR does not add CI jobs that would detect those issues. This will come through PR in the mbedtls repository for which this PR is a prerequisite: https://github.com/ARMmbed/mbedtls/pull/2684 (or in several PRs that 2684 may be split into).

Backports: https://github.com/ARMmbed/mbedtls/pull/2754 https://github.com/ARMmbed/mbedtls/pull/2755

PR #149, opened 2019-06-14
Fix misuse of signed ints in the HAVEGE module

gilles-peskine-arm:
The elements of the HAVEGE state are manipulated with bitwise operations, with the expectations that the elements are 32-bit unsigned integers (or larger). But they are declared as `int`, and so the code has undefined behavior. Clang with Asan correctly points out some shifts that reach the sign bit.

Since these are supposed to be 32-bit unsigned integers, declare them as `uint32_t`.

This is technically an API break, since the type `mbedtls_havege_state` is exposed in a public header. However normal applications should not be affected.

Fix https://github.com/ARMmbed/mbedtls/issues/2598

PR extracted from https://github.com/ARMmbed/mbed-crypto/pull/144 which combined this with other things needed to pass Clang+Asan with all configuration options.

When this PR is merged, it'll break the build in mbedtls due to the incompatible change in `havege.h` (this is what I mean by the label "needs: preceding PR"). https://github.com/ARMmbed/mbedtls/pull/2699 will be needed to repair the build, so the merge of these two PRs should be coordinated.

### DIFF
--- a/library/havege.c
+++ b/library/havege.c
@@ -38,6 +38,7 @@
 #include "mbedtls/timing.h"
 #include "mbedtls/platform_util.h"
 
+#include <stdint.h>
 #include <string.h>
 
 /* ------------------------------------------------------------------------
@@ -54,7 +55,7 @@
  * ------------------------------------------------------------------------
  */
 
-#define SWAP(X,Y) { int *T = (X); (X) = (Y); (Y) = T; }
+#define SWAP(X,Y) { uint32_t *T = (X); (X) = (Y); (Y) = T; }
 
 #define TST1_ENTER if( PTEST & 1 ) { PTEST ^= 3; PTEST >>= 1;
 #define TST2_ENTER if( PTEST & 1 ) { PTEST ^= 3; PTEST >>= 1;
@@ -77,7 +78,7 @@
     PTX = (PT1 >> 18) & 7;                              \
     PT1 &= 0x1FFF;                                      \
     PT2 &= 0x1FFF;                                      \
-    CLK = (int) mbedtls_timing_hardclock();                            \
+    CLK = (uint32_t) mbedtls_timing_hardclock();        \
                                                         \
     i = 0;                                              \
     A = &WALK[PT1    ]; RES[i++] ^= *A;                 \
@@ -100,7 +101,7 @@
                                                         \
     IN = (*A >> (5)) ^ (*A << (27)) ^ CLK;              \
     *A = (*B >> (6)) ^ (*B << (26)) ^ CLK;              \
-    *B = IN; CLK = (int) mbedtls_timing_hardclock();                   \
+    *B = IN; CLK = (uint32_t) mbedtls_timing_hardclock();       \
     *C = (*C >> (7)) ^ (*C << (25)) ^ CLK;              \
     *D = (*D >> (8)) ^ (*D << (24)) ^ CLK;              \
                                                         \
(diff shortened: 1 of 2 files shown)


library/havege.c
```diff
@@ -158,10 +159,11 @@
  */
 static void havege_fill( mbedtls_havege_state *hs )
 {
-    int i, n = 0;
-    int  U1,  U2, *A, *B, *C, *D;
-    int PT1, PT2, *WALK, RES[16];
-    int PTX, PTY, CLK, PTEST, IN;
+    size_t n = 0;
+    unsigned i;
```
Patater: Why should `i` not be `size_t`? It's used to index `RES`, as in `ONE_ITERATION` (`RES[(i - 8) ^ PTY]`), and array indices tend to be `size_t`.

library/havege.c
```diff
@@ -38,6 +38,7 @@
 #include "mbedtls/timing.h"
 #include "mbedtls/platform_util.h"
 
+#include <stdint.h>
```
Patater: Style: Should we bother to include `stdint.h` again as we'd have already included it via including `havege.h`?

library/havege.c
gilles-peskine-arm: `i` only takes values from 0 to 15 and is used in an inner loop, so it's more correct to have it of “the most efficient type” than “the correct type for indexing arrays”. But it isn't worth giving the reader pause for that. The compiler can easily recognize what the actual range of `i` is and optimize accordingly anyway.

library/havege.c
gilles-peskine-arm: I tend to prefer to include headers for what they define, not for what headers they pull in. According to this style, since `havege.c` uses `uint32_t`, it includes `stdint.h`.

Patater: Is this the convention we use in the rest of Mbed TLS as well? I'm happy to use whatever the library already does.

It's slightly more efficient to not `#include` a header if you know the types are already available, as they must since the declarations we use from `havege.h` require it, because the preprocessor doesn't need to read in that file at all. For large projects, and especially on slow filesystems (e.g. Windows, network drives), this can add up to a big savings in compile time. It's not uncommon to have a style that encourages efficiency. (C++ style guides recommending pre-increment operators over post-increment operators in for loops is one example of this.)

library/havege.c
Patater: Why not `uint_fast8_t`?

gilles-peskine-arm: We don't officially require C99 (well, I think we made a blog post that says we do, but we still test with non-C99 compilers). We use `stdint.h` but we only assume that it provides the “basic” types: `uintNN_t`. Are we ok with adding a requirement for `uint_ADJECTIVENN_t`? Should we start using it in more places?

library/havege.c
gilles-peskine-arm: I don't remember that we ever discussed this, and I don't think we're doing or not-doing this with any noticeable consistency.

All our header files have garden-variety include guards. Compilers typically optimize well for that case. In any case, compilation time is the least of our worries. This is not a 100000-file C++ project. Being a security project, we value security, readability and ease of maintenance rather more than runtime performance, let alone build-time performance. Also, if you want to optimize compilation time, the biggest thing by far would be to get rid of cmake!

gilles-peskine-arm:
@k-stachowiak CI fails on Mbed TLS builds. There's an incompatible change in a header file, so the CI won't pass if it runs with one side updated but not the other. The CI passes on https://github.com/ARMmbed/mbedtls/pull/2699 which has the header changes on both sides.

gilles-peskine-arm:
Note that https://github.com/ARMmbed/mbed-crypto/pull/149 and https://github.com/ARMmbed/mbedtls/pull/2699 need to be merged in the same time frame, because they make an incompatible change to a header file. The sequence of operations is as follows:

1. Merge https://github.com/ARMmbed/mbed-crypto/pull/149 into `mbed-crypto/development`. This will break the build of `mbedtls/development` against `mbed-crypto/development`.
2. Change https://github.com/ARMmbed/mbedtls/pull/2699 to update the crypto submodule to the result of the previous step. 
3. Merge https://github.com/ARMmbed/mbedtls/pull/2699 into `mbedtls/development`.

PR #150, opened 2019-06-18
ChangeLog: Add a ChangeLog for Mbed Crypto

Patater:
The commit messages are not clear enough when something is worthy of
mention in the release notes. Re-add a ChangeLog file for Mbed Crypto,
with a different name from the Mbed TLS ChangeLog, to enable easier
merging from Mbed TLS to Mbed Crypto (for things like shared tool
updates).

### DIFF
--- a/ChangeLog_crypto.md
+++ b/ChangeLog_crypto.md
@@ -1,0 +1,114 @@
+
+# Mbed Crypto ChangeLog
+
+
+## Mbed Crypto 1.1.0 released xxxx-xx-xx
+
+### Security
+ * Make `mbedtls_ecdh_get_params` return an error if the second key
+   belongs to a different group from the first. Before, if an application
+   passed keys that belonged to different group, the first key's data was
+   interpreted according to the second group, which could lead to either an
+   error or a meaningless output from `mbedtls_ecdh_get_params`. In the latter
+   case, this could expose at most 5 bits of the private key.
+
+### Features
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm:
If you're going to use the same structure as the Mbed TLS changelog, please use the same format. Making the changelog look marginally nicer when viewed through an interface that renders markdown isn't worth the added burden of coping with two different formats.

If we're going to use a different changelog format, let's switch to one that doesn't result in merge conflicts. Some projects such as [Gitlab](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/doc/development/changelog.md) have switched to one entry per file, collected into a consolidated file when you make a release. How about doing this?

Patater:
We can change the format because we don't expect to copy paste between the two ChangeLogs. This file even has a different name so no merges will happen between the two projects on accident. The only thing we could copy paste between ChangeLogs are tools changes (e.g. new feature in ABI checker), but we generally don't put those into ChangeLogs.

Patater:
The move to markdown was to make turning a ChangeLog into release notes as easy as a copy paste, since release notes are markdown format already.

Patater:
Looking at GitLab, "one file per entry, and then compiling the entries into the overall CHANGELOG.md file during the release process." sounds good to me.

Patater:
We'll do ChangeLog a different way than this.

PR #152, opened 2019-06-18
Enable use of Mbed TLS and Mbed Crypto as a CMake subproject

Patater:
Remove use of CMAKE_SOURCE_DIR in case mbedtls is built from within
another CMake project. Define MBEDTLS_DIR to ${CMAKE_CURRENT_SOURCE_DIR}
in the main CMakeLists.txt file and refer to that when defining target
include paths to enable mbedtls to be built as a sub project.

Fixes https://github.com/ARMmbed/mbedtls/issues/2609

Signed-off-by: Ashley Duncan <ashes.man@gmail.com>
Signed-off-by: Jaeden Amero <jaeden.amero@arm.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,9 @@ else()
     project("mbed TLS" C)
 endif()
 
+# Set the project root directory.
+set(MBEDTLS_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+
 option(ENABLE_PROGRAMS "Build mbed TLS programs." ON)
 
 option(UNSAFE_BUILD "Allow unsafe builds. These builds ARE NOT SECURE." OFF)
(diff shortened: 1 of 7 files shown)


Patater:
Rebased to add MBEDTLS_DIR setting to library/CMakeLists.txt, as that's how Mbed TLS includes Mbed Crypto.

Patater:
Rebased to enable use of tests CMakeList.txt without crypto/CMakeLists.txt being included.

Patater:
> The changes look good to me. I reviewed an analogous patch for Mbed TLS itself ([ARMmbed/mbedtls#2706](https://github.com/ARMmbed/mbedtls/pull/2706)) and I quite liked the fact that that patch had tests. Why is it that we do not have one here? Is this pointless? Is it possible add the test?

It's not pointless. We get similar coverage from our CI system, since we build Mbed Crypto as a submodule of Mbed TLS. However, we can add an explicit test to build a program with Mbed Crypto as a standalone library via a CMake subfolder. I've added a commit to do this.

Patater:
Rebased to add a regression test and to change the minimum CMake version to 2.6, to match the version required by the top level project `CMakeLists.txt`.

andresag01:
@Patater: Thanks for adding the test. I will wait for your answer regarding the test in https://github.com/ARMmbed/mbedtls/pull/2706#discussion_r297176754 before approving your changes.

Patater:
Rebased to move cleanup of the CMake subproject test to `cleanup()`.

Patater:
Rebased to update to latest `development`, in order to avoid test failures due to `assert.h` being incompatible with older versions of tests.

Issue #137, opened 2019-06-04
different behavior in `calloc` on different libraries when trying to allocate 0 bytes


RonEld:
### Description

[man calloc](https://linux.die.net/man/3/calloc) describes the following:
>If nmemb or size is 0, then calloc() returns either NULL, or a unique pointer value that can later be successfully passed to free().

This causes different behavior when calling the library wiht 0 bytes to allocate.
For example:
a call to  `psa_import_key()` will result in an error when `slen` is 0 because in `prepare_raw_data_slotoreoare_data_slot()` we will have:
~~~
    /* Allocate memory for the key */
    raw->bytes = PSA_BITS_TO_BYTES( bits );
    raw->data = mbedtls_calloc( 1, raw->bytes  );
    if( raw->data == NULL )
    {
        raw->bytes = 0;
        return( PSA_ERROR_INSUFFICIENT_MEMORY );
    }
~~~

and a call to `psa_key_derivation()` with the same empty key could result in a failure because in `psa_generator_tls12_prf_setup()`:
~~~
   tls12_prf->key = mbedtls_calloc( 1, key_len + 1 );
    if( tls12_prf->key == NULL )
        return( PSA_ERROR_INSUFFICIENT_MEMORY );
~~~

If using `memory_buffer_alloc` feature, the calloc calls would return `NULL` and not a valid pointer because:
~~~
    if( n == 0 || size == 0 || len / n != size )
        return( NULL );
~~~

There are several ways to fix this:
1. Don't call `mbedtls_calloc()` for a zero length -> This could cause crashes in the library for some cases where the data cannot be zero lengthed
2. Always add one more to the size to allocate -> This adds some additional RAM. One could argue this is insignificant.
3. Have the `memory_buffer_alloc()` return a 4 byte aligned pointer for zero length allocation request. This perhaps will fix the specific use case, but on other platforms, that have some calloc function that returns `NULL`, as it's allowed,the error will remain.
### Issue request type

    [ ] Question
    [ ] Enhancement
    [x] Bug

hanno-arm:
FWIW, in https://github.com/ARMmbed/mbedtls/pull/2478 I use and argue for (2) of always adding a byte.

hanno-arm:
Alternatively, one can of course allocate `len ? len : 1` bytes.

gilles-peskine-arm:
[...]

PR #154, opened 2019-06-20
Update TLS 1.2 PRF to multipart API

yanesca:
The TLS PRF key derivation algorithms (PSA_ALG_TLS12_PRF and PSA_ALG_TLS12_PSK_TO_MS) are only implemented for the old-style PSA key derivation API (psa_key_derivation) that takes all inputs at once. This API was removed in PSA crypto API 1.0 beta1 in favor of a new API that takes inputs step by step. psa_key_derivation is still present in the code but only until the existing tests and the TLS code have been updated.

Goals of this PR:
- Implement PSA_ALG_TLS12_PRF and PSA_ALG_TLS12_PSK_TO_MS for psa_key_derivation_setup and friends.
- Ensure that there is at least one positive test case for each of these two algorithm families that uses the new API.


library/psa_crypto.c
```diff
@@ -4111,14 +4125,123 @@ static psa_status_t psa_key_derivation_tls12_prf_generate_next_block(
...
+    psa_hash_operation_t backup = PSA_HASH_OPERATION_INIT;
+    psa_status_t status, cleanup_status;
+
+    /* We can't be wanting more output after block 0xff, otherwise
+     * the capacity check in psa_key_derivation_output_bytes() would have
+     * prevented this call. It could happen only if the operation
+     * object was corrupted or if this function is called directly
+     * inside the library. */
+    if( tls12_prf->block_number == 0xff )
+        return( PSA_ERROR_BAD_STATE );
```
gilles-peskine-arm: “Can't happen” is more of a `CORRUPTION_DETECTED` error.

library/psa_crypto.c
```diff
@@ -4768,26 +5059,23 @@ static psa_status_t psa_key_derivation_input_raw(
...
-    {
-        case PSA_KEY_DERIVATION_INPUT_LABEL:
-        case PSA_KEY_DERIVATION_INPUT_SALT:
-        case PSA_KEY_DERIVATION_INPUT_INFO:
-            return( psa_key_derivation_input_raw( operation, step,
-                                                  data, data_length ) );
-        default:
-            return( PSA_ERROR_INVALID_ARGUMENT );
-    }
+    if( step == PSA_KEY_DERIVATION_INPUT_SECRET )
```
gilles-peskine-arm: Blacklisting instead of whitelisting inputs that can be direct buffers probably saves a few bytes of code, but it's more fragile. I'd prefer to keep whitelisting.

yanesca: Saving a couple of bytes wasn't the only reason for removing it.

My thinking was that the specific functions called inside `psa_key_derivation_input_internal()` need to check this again anyway. This introduces redundancy that makes maintenance more error prone.

Also checking a whitelist here gives limited value. For example if the user calls this for a TLS 1.2 PRF operation with `PSA_KEY_DERIVATION_INPUT_INFO` we still depend on the specialised input function to catch it. This means that even though we checked it here, we still can't make any useful assumption about this parameter later down in the call stack.

gilles-peskine-arm: Ok, good point.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -4077,45 +4087,105 @@ void key_derivation_init( )
...
+    PSA_ASSERT( psa_crypto_init( ) );
+
+    PSA_ASSERT( psa_key_derivation_setup( &operation, alg ) );
+
+    TEST_EQUAL( psa_key_derivation_set_capacity( &operation, capacity ),
+                expected_status );
+
+exit:
+    psa_key_derivation_abort( &operation );
+    mbedtls_psa_crypto_free( );
```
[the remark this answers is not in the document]
gilles-peskine-arm: #132 is merged now.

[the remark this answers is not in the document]
yanesca: #132 has been merged. I need to rebase before I can update this. Would you prefer me to do it now or rather after having all other review and rework finished?

yanesca:
@gilles-peskine-arm I hope this helps a little with reviewing the test coverage:

Tests adapted in this PR:                                                                         
                                                                                 
- derive_key_policy()                                                            
- derive_setup()                                                                 
- derive_output()                                                                
                                                                                 
Tests added in this PR to maintain test coverage:                                                                           
                                                                                 
- derive_input                                                                   
- derive_set_capacity()                                                        
                                                                                 
Tests not affected by changes in this PR:                                                                    
                                                                                 
- key_derivation_init()                                                          
- test_derive_invalid_key_derivation_tests()                                     
- persistent_key_load_key_from_storage()                                         
                                                                                 
Tests to be migrated in the followup PR:                                                                     
                                                                                 
- exercise_key_derivation_key()                                                  
- derive_full()                                                                  
- derive_key_exercise()                                                          
- derive_key_export()                                                            
- derive_invalid_key_derivation_state()

tests/suites/test_suite_psa_crypto.function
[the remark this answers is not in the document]
gilles-peskine-arm: As far as I'm concerned, now is a good time to rebase.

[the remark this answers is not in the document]
yanesca: I have rebased the PR and fixed this issue in the original commits.

yanesca:
Original branch before the rebase:
https://github.com/yanesca/mbed-crypto/tree/iotcrypt-789-update-tls-prf-to-multipart-pre-rebase

include/psa/crypto_struct.h
```diff
@@ -228,19 +277,17 @@ struct psa_key_derivation_s
     size_t capacity;
     union
     {
-        struct
-        {
-            uint8_t *data;
-            size_t size;
-        } buffer;
+        /* Make the union non-empty even with no supported algorithms. */
+        uint8_t dummy;
```
andresag01: Is this necessary? Why cant we have the union guarded by `#if defined(MBEDTLS_MD_C)`? That way we could remove the dummy

library/psa_crypto.c
```diff
@@ -4413,7 +4564,9 @@ static psa_status_t psa_key_derivation_tls12_prf_setup(
 
     return( PSA_SUCCESS );
 }
+#endif /* PSA_PRE_1_0_KEY_DERIVATION */
 
+#if defined(PSA_PRE_1_0_KEY_DERIVATION)
```
andresag01: These guard and the previous #endif seems a bit redundant given that both code blocks have the same conditional macro checks.

include/psa/crypto_extra.h
```diff
@@ -224,6 +224,17 @@ void mbedtls_psa_get_stats( mbedtls_psa_stats_t *stats );
 psa_status_t mbedtls_psa_inject_entropy(const unsigned char *seed,
                                         size_t seed_size);
 
+/*
+ * If this option is not turned on, then the function `psa_key_derivation()`
+ * is removed.
+ *
```
AndrzejKurek: Minor: commit 683535935 - description typo: `[...] tests regarding key derivation **hasn't** been adapted to [...]` -> `[...] tests regarding key derivation **haven't** been adapted to [...]`

library/psa_crypto.c
```diff
@@ -4554,7 +4692,9 @@ static psa_status_t psa_key_derivation_internal(
...
```
andresag01: Is this conditional and the previous #endif redundant? The both seem to have the same macro conditional

library/psa_crypto.c
```diff
@@ -3902,7 +3902,7 @@ psa_status_t psa_key_derivation_abort( psa_key_derivation_operation_t *operation
     {
         status = PSA_ERROR_BAD_STATE;
     }
-    memset( operation, 0, sizeof( *operation ) );
+    mbedtls_platform_zeroize( operation, sizeof( *operation ) );
```
AndrzejKurek: Typo in commit 32fc7d7b6 - unclosed `(` and probably missing preposition `with` (`contain buffers part`).

tests/suites/test_suite_psa_crypto.function
```diff
@@ -1799,11 +1799,19 @@ void derive_key_policy( int policy_usage,
...
 
-    status = psa_key_derivation( &operation, handle,
-                                 exercise_alg,
-                                 NULL, 0,
-                                 NULL, 0,
-                                 1 );
+    PSA_ASSERT( psa_key_derivation_setup( &operation, exercise_alg ) );
+
+    if( PSA_ALG_IS_TLS12_PRF( exercise_alg ) ||
+            PSA_ALG_IS_TLS12_PSK_TO_MS( exercise_alg ) )
```
AndrzejKurek: Please use curly brackets for multiline operations after an if.

library/psa_crypto.c
```diff
@@ -4111,14 +4125,123 @@ static psa_status_t psa_key_derivation_tls12_prf_generate_next_block(
...
+    if( status != PSA_SUCCESS )
+        goto cleanup;
+    status = psa_hash_clone( &backup, &tls12_prf->hmac.hash_ctx );
+    if( status != PSA_SUCCESS )
+        goto cleanup;
+
+
+cleanup:
+
+    cleanup_status = psa_hash_abort( &backup );
```
AndrzejKurek: `cleanup_status` is only used here. How about:
```suggestion
     if( status == PSA_SUCCESS )
         status = psa_hash_abort( &backup );
     else
         psa_hash_abort( &backup );
         
         return( status );
         
     
        status = cleanup_status;
```

include/psa/crypto_extra.h
```diff
@@ -283,9 +283,6 @@ psa_status_t psa_key_derivation(psa_key_derivation_operation_t *operation,
                                 size_t capacity);
```
AndrzejKurek: Minor: missing word in commit ac032857ed - `it could be potentially empty if MBEDTLS_MD_C` + `is defined.`

library/psa_crypto.c
```diff
@@ -4851,17 +4851,11 @@ psa_status_t psa_key_derivation_input_bytes( psa_key_derivation_operation_t *ope
                                              const uint8_t *data,
                                              size_t data_length )
 {
```
AndrzejKurek: Minor: missing word in commit 6f77d33d176 - `is redundant brings a very little`

library/Makefile
```diff
@@ -11,6 +10,9 @@ LOCAL_LDFLAGS =
 
 ifdef DEBUG
 LOCAL_CFLAGS += -g3
+CFLAGS	?= -O0
```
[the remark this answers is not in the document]
yanesca: Good point, I'll remove this commit from the PR.

include/psa/crypto_struct.h
yanesca: That is an option too. I decided to go with the dummy member, because this way
- we don't need to have ifdefs in the macro initialiser (`PSA_KEY_DERIVATION_OPERATION_INIT`)
- it is easier to introduce new members in the union that depend on other compile time flags
- we already use this solution with other similar structs in the library

library/psa_crypto.c
yanesca: This redundancy is intentional, makes it easier to remove/refactor these functions individually.

library/psa_crypto.c
yanesca: This redundancy is intentional, makes it easier to remove/refactor these functions individually.

tests/suites/test_suite_psa_crypto.function
yanesca: I am sorry, I am probably missing something here, could you please clarify?

library/psa_crypto.c
```diff
@@ -4715,30 +4860,171 @@ static psa_status_t psa_hkdf_input( psa_hkdf_key_derivation_t *hkdf,
...
+        return( PSA_ERROR_INVALID_ARGUMENT );
+
+    /* Quoting RFC 4279, Section 2:
+     *
+     * The premaster secret is formed as follows: if the PSK is N octets
+     * long, concatenate a uint16 with the value N, N zero octets, a second
+     * uint16 with the value N, and the PSK itself.
+     */
+
+    pms[0] = ( data_length >> 8 ) & 0xff;
```
AndrzejKurek: This is fine, but the offset computations are getting a bit lengthy at the end. How about having a dedicated pointer and moving through pms with it, or using a macro similar to `MBEDTLS_ASN1_CHK_ADD`?

library/psa_crypto.c
```diff
@@ -4111,14 +4125,123 @@ static psa_status_t psa_key_derivation_tls12_prf_generate_next_block(
...
+     *
+     * P_hash(secret, seed) = HMAC_hash(secret, A(1) + seed) +
+     *                        HMAC_hash(secret, A(2) + seed) +
+     *                        HMAC_hash(secret, A(3) + seed) + ...
+     *
+     * A(0) = seed
+     * A(i) = HMAC_hash(secret, A(i-1))
+     *
+     * The `psa_tls12_prf_key_derivation` structure saves the block
+     * `HMAC_hash(secret, A(i) + seed)` from which the output
```
AndrzejKurek: I would just add that i equals to block_number.

library/psa_crypto.c
```diff
@@ -4111,14 +4125,123 @@ static psa_status_t psa_key_derivation_tls12_prf_generate_next_block(
...
+    /* We can't be wanting more output after block 0xff, otherwise
+     * the capacity check in psa_key_derivation_output_bytes() would have
+     * prevented this call. It could happen only if the operation
+     * object was corrupted or if this function is called directly
+     * inside the library. */
+    if( tls12_prf->block_number == 0xff )
+        return( PSA_ERROR_CORRUPTION_DETECTED );
+
+    /* We need a new block */
+    ++tls12_prf->block_number;
```
AndrzejKurek: How is a leftover `psa_tls12_prf_key_derivation_t` structure handled after a failure? Since the block number is incremented first, I imagine that it has to be discarded?

yanesca: This is an internal function and we abort the operation on failure up in the call stack before the struct reaches the user. Keeping a consistent state would be indeed more robust, but needs extra checks and operations. The overhead is not big in this instance, but it adds up and can mean a considerable difference in code size if we want to do it everywhere.

AndrzejKurek: I'm fine with that, I just wanted to make sure that it's a conscious decision that the struct cannot be reused.

library/psa_crypto.c
yanesca: Good idea!

tests/suites/test_suite_psa_crypto.function
AndrzejKurek: Sure. I'm not sure if it's written down anywhere, but I believe that if the operation that is to be performed after an if is multi-line, curly brackets are preferred to maintain visibility. See here: https://github.com/ARMmbed/mbed-crypto/blob/54b8eabd76ddca2207d094f4d7d24398f016486c/library/psa_crypto.c#L3462

library/psa_crypto.c
yanesca: I thought about that, but my intuition was that it would take more code size. Now I wrote a toy example to compare them and this is what I've got:

With temporary variable:
```asm
00000000 <test1>:
   0:   b510            push    {r4, lr}
   2:   0004            movs    r4, r0
   4:   4803            ldr     r0, [pc, #12]   ; (14 <test1+0x14>)
   6:   f7ff fffe       bl      0 <printf>
   a:   2c00            cmp     r4, #0
   c:   d000            beq.n   10 <test1+0x10>
   e:   0020            movs    r0, r4
  10:   bd10            pop     {r4, pc}
  12:   46c0            nop                     ; (mov r8, r8)
  14:   00000000        .word   0x00000000
```

And without:
```asm
00000018 <test2>:
  18:   b510            push    {r4, lr}
  1a:   1e04            subs    r4, r0, #0
  1c:   d105            bne.n   2a <test2+0x12>
  1e:   4805            ldr     r0, [pc, #20]   ; (34 <test2+0x1c>)
  20:   f7ff fffe       bl      0 <printf>
  24:   0004            movs    r4, r0
  26:   0020            movs    r0, r4
  28:   bd10            pop     {r4, pc}
  2a:   4803            ldr     r0, [pc, #12]   ; (38 <test2+0x20>)
  2c:   f7ff fffe       bl      0 <puts>
  30:   e7f9            b.n     26 <test2+0xe>
  32:   46c0            nop                     ; (mov r8, r8)
  34:   00000000        .word   0x00000000
  38:   00000006        .word   0x00000006
```

The temporary variable seems to save 3 instructions (6 byte of code size when using thumb, 12 bytes otherwise). Therefore I think it is better to keep this as it is.

programs/Makefile
```diff
@@ -19,6 +18,9 @@ endif
 
 ifdef DEBUG
 LOCAL_CFLAGS += -g3
+CFLAGS	?= -O0
```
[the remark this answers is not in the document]
yanesca: Please see my earlier comment

tests/Makefile
```diff
@@ -23,6 +22,9 @@ endif
 
 ifdef DEBUG
 LOCAL_CFLAGS += -g3
+CFLAGS	?= -O0
+else
+CFLAGS	?= -O2
```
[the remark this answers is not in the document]
yanesca: Please see my earlier comment

tests/suites/test_suite_psa_crypto.function
yanesca: Didn't know that. Thanks, it is indeed more readable like that.

yanesca:
@andresag01, @AndrzejKurek Thank you for the review! I believe I have addressed all of your comments and the PR is ready for review again.

include/psa/crypto_struct.h
gilles-peskine-arm: It's just a coincidence that all the KDF we have now use a hash internally. We're likely to add something else soon, such as NIST KDF built on CMAC.

library/Makefile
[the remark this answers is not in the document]
gilles-peskine-arm: I'd prefer `DEBUG` to set `-O0` but it doesn't belong in this PR.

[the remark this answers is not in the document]
andresag01: @gilles-peskine-arm: Interesting, why is that? IMHO it only causes problems when debugging because the compiled code is extremely different between -O2 and -O0. Also, I am not sure people rely on -O0 much because the compiled code is really poor quality.

include/psa/crypto_struct.h
andresag01: ok

library/Makefile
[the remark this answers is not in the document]
yanesca: I personally prefer `-O0` because most of the bugs are present there as well and it allows to debug the C source instead of raw assembly. Of course in some nasty cases the bug is not reproducible with `-O0`, but so far those have been very rare.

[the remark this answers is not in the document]
gilles-peskine-arm: If you do printf debugging then there's no point in changing `-O` when compiling for debugging. But we don't do that except sometimes for TLS. When you're using a debugger, `-O0` is a lot easier to work with: stepping in code follows the code, breakpoints trigger when you expect them… I can't remember a single bug in Mbed TLS that went away with `-O0` and that Asan or Valgrind didn't reveal.

yanesca:
@gilles-peskine-arm I thought about this and I was conflicted back then, when I added this commit, but I decided not to add the additional check. 

My thinking was that it is two checks - the memcopy needs one as well - each check is a test and a branch. That is 4 instructions, which means 8-16 bytes more code (now I can see that returning early it can be done with a single check halving the code size penalty, but still).

As you mentioned, this does not occur the TLS use case and these are TLS specific functions. I admit that I am guilty in the sin of premature optimisation, but my preferred solution would be to document it properly instead of adding checks.

What do you think?

gilles-peskine-arm:
I think we should support it because some protocols use KDFs with zero-size inputs. It's already come up, for example EAP uses the TLS PRF with an empty `secret` (https://github.com/ARMmbed/mbed-crypto/issues/137#issuecomment-498607672) to generate an nonce.

Patater:
Most tests are passing as expected, but the API checking script is complaining. Could you take a look please, @yanesca?

Not merging this PR until we have an issue to track fixing zero-length seed or label.

yanesca:
I have pushed a new commit fixing the zero length allocation issue. @gilles-peskine-arm , @AndrzejKurek could you please review the new commit?

@Patater The API checking script is complaining, because this PR cleans up an obsolete feature and breaks the API. This is intentional.

Patater:
All expected tests passing (API change intentional, testing as submodule of TLS fails on the API branch intentionally for now), two approved reviewers, merging.

PR #160, opened 2019-07-04
Iotcrypt 790 update tests to multipart key derivation

yanesca:
The legacy function` psa_key_derivation` in Mbed Crypto is from an early draft of the PSA Crypto API. It was superseded in API 1.0 beta1 by an API with multiple input steps: `psa_key_derivation_setup` and friends.

Goal of this PR:

- Update `key_ladder_demo` to the new API.
- Update the tests to the new API, with at least equivalent coverage.


gilles-peskine-arm:
The ABI checked script failed with an empty report. This looks like https://github.com/ARMmbed/mbedtls/issues/2672 whose fix hasn't made it to the psa-api-1.0-beta branch yet, but that bug is because removed symbols aren't reported, yet this PR doesn't remove any symbols. https://github.com/ARMmbed/mbed-crypto/pull/154 does remove symbols by design but it was merged before this PR was made.

yanesca:
This PR only changes tests and a sample application. Shouldn't the API/ABI script only check changes in the library?

Patater:
ABI/API checker will complain if your branch is different (older or newer) in API than what's in the target branch. The API/ABI tests are passing at https://jenkins-internal.mbed.com/blue/organizations/jenkins/mbed-crypto-pr/detail/PR-160-merge/1/pipeline/174 (this is the merge result). It would fail on the head test if your branch is behind in API.

PR #162, opened 2019-07-09
Unsupported feature support

RonEld:
This PR adds support for `MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED` in tests and applications.
This supersedes the crypto part of https://github.com/ARMmbed/mbedtls/pull/2124

### DIFF
--- a/library/aria.c
+++ b/library/aria.c
@@ -38,9 +38,8 @@
 #include <string.h>
 
 #if defined(MBEDTLS_SELF_TEST)
-#if defined(MBEDTLS_PLATFORM_C)
 #include "mbedtls/platform.h"
-#else
+#if !defined(MBEDTLS_PLATFORM_C)
 #include <stdio.h>
 #define mbedtls_printf printf
 #endif /* MBEDTLS_PLATFORM_C */
--- a/library/aria.c
+++ b/library/aria.c
@@ -938,7 +938,7 @@ static const uint8_t aria_test2_ctr_ct[3][48] =         // CTR ciphertext
  */
 int mbedtls_aria_self_test( int verbose )
 {
-    int i;
+    int i, ret = 0;
     uint8_t blk[MBEDTLS_ARIA_BLOCKSIZE];
     mbedtls_aria_context ctx;
 
@@ -960,15 +960,45 @@ int mbedtls_aria_self_test( int verbose )
         /* test ECB encryption */
         if( verbose )
             mbedtls_printf( "  ARIA-ECB-%d (enc): ", 128 + 64 * i );
-        mbedtls_aria_setkey_enc( &ctx, aria_test1_ecb_key, 128 + 64 * i );
-        mbedtls_aria_crypt_ecb( &ctx, aria_test1_ecb_pt, blk );
+        ret = mbedtls_aria_setkey_enc( &ctx, aria_test1_ecb_key, 128 + 64 * i );
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
+        {
+            if( verbose != 0 )
+                mbedtls_printf( "skipped\n" );
+            continue;
+        }
+        else if( ret != 0 )
+        {
+            return( 1 );
+        }
+        ret = mbedtls_aria_crypt_ecb( &ctx, aria_test1_ecb_pt, blk );
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
+        {
+            if( verbose )
+                mbedtls_printf( "skipped\n" );
+            continue;
+        }
+        else if( ret != 0 )
+        {
+            return( 1 );
+        }
         if( memcmp( blk, aria_test1_ecb_ct[i], MBEDTLS_ARIA_BLOCKSIZE ) != 0 )
             ARIA_SELF_TEST_IF_FAIL;
 
         /* test ECB decryption */
         if( verbose )
             mbedtls_printf( "  ARIA-ECB-%d (dec): ", 128 + 64 * i );
-        mbedtls_aria_setkey_dec( &ctx, aria_test1_ecb_key, 128 + 64 * i );
+        ret = mbedtls_aria_setkey_dec( &ctx, aria_test1_ecb_key, 128 + 64 * i );
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
+        {
+            if( verbose )
+                mbedtls_printf( "skipped\n" );
+            continue;
+        }
+        else if( ret != 0 )
+        {
+            return(1);
+        }
         mbedtls_aria_crypt_ecb( &ctx, aria_test1_ecb_ct[i], blk );
         if( memcmp( blk, aria_test1_ecb_pt, MBEDTLS_ARIA_BLOCKSIZE ) != 0 )
             ARIA_SELF_TEST_IF_FAIL;
@@ -985,22 +1015,62 @@ int mbedtls_aria_self_test( int verbose )
         /* Test CBC encryption */
         if( verbose )
             mbedtls_printf( "  ARIA-CBC-%d (enc): ", 128 + 64 * i );
-        mbedtls_aria_setkey_enc( &ctx, aria_test2_key, 128 + 64 * i );
+        ret = mbedtls_aria_setkey_enc( &ctx, aria_test2_key, 128 + 64 * i );
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
+        {
+            if( verbose )
+                mbedtls_printf( "skipped\n" );
+            continue;
+        }
+        else if( ret != 0 )
+        {
+            return(1);
+        }
         memcpy( iv, aria_test2_iv, MBEDTLS_ARIA_BLOCKSIZE );
         memset( buf, 0x55, sizeof( buf ) );
-        mbedtls_aria_crypt_cbc( &ctx, MBEDTLS_ARIA_ENCRYPT, 48, iv,
-            aria_test2_pt, buf );
+        ret = mbedtls_aria_crypt_cbc( &ctx, MBEDTLS_ARIA_ENCRYPT, 48, iv,
+                                      aria_test2_pt, buf );
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
+        {
+            if( verbose )
+                mbedtls_printf( "skipped\n" );
+            continue;
+        }
+        else if( ret != 0 )
+        {
+            return(1);
+        }
         if( memcmp( buf, aria_test2_cbc_ct[i], 48 ) != 0 )
             ARIA_SELF_TEST_IF_FAIL;
 
         /* Test CBC decryption */
         if( verbose )
             mbedtls_printf( "  ARIA-CBC-%d (dec): ", 128 + 64 * i );
-        mbedtls_aria_setkey_dec( &ctx, aria_test2_key, 128 + 64 * i );
+        ret = mbedtls_aria_setkey_dec( &ctx, aria_test2_key, 128 + 64 * i );
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
+        {
+            if( verbose )
+                mbedtls_printf( "skipped\n" );
+            continue;
+        }
+        else if( ret != 0 )
+        {
+            return(1);
+        }
         memcpy( iv, aria_test2_iv, MBEDTLS_ARIA_BLOCKSIZE );
         memset( buf, 0xAA, sizeof( buf ) );
-        mbedtls_aria_crypt_cbc( &ctx, MBEDTLS_ARIA_DECRYPT, 48, iv,
-            aria_test2_cbc_ct[i], buf );
+        ret = mbedtls_aria_crypt_cbc( &ctx, MBEDTLS_ARIA_DECRYPT, 48, iv,
+                                      aria_test2_cbc_ct[i], buf );
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
+        {
+            if( verbose )
+                mbedtls_printf( "skipped\n" );
+            continue;
+        }
+        else if( ret != 0 )
+        {
+            return(1);
+        }
         if( memcmp( buf, aria_test2_pt, 48 ) != 0 )
             ARIA_SELF_TEST_IF_FAIL;
     }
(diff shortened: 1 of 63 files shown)


RonEld:
The CI failure is because the changes in `platform.h` and in `hepers.function` related to inclusion of `platform.h` should be merged to the TLs repository as well. (This is done in https://github.com/ARMmbed/mbedtls/pull/2124)

RonEld:
I have rebased to resolved conflicts, fixed a bug in the tests, and added support to new tests

library/pkparse.c
```diff
@@ -545,7 +546,18 @@ static int pk_get_rsapubkey( unsigned char **p,
 
     if( ( ret = mbedtls_rsa_import_raw( rsa, *p, len, NULL, 0, NULL, 0,
                                         NULL, 0, NULL, 0 ) ) != 0 )
-        return( MBEDTLS_ERR_PK_INVALID_PUBKEY );
+    {
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
+        {
+            ret = MBEDTLS_ERR_PK_INVALID_PUBKEY |
+                  MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED;
```
Patater: Can these be or'd together like this? Do users use them as bitfields?

library/pkparse.c
```diff
@@ -901,9 +947,14 @@ static int pk_parse_key_sec1_der( mbedtls_ecp_keypair *eck,
             {
                 /*
                  * The only acceptable failure mode of pk_get_ecpubkey() above
-                 * is if the point format is not recognized.
+                 * is if the point format is not recognized. if the underlying
+                 * platform does not support this specific error,
+                 * we should return that specific error.
                  */
-                if( ret != MBEDTLS_ERR_ECP_FEATURE_UNAVAILABLE )
+                if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
```
Patater: Should this be `(ret & MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED)` if `ret` is a bitfield now?

tests/suites/helpers.function
```diff
@@ -107,15 +108,53 @@ typedef enum
...
  * \param   TEST    The test expression to be tested.
  */
-#define TEST_ASSERT( TEST )                                 \
-    do {                                                    \
-       if( ! (TEST) )                                       \
-       {                                                    \
-          test_fail( #TEST, __LINE__, __FILE__ );           \
-          goto exit;                                        \
-       }                                                    \
+
```
Patater: Don't add this newline

tests/suites/helpers.function
```diff
@@ -107,15 +108,53 @@ typedef enum
...
+#define TEST_ASSERT( TEST )                                                   \
+    do {                                                                      \
+        if( ! (TEST) )                                                        \
+        {                                                                     \
+            set_test_result( #TEST, __LINE__, __FILE__, TEST_RESULT_FAILED ); \
```
Patater: Why not keep using `test_fail()` and `test_skip()`?

tests/suites/helpers.function
```diff
@@ -423,17 +461,11 @@ jmp_buf jmp_tmp;
 /*----------------------------------------------------------------------------*/
 /* Helper Functions */
 
-void test_fail( const char *test, int line_no, const char* filename )
```
Patater: Instead of deleting `test_fail`, we could have it be a static function that calls `set_test_result()`

tests/suites/target_test.function
```diff
@@ -90,18 +90,19 @@ uint8_t receive_byte()
...
+                        greentea_getc(),
+                        greentea_getc(),
+                        greentea_getc(),
+                        greentea_getc(),
+                        greentea_getc(),
+                        greentea_getc(),
+                        greentea_getc()
+                      };
+    const uint8_t c[9] = { c_be[6], c_be[7], c_be[4], c_be[5], c_be[2],
+                           c_be[3], c_be[0], c_be[1], '\0' };
```
Patater: Why does this need to change as part of "Unsupported feature support"

library/aria.c
```diff
@@ -960,15 +959,45 @@ int mbedtls_aria_self_test( int verbose )
...
         if( verbose )
             mbedtls_printf( "  ARIA-ECB-%d (enc): ", 128 + 64 * i );
-        mbedtls_aria_setkey_enc( &ctx, aria_test1_ecb_key, 128 + 64 * i );
-        mbedtls_aria_crypt_ecb( &ctx, aria_test1_ecb_pt, blk );
+        ret = mbedtls_aria_setkey_enc( &ctx, aria_test1_ecb_key, 128 + 64 * i );
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
+        {
+            if( verbose != 0 )
+                mbedtls_printf( "skipped\n" );
+            continue;
```
Patater: We are adding a lot of branches to handle FEATURE_UNSUPPORTED from many places. How will we test this? Our code coverage is going to suffer a lot unless we come up with and write tests to hit each of these new branches. This can also help us find bugs if we don't properly handle FEATURE_UNSUPPORTED errors somewhere.

library/aria.c
```diff
@@ -960,15 +959,45 @@ int mbedtls_aria_self_test( int verbose )
...
+        }
+        else if( ret != 0 )
+        {
+            return( 1 );
+        }
+        ret = mbedtls_aria_crypt_ecb( &ctx, aria_test1_ecb_pt, blk );
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
+        {
+            if( verbose )
+                mbedtls_printf( "skipped\n" );
```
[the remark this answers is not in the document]
RonEld: I agree, but we can't predict when a platform will return this error, and we allow returning this error for any crypto operation

library/chacha20.c
```diff
@@ -550,7 +549,14 @@ int mbedtls_chacha20_self_test( int verbose )
...
 
-        ASSERT( 0 == ret, ( "error code: %i\n", ret ) );
+
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
+        {
+            if( verbose != 0 )
+                mbedtls_printf( "skipped\n" );
+            continue;
+        }
+        else ASSERT( 0 == ret, ( "error code: %i\n", ret ) );
```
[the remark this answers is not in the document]
RonEld: ok

library/pkparse.c
RonEld: Probably. But I f remember correct, when I tested this, in this specific branch, it was always `MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED`. But better to check the low leverl error

library/pkparse.c
RonEld: The Or here is basically the high level and the low level. If I remember correct, the `MBEDTLS_ERR_PK_INVALID_PUBKEY` is still used in caller functions

tests/suites/helpers.function
RonEld: ok

tests/suites/helpers.function
RonEld: in order not to duplicate code.
I believe this came from [this comment](https://github.com/ARMmbed/mbedtls/pull/2124#discussion_r241462535)

tests/suites/target_test.function
RonEld: It doesn't but this bug fix was discovered as part of this PR, so I put it as a seperate commit.

library/aria.c
RonEld: I agree, but as mentioned, since we allow this eror to be returned in every function, I don't see another option.
My suggestion was to have "dummy targets" that will return this error in different functions

tests/suites/target_test.function
Patater: OK

tests/suites/helpers.function
RonEld: initially `set_test_result()` was static, but this caused compilation warnings, as declared functions that are not used, in modules that are configured out. The linker probably doesn't remove static functions.
In addition, since I thought putting one single function would make it cleaner

library/aria.c
Patater: Yes, we need dummy targets (more commonly known as test mocks) in order to complete this feature.

Patater:
Needs to pass CI

library/pkparse.c
RonEld: I am not sure this is needed. The low level driver could return `MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED`, which in this case it is returned from `pk_get_ecpubkey()`.  In this check, if it is unsupported, we modify the code to `MBEDTLS_ERR_PK_KEY_INVALID_FORMAT | MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED`

library/pkparse.c
gilles-peskine-arm: In principle, yes: an error code from a “high-level” module of Mbed TLS can be either a high-level error code, or a bitwise-or of a high-level error code and a low-level error code. See `error.h`.

In practice, I've seen bugs before from code expecting a high-level error code and getting one bitwise-ored with a low-level error code. So I wouldn't be surprised if this broke some existing code. We need to document this change clearly and in a way that people will find. This used to be reasonable code (abbreviated), but no longer is:
```
switch (mbedtls_pk_parse_key(…)) {
    case 0: do stuff; break;
    case MBEDTLS_ERR_PK_INVALID_PUBKEY: /*algorithm not supported, try the next key*/ continue;
    default: printf("Fatal error parsing key. Giving up.\n"); exit(1);
}
```

library/pkparse.c
```diff
@@ -545,7 +546,18 @@ static int pk_get_rsapubkey( unsigned char **p,
 
     if( ( ret = mbedtls_rsa_import_raw( rsa, *p, len, NULL, 0, NULL, 0,
                                         NULL, 0, NULL, 0 ) ) != 0 )
-        return( MBEDTLS_ERR_PK_INVALID_PUBKEY );
+    {
+        if( ret == MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED )
```
gilles-peskine-arm: This block of code is repeated. Please make an auxiliary function.

gilles-peskine-arm: I'm a bit lost. When would a function from the `rsa` module return `MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED` vs `MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED | MBEDTLS_ERR_RSA_xxx`?

tests/suites/test_suite_ecp.function
```diff
@@ -1057,9 +1057,6 @@ void mbedtls_ecp_gen_key( int id )
     TEST_ASSERT_RET( mbedtls_ecp_gen_key( id, &key, &rnd_pseudo_rand,
                                           &rnd_info ), 0 );
 
-    TEST_ASSERT_RET( mbedtls_ecp_check_pubkey( &key.grp, &key.Q ), 0 );
```
gilles-peskine-arm: Why have these two lines been removed? It doesn't seem to have anything to do with "Adjust pk tests to TEST_ASSERT_RET".

library/aria.c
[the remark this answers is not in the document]
gilles-peskine-arm: A platform could support a cipher in only one direction. This is most common in the encrypt direction, which is all you need for many modern AEAD modes (e.g. CCM, GCM). This could also be in the decrypt direction only for an accelerator that's only ever meant for CBC decryption, but that's a lot rarer.

tests/suites/helpers.function
```diff
@@ -107,15 +108,52 @@ typedef enum
...
+ *          individual test in a test case, comparing result to an expected
+ *          result given as parameter, and skips the test in case the underlying
+ *          alternative implementation doesn't support the feature.
+ *
+ *          It allows a library function to return a value and return an error
+ *          code that can be tested.
+ *          It receives an expected return code and:
+ *          - If the underlying alternative implementation does not support the feature,
+ *          by returning MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED, the test is skipped.
+ *          - If and fails in case the expression
```
gilles-peskine-arm: Grammar is mistake and I'm not even sure what you were trying to say.

tests/suites/helpers.function
```diff
@@ -107,15 +108,52 @@ typedef enum
...
+ *          results different.
+ *
+ *
+ * \param   TEST    The test expression to be tested.
+ * \param   EXP_RET The expected return code from the function
+ */
```
gilles-peskine-arm: Please document when this function jumps (as is done for `TEST_ASSERT`).

RonEld:
All the failed CI tests are in the tls testing

Patater:
> All the failed CI tests are in the tls testing

Which means if we merge this, TLS tests will start failing when they update their submodule. The TLS tests should be passing before merge, or we need to understand how to make them pass again when the submodule is updated.

RonEld:
@Patater 
I guess https://github.com/ARMmbed/mbedtls/pull/2124 should be merged before this PR then, updating the TLS PR with this branch as the submodule

tests/suites/test_suite_pk.function
```diff
@@ -203,78 +203,78 @@ void valid_parameters( )
     TEST_VALID_PARAM( mbedtls_pk_restart_free( NULL ) );
 #endif
 
-    TEST_ASSERT( mbedtls_pk_setup( &pk, NULL ) ==
-                 MBEDTLS_ERR_PK_BAD_INPUT_DATA );
+    TEST_ASSERT_RET( mbedtls_pk_setup( &pk, NULL ),
```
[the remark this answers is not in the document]
RonEld: >Valid/invalid parameter tests should never reach an alt implementation, should they?

I agree that for the PK layer the rejection should be in the pk layer, so perhaps no need for `TEST_ASSERT_RET` here, but in general, valid\invalid parameter tests should reach an alt implementation.

library/pkparse.c
RonEld: Why would there be two low level errors? I don't see a place where`MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED | MBEDTLS_ERR_RSA_xxx` is returned.
The previous behavior was returning ` MBEDTLS_ERR_PK_INVALID_PUBKEY` for *any* failure, now, in case of `MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED` returned by the rsa module, we still return `MBEDTLS_ERR_PK_INVALID_PUBKEY` but we also add the low level `MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED`

programs/test/benchmark.c
```diff
@@ -843,16 +990,14 @@ int main( int argc, char *argv[] )
...
-
             mbedtls_snprintf( title, sizeof( title ), "ECDSA-%s",
                                               curve_info->name );
+            CHECK_AND_CONTINUE( mbedtls_ecdsa_genkey( &ecdsa, curve_info->grp_id, myrand, NULL ) );
+            ecp_clear_precomputed( &ecdsa.grp );
+
             TIME_PUBLIC( title, "sign",
-                    ret = mbedtls_ecdsa_write_signature( &ecdsa, MBEDTLS_MD_SHA256, buf, curve_info->bit_size,
-                                                tmp, &sig_len, myrand, NULL ) );
+                    CHECK_AND_BREAK( mbedtls_ecdsa_write_signature( &ecdsa, MBEDTLS_MD_SHA256, buf, curve_info->bit_size,
```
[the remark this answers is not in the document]
RonEld: >Breaking here leaves ecdsa unfreed. 

No it won't. THe break is done within [the loop inside `TIME_PUBLIC`](https://github.com/ARMmbed/mbed-crypto/blob/development/programs/test/benchmark.c#L185) macro., going to next line in this code, which frees the context

tests/suites/helpers.function
RonEld: Actually it's a rebase error, I'll fix

gilles-peskine-arm:
Relabeling as "needs: work" because there are merge conflicts. Let's arrange a time when we're available for review soon after this has been rebased, or agree to include a merge commit.

RonEld:
All the failed tests are due to missing declaration of `MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED`. only on Mbed builds, probably because this definition is in a header file that wasn't imported yet to Mbed OS

gilles-peskine-arm:
[Mbed Crypto has been merged back into Mbed TLS](https://github.com/ARMmbed/mbed-crypto/issues/374). Due to the nature of this pull request, it does not make sense to apply it to Mbed Crypto separately. Please bring these changes back into https://github.com/ARMmbed/mbedtls/pull/2124 .

PR #168, opened 2019-07-12
crypto_se_driver: fix a bug in the importing test

AndrzejKurek:


Patater:
@AndrzejKurek @dgreen-arm We don't need this PR anymore, do we? Please re-open if we do.

PR #170, opened 2019-07-12
Use stdint.h types

### LINKED ISSUE #169: Change unsigned char to uint8_t in the API.
### Description

Release 1.0 beta3 of the API Specification states that "A buffer parameter always has the type uint8_t * or const uint8_t *..." However there are six functions in the document

psa_cipher_generate_iv
psa_cipher_set_iv
psa_cipher_update
psa_aead_generate_nonce
psa_aead_set_nonce
psa_aead_update

that specify the buffer type as unsigned char, which is effectively equivalent to uint8_t, but it's good to be consistent.

### Issue request type

    [ ] Question
    [ ] Enhancement
    [x ] Bug

gilles-peskine-arm:
Follow MISRA C 2012 rules by using exact width types from stdint.h.

Fix #169.

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -1853,7 +1853,7 @@ psa_status_t psa_cipher_decrypt_setup(psa_cipher_operation_t *operation,
  * \retval #PSA_ERROR_CORRUPTION_DETECTED
  */
 psa_status_t psa_cipher_generate_iv(psa_cipher_operation_t *operation,
-                                    unsigned char *iv,
+                                    uint8_t *iv,
                                     size_t iv_size,
                                     size_t *iv_length);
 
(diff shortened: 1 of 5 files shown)


gilles-peskine-arm:
CI ok for pr-merge crypto except for the ABI/API check which as expected tells us that some `unsigned char` have changed to `uint8_t`. pr-head has failures due to the branch being older. TLS failures are expected in the API branch. So CI is PASS.

Patater:
Changes made so far are OK. Nothing left in `crypto.h`. I'm not sure that's complete, though.

- Do we want to also update `crypto_extra.h`?
- Do we want to also update `crypto_struct.h` where it gives examples of APIs updated by this PR that use the old types?
- Any other headers to update?

gilles-peskine-arm:
@Patater Good point. I made the same change in implementation-specific and internal APIs (but only psa stuff, not mbedtls stuff). This required a rebase or merge and given that the changes are small I rebased.

PR #173, opened 2019-07-15
Prepare for removing crypto from mbedtls

Patater:
As part of work on removing duplicate crypto files from Mbed TLS, some changes are needed in Mbed Crypto. As part of doing this work, a few minor bugs were discovered and fixed as well.


Patater:
Rebased to avoid removal of psa_util.h (which is needed by PK and Cipher, both of which continue to be part of Mbed Crypto for the time being), to get cpp_dummy_build passing, and to amend the removal of certs.h commit to pass check-generated-files.

Patater:
Rebased to update base to latest development branch.

yanesca:
Looks good to me.

PR #174, opened 2019-07-15
Add mock tests for hooks for secure element drivers

dgreen-arm:
This PR add hooks for generating keys, signing and verification using secure element driver code.


include/psa/crypto_se_driver.h
```diff
@@ -849,26 +849,11 @@ typedef psa_status_t (*psa_drv_se_export_key_t)(psa_key_slot_number_t key,
...
 typedef psa_status_t (*psa_drv_se_generate_key_t)(psa_key_slot_number_t key_slot,
                                                   psa_key_type_t type,
                                                   psa_key_usage_t usage,
-                                                  size_t bits,
-                                                  const void *extra,
-                                                  size_t extra_size,
-                                                  uint8_t *p_pubkey_out,
-                                                  size_t pubkey_out_size,
-                                                  size_t *p_pubkey_length);
+                                                  size_t bits);
```
AndrzejKurek: Why did you drop support for pubkey exporting during key generation?

library/psa_crypto.c
```diff
@@ -3007,6 +3010,24 @@ psa_status_t psa_asymmetric_sign( psa_key_handle_t handle,
...
+        if( drv->asymmetric == NULL ||
+            drv->asymmetric->p_sign == NULL )
+        {
+            status = PSA_ERROR_NOT_SUPPORTED;
+            goto exit;
+        }
+        status = drv->asymmetric->p_sign(
+            slot->data.se.slot_number, alg, hash, hash_length, signature,
+            signature_size, signature_length );
+        goto exit;
```
AndrzejKurek: Do we need this goto?

library/psa_crypto.c
```diff
@@ -3007,6 +3010,24 @@ psa_status_t psa_asymmetric_sign( psa_key_handle_t handle,
...
+#if defined(MBEDTLS_PSA_CRYPTO_SE_C)
+    driver = psa_get_se_driver_entry( slot->lifetime );
+    if( driver != NULL )
+    {
+        const psa_drv_se_t *drv = psa_get_se_driver_methods( driver );
+        if( drv->asymmetric == NULL ||
+            drv->asymmetric->p_sign == NULL )
+        {
+            status = PSA_ERROR_NOT_SUPPORTED;
+            goto exit;
```
AndrzejKurek: Do we need this goto?

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
```diff
@@ -125,12 +114,16 @@ static psa_status_t mock_import( psa_key_slot_number_t slot_number,
...
 
     mock_import_data.called++;
-    mock_import_data.slot_number = slot_number;
-    mock_import_data.lifetime = lifetime;
-    mock_import_data.type = type;
-    mock_import_data.algorithm = algorithm;
-    mock_import_data.usage = usage;
-    mock_import_data.data_length = data_length;
+
+    if( mock_import_data.return_value == PSA_SUCCESS )
```
AndrzejKurek: This should be set regardlesss of the result.

dgreen-arm: Do we want to set all these values even on failure? The test seems to indicate that `type` at least should remain unset if `mock_import_return_value` isn't `PSA_SUCCESS`

library/psa_crypto.c
dgreen-arm: I believe we need this one, as there isn't anything more we can do at this point.

library/psa_crypto.c
dgreen-arm: No, it goes to exit at this point anyway, I'll remove it.

include/psa/crypto_se_driver.h
dgreen-arm: I've fixed this based on the discussion we had.

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
AndrzejKurek: The mock structures only indicate with what arguments was the function called, regardless of any error.

Patater:
@dgreen-arm I've rebased this branch on the latest API branch. Available at https://github.com/Patater/mbed-crypto/tree/psa-se-driver-hooks. Would recommend updating this PR with that branch, changing the target branch of this PR, and continuing the review from there.

dgreen-arm:
There are currently two issues to note with this PR. Firstly, the tests aren't currently being run as `MBEDTLS_PSA_CRYPTO_SE_C` isn't enabled by default. Would adding a test case in all.sh be the best approach for this?
Second, the mock import failure test currently doesn't work with how the mocks are constructed. Either the test or the mock needs updating.

gilles-peskine-arm:
There's an `all.sh` job with SE support enabled since #157 so rebasing on top of that would give you that.

dgreen-arm:
Oh, I hadn't realized that. This is based on #157, so that issue is already solved.

gilles-peskine-arm:
I broke off the hooks for generate/sign/verify in https://github.com/ARMmbed/mbed-crypto/pull/211, without the mock tests, to get a basis to make the whole stack work on real hardware. This PR should be rebased on top of a merge of #211, to add the tests.

library/psa_crypto.c
```diff
@@ -1835,7 +1835,7 @@ psa_status_t psa_import_key( const psa_key_attributes_t *attributes,
     if( driver != NULL )
     {
         const psa_drv_se_t *drv = psa_get_se_driver_methods( driver );
-        size_t bits;
+        size_t bits = PSA_MAX_KEY_BITS + 1;
```
AndrzejKurek: Why is this change needed?

Patater: Please leave a comment in the code explaining the reason, to improve the code's readability. Others will have the same question.

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
```diff
@@ -0,0 +1,207 @@
...
+                                 size_t *bits )
+{
+    (void) drv_context;
+    (void) data;
+    (void) bits;
+
+    mock_import_data.called++;
+    mock_import_data.key_slot = key_slot;
+    mock_import_data.attributes.core.lifetime = attributes->core.lifetime;
+    mock_import_data.attributes.core.type = attributes->core.type;
```
Patater: Why do we need to save the attributes at all? I know we have asserts later on, but that's only testing that the mock saved this data properly, and if the mock doesn't save the data, we wouldn't need a test either. What in the core implementation is being tested if anything by having the mock attributes set here?

Perhaps we are testing that the core will pass the attributes down to the driver properly, in which case this makes sense. However, we should be checking more than just type and lifetime, shouldn't we?

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
```diff
@@ -0,0 +1,207 @@
...
+    psa_set_key_lifetime( &attributes, lifetime );
+    psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_EXPORT );
+    psa_set_key_type( &attributes, PSA_KEY_TYPE_RAW_DATA );
+    PSA_ASSERT( psa_import_key( &attributes,
+                                key_material, sizeof( key_material ),
+                                &handle ) );
+
+    TEST_ASSERT( mock_allocate_data.called == 1 );
+    TEST_ASSERT( mock_import_data.called == 1 );
+    TEST_ASSERT( mock_import_data.attributes.core.type == PSA_KEY_TYPE_RAW_DATA );
```
Patater: In commit "crypto_se_driver: add mock tests", we are testing that the type is set correctly. Should we also test that the lifetime was set correctly?

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
```diff
@@ -36,6 +44,26 @@ static void mock_teardown( void )
...
+                                   size_t *pubkey_length )
+{
+    (void) drv_context;
+    (void) pubkey;
+    (void) pubkey_length;
+
+    mock_generate_data.called++;
+    mock_generate_data.key_slot = key_slot;
+    mock_generate_data.attributes.core.type = attributes->core.type;
+    mock_generate_data.pubkey_size = pubkey_size;
```
Patater: Just confirming, it's not required as part of the driver contract to set pubkey or pubkey_length. Correct? Nothing in core uses those?

If nothing in core uses those, what's the point of having them other than we expect we might want to use them in the future and want API consistency between this and the next driver model version?

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
```diff
@@ -205,3 +233,43 @@ exit:
...
+    psa_set_key_id( &attributes, id );
+    psa_set_key_lifetime( &attributes, lifetime );
+    psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_EXPORT );
+    psa_set_key_type( &attributes, PSA_KEY_TYPE_RAW_DATA );
+    PSA_ASSERT( psa_generate_key( &attributes, &handle ) );
+    TEST_ASSERT( mock_generate_data.called == 1 );
+
+    if( expected_result == PSA_SUCCESS )
+    {
+        PSA_ASSERT( psa_destroy_key( handle ) );
```
Patater: Should we check that allocate was called in commit "crypto_se_driver: add key generation mock and test"?

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
```diff
@@ -244,6 +255,7 @@ void mock_generate( )
     psa_key_handle_t handle = 0;
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
 
+    mock_generate_data.return_value = mock_generate_return_value;
```
Patater: In commit "crypto_se_driver: add an error injection mechanism to the mocks", Should generate also have `mock_allocate_data.return_value = mock_alloc_return_value;` like import does?

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
```diff
@@ -108,6 +117,23 @@ psa_status_t mock_export( psa_drv_se_context_t *context,
...
+                                 size_t data_size,
+                                 size_t *p_data_length )
+{
+    (void) context;
+    (void) p_data;
+    (void) p_data_length;
+
+    mock_export_public_data.called++;
+    mock_export_public_data.slot_number = slot_number;
+    mock_export_public_data.data_size = data_size;
```
Patater: Are we going to check anything about what the core gave us for slot_number or data_size? I don't see the point in saving these unless we look at their values during our test, to ensure the core passed on the information from the top level to the driver level correctly.

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
```diff
@@ -53,12 +63,13 @@ static struct
...
 static void mock_teardown( void )
 {
+    memset( &mock_generate_data, 0, sizeof( mock_generate_data ) );
     memset( &mock_import_data, 0, sizeof( mock_import_data ) );
     memset( &mock_export_data, 0, sizeof( mock_export_data ) );
     memset( &mock_export_public_data, 0, sizeof( mock_export_public_data ) );
+    memset( &mock_sign_data, 0, sizeof( mock_sign_data ) );
     memset( &mock_allocate_data, 0, sizeof( mock_allocate_data ) );
     memset( &mock_destroy_data, 0, sizeof( mock_destroy_data ) );
-    memset( &mock_generate_data, 0, sizeof( mock_generate_data ) );
```
Patater: In commit " crypto_se_driver: add signing mock test", don't move the location of mock_generate_data memsetting.

library/psa_crypto.c
Patater: I see this change was added in commit "Parametrize key bits in import mock test", which doesn't seem right if all one intends to do is updating a test as the title would imply. If the core has to be updated together with the test change, revise the commit message. If the core has a bug, explain it in the commit message and make a change to fix the bug separate from the mock test updating.

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
AndrzejKurek: The idea behind saving all of the parameters (apart from pointers) passed to the mocks was that it's irrelevant for them what the tests check and once the mocks have been created, no real changes to them need to be introduced. If anyone wants to extend the tests and check a different parameter - only a test has to be added/updated.

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
dgreen-arm: Fixed

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
dgreen-arm: Another point in favor of saving the values by default is that we'll get unused-parameter errors if we don't do something with them.

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
dgreen-arm: I've added this, along with a corresponding test.

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
dgreen-arm: Yes we should, I've added it

library/psa_crypto.c
dgreen-arm: I've split this out into a separate commit and added an explanation.

dgreen-arm:
The ABI check is failing as the fix hasn't been merged into the `psa-api-1.0-beta` branch yet, however I've run it locally and there are no changes, as expected.

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
Patater: OK, please make sure we are consistently saving all the values (other than pointers).

library/psa_crypto.c
Patater: Please add a comment in the code about why you chose the value you did.

dgreen-arm: Added a code comment.

dgreen-arm:
I've reworked this to save and check all the attributes in the import and generate mock tests. I've also rebased it onto development to (hopefully) get the PR-head tests passing.

Previous version available here: https://github.com/dgreen-arm/mbed-crypto/tree/psa-se-driver-hooks-3
Updated attributes here, but before rebase here: https://github.com/dgreen-arm/mbed-crypto/tree/psa-se-driver-hooks-4

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
dgreen-arm: I've updated this to check all the set attributes.

tests/suites/test_suite_psa_crypto_se_driver_hal_mocks.function
Patater: It's a bit YAGNI as per our current implementation. However, we feel we need a stable PSA Crypto API and must allow for this capability in the API even if not implemented yet.

PR #178, opened 2019-07-17
New config.h option to make SHA-512 smaller

mpg:
Add `MBEDTLS_SHA512_SMALLER`, parallel to `MBEDTLS_SHA256_SMALLER`.

Sizes, as measured with `arm-none-eabi-gcc -Wall -Wextra -Iinclude -Os -mcpu=cortex-m0plus -mthumb -c library/sha512.c && arm-none-eabi-size sha512.o`:

* without `MBEDTLS_SHA512_SMALLER`: 5691
* with `MBEDTLS_SHA512_SMALLER`: 3411
* saving: 2280 bytes

Speed impact on my laptop's i7, building with GCC -O2 and running `programs/test/benchmark sha512`:

* larger version: 294000 KB/s
* smaller version: 252000 KB/s
* slowdown: -14%

Todo:

- [ ] measure speed impact on target and expand config.h documentation


mpg:
@Patater Thanks for your review! Yes, since the option is not excluded from `config.pl full` it will be tested as part of full-config tests in `all.sh`. (Arguably it may make sense to exclude this option as it's about an alternative implementation, not an additional feature - but I mirrored what was done with the corresponding SHA-256 option, which is included in full configs.)

library/sha512.c
```diff
@@ -256,6 +256,25 @@ int mbedtls_internal_sha512_process( mbedtls_sha512_context *ctx,
...
+        }
+        else
+        {
+            W[i] = S1(W[i -  2]) + W[i -  7] +
+                   S0(W[i - 15]) + W[i - 16];
+        }
+
+        P( A[0], A[1], A[2], A[3], A[4], A[5], A[6], A[7], W[i], K[i] );
+
+        temp1 = A[7]; A[7] = A[6]; A[6] = A[5]; A[5] = A[4]; A[4] = A[3];
```
gilles-peskine-arm: You can save a few more bytes with
```
        temp1 = A[7];
        for( k = 1; k < 8; k++ )
            A[k] = A[k - 1];
        A[0] = temp1;
```

On the other hand, I tried removing the rotation of `A` and running
```
        P( A[(80-i+0)%8], A[(80-i+1)%8], A[(80-i+2)%8], A[(80-i+3)%8], A[(80-i+4)%8], A[(80-i+5)%8], A[(80-i+6)%8], A[(80-i+7)%8],
           W[i], K[i] );
```
but that has almost the same code size as your version. I wonder if there's more to gain there by tweaking `P` and its auxiliary macros to use temporaries for certain calculations.

Patater:
CI on merge result is all passing. Jenkins failed to notify GitHub with the result.

PR #182, opened 2019-07-19
Replace some macros by functions in ecp

gilles-peskine-arm:
Replace some frequently-used macros by inline functions: instead of
calling MOD_{ADD,SUB,MUL} after the mbedtls_mpi_{add,sub,mul}_mpi,
call a function mbedtls_mpi_xxx_mod that does the same.

In the baremetal config, with `gcc -Os -mthumb -mcpu=cortex-m0plus`,
`ecp.o` goes down from 13878 bytes to 12234.

No noticeable performance change for benchmarks on x86_64 with either
`gcc -O2` or `gcc -Os`.

### DIFF
--- a/library/ecp.c
+++ b/library/ecp.c
@@ -1080,6 +1080,18 @@ cleanup:
         INC_MUL_COUNT                                                   \
     } while( 0 )
 
+static inline int mbedtls_mpi_mul_mod( const mbedtls_ecp_group *grp,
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm:
The PR-merge job only failed due because the ABI checking script couldn't run, which is a known bug, fixed after the CI ran. Since this PR obviously didn't change the ABI, we don't need to re-run the CI.

PR #183, opened 2019-07-19
Secure element key registration

gilles-peskine-arm:
Add an API method that allows applications to register a key in a secure element.

Successor of #202. The first new commit is "psa_start_key_creation: take the method as a parameter".

This PR has undergone significant rework. See https://github.com/ARMmbed/mbed-crypto/pull/183#issuecomment-515908690 . Previous iterations are available as [1](https://github.com/gilles-peskine-arm/mbed-crypto/tree/psa-se_driver-register_key-1), [2](https://github.com/gilles-peskine-arm/mbed-crypto/tree/psa-se_driver-register_key-2), [3](https://github.com/gilles-peskine-arm/mbed-crypto/tree/psa-se_driver-register_key-3), [4](https://github.com/gilles-peskine-arm/mbed-crypto/tree/psa-se_driver-register_key-4).

### DIFF
--- a/include/psa/crypto_extra.h
+++ b/include/psa/crypto_extra.h
@@ -175,6 +175,44 @@ static inline void psa_clear_key_slot_number(
     attributes->core.flags &= ~MBEDTLS_PSA_KA_FLAG_HAS_SLOT_NUMBER;
 }
 
+/** Register a key that is already present in a secure element.
+ *
+ * The key must be located in a secure element designated by the
+ * lifetime field in \p attributes, in the slot set with
+ * psa_set_key_slot_number() in the attribute structure.
+ * This function makes the key available through the key identifier
+ * specified in \p attributes.
+ *
+ * \param[in] attributes        The attributes of the existing key.
+ *
+ * \retval #PSA_SUCCESS
+ *         The key was successfully registered.
+ *         Note that depending on the design of the driver, this may or may
+ *         not guarantee that a key actually exists in the designated slot
+ *         and is compatible with the specified attributes.
+ * \retval #PSA_ERROR_ALREADY_EXISTS
+ *         There is already a key with the identifier specified in
+ *         \p attributes.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         \p attributes specifies a lifetime which is not located
+ *         in a secure element.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         No slot number is specified in \p attributes,
+ *         or the specified slot number is not valid.
+ * \retval #PSA_ERROR_NOT_PERMITTED
+ *         The caller is not authorized to register the specified key slot.
+ * \retval #PSA_ERROR_INSUFFICIENT_MEMORY
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_HARDWARE_FAILURE
+ * \retval #PSA_ERROR_CORRUPTION_DETECTED
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+psa_status_t mbedtls_psa_register_se_key(
+    const psa_key_attributes_t *attributes);
+
 #endif /* MBEDTLS_PSA_CRYPTO_SE_C */
 
 /**@}*/
--- a/include/psa/crypto_se_driver.h
+++ b/include/psa/crypto_se_driver.h
@@ -818,7 +818,27 @@ typedef enum
     PSA_KEY_CREATION_GENERATE, /**< During psa_generate_key() */
     PSA_KEY_CREATION_DERIVE, /**< During psa_key_derivation_output_key() */
     PSA_KEY_CREATION_COPY, /**< During psa_copy_key() */
-    PSA_KEY_CREATION_REGISTER, /*TEMPORARY*/
+
+#ifndef __DOXYGEN_ONLY__
+    /** A key is being registered with mbedtls_psa_register_se_key().
+     *
+     * The core only passes this value to
+     * psa_drv_se_key_management_t::p_validate_slot_number, not to
+     * psa_drv_se_key_management_t::p_allocate. The call to
+     * `p_validate_slot_number` is not followed by any other call to the
+     * driver: the key is considered successfully registered if the call to
+     * `p_validate_slot_number` succeeds, or if `p_validate_slot_number` is
+     * null.
+     *
+     * With this creation method, the driver must return #PSA_SUCCESS if
+     * the given attributes are compatible with the existing key in the slot,
+     * and #PSA_ERROR_DOES_NOT_EXIST if the driver can determine that there
+     * is no key with the specified slot number.
+     *
+     * This is an Mbed Crypto extension.
+     */
+    PSA_KEY_CREATION_REGISTER,
+#endif
 } psa_key_creation_method_t;
 
 /** \brief A function that allocates a slot for a key.
--- a/include/psa/crypto_se_driver.h
+++ b/include/psa/crypto_se_driver.h
@@ -818,6 +818,7 @@ typedef enum
     PSA_KEY_CREATION_GENERATE, /**< During psa_generate_key() */
     PSA_KEY_CREATION_DERIVE, /**< During psa_key_derivation_output_key() */
     PSA_KEY_CREATION_COPY, /**< During psa_copy_key() */
+    PSA_KEY_CREATION_REGISTER, /*TEMPORARY*/
 } psa_key_creation_method_t;
 
 /** \brief A function that allocates a slot for a key.
--- a/include/psa/crypto_se_driver.h
+++ b/include/psa/crypto_se_driver.h
@@ -810,25 +810,45 @@ typedef struct {
  */
 /**@{*/
 
+/** An enumeration indicating how a key is created.
+ */
+typedef enum
+{
+    PSA_KEY_CREATION_IMPORT, /**< During psa_import_key() */
+    PSA_KEY_CREATION_GENERATE, /**< During psa_generate_key() */
+    PSA_KEY_CREATION_DERIVE, /**< During psa_key_derivation_output_key() */
+    PSA_KEY_CREATION_COPY, /**< During psa_copy_key() */
+} psa_key_creation_method_t;
+
 /** \brief A function that allocates a slot for a key.
  *
  * To create a key in a specific slot in a secure element, the core
  * first calls this function to determine a valid slot number,
  * then calls a function to create the key material in that slot.
- * For example, in nominal conditions (that is, if no error occurs),
- * the effect of a call to psa_import_key() with a lifetime that places
- * the key in a secure element is the following:
+ * In nominal conditions (that is, if no error occurs),
+ * the effect of a call to a key creation function in the PSA Cryptography
+ * API with a lifetime that places the key in a secure element is the
+ * following:
  * -# The core calls psa_drv_se_key_management_t::p_allocate
  *    (or in some implementations
  *    psa_drv_se_key_management_t::p_validate_slot_number). The driver
  *    selects (or validates) a suitable slot number given the key attributes
  *    and the state of the secure element.
- * -# The core calls psa_drv_se_key_management_t::p_import to import
- *    the key material in the selected slot.
+ * -# The core calls a key creation function in the driver.
  *
- * Other key creation methods lead to similar sequences. For example, the
- * sequence for psa_generate_key() is the same except that the second step
- * is a call to psa_drv_se_key_management_t::p_generate.
+ * The key creation functions in the PSA Cryptography API are:
+ * - psa_import_key(), which causes
+ *   a call to `p_allocate` with \p method = #PSA_KEY_CREATION_IMPORT
+ *   then a call to psa_drv_se_key_management_t::p_import.
+ * - psa_generate_key(), which causes
+ *   a call to `p_allocate` with \p method = #PSA_KEY_CREATION_GENERATE
+ *   then a call to psa_drv_se_key_management_t::p_import.
+ * - psa_key_derivation_output_key(), which causes
+ *   a call to `p_allocate` with \p method = #PSA_KEY_CREATION_DERIVE
+ *   then a call to psa_drv_se_key_derivation_t::p_derive.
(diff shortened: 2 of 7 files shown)


include/psa/crypto_extra.h
```diff
@@ -154,6 +154,60 @@ typedef struct mbedtls_psa_stats_s
  */
 void mbedtls_psa_get_stats( mbedtls_psa_stats_t *stats );
```
AndrzejKurek: Typo in commit c3d4baa - "pplications"

include/psa/crypto_extra.h
```diff
@@ -206,6 +206,44 @@ static inline psa_key_slot_number_t psa_get_key_slot_number(
...
+ * lifetime field in \p attributes, in the slot set with
+ * psa_set_key_slot_number() in the attribute structure.
+ * This function makes the key available through the key identifier
+ * specified in \p attributes.
+ *
+ * \param[in] attributes        The attributes of the existing key.
+ *
+ * \retval #PSA_SUCCESS
+ *         The key was successfully registered.
+ *         Note that depending on the driver's
```
AndrzejKurek: Unfinished sentence.

include/psa/crypto_extra.h
```diff
@@ -206,6 +206,44 @@ static inline psa_key_slot_number_t psa_get_key_slot_number(
...
+ * \retval #PSA_ERROR_ALREADY_EXISTS
+ *         There is already a key with thw identifier specified in
```
AndrzejKurek: Typo - `thw` -> `the`

include/psa/crypto_extra.h
```diff
@@ -206,6 +206,44 @@ static inline psa_key_slot_number_t psa_get_key_slot_number(
...
+ *         \p attributes.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         \p attributes specifies a lifetime which is not located
+ *         in a secure eleement.
```
AndrzejKurek: Typo - `eleement` -> `element`

include/psa/crypto_se_driver.h
```diff
@@ -814,6 +814,30 @@ typedef psa_status_t (*psa_drv_se_allocate_key_t)(
...
 
+/** \brief A function that determines whether a slot number is valid
+ * for a key.
+ *
+ * \param[in,out] drv_context       The driver context structure.
+ * \param[in] attributes    Attributes of the key.
+ * \param[in] key_slot      Slot where the key is to be stored.
+ *
+ * \retval #PSA_SUCCESS
+ *         The given slot number is valid.for a key with the given
```
AndrzejKurek: Typo - `.` -> ` `

library/psa_crypto.c
```diff
@@ -1386,10 +1414,17 @@ static psa_status_t psa_start_key_creation(
...
-                                           &slot->data.se.slot_number );
-        if( status != PSA_SUCCESS )
-            return( status );
+        if( method != PSA_KEY_CREATION_REGISTER )
+        {
+            status = psa_find_se_slot_for_key( attributes, *p_drv,
+                                               &slot->data.se.slot_number );
+            if( status != PSA_SUCCESS )
+                return( status );
+        }
```
AndrzejKurek: Maybe a call to `p_check_slot` could be made here if the method is `PSA_KEY_CREATION_REGISTER`? Just to check that what the application is trying to register is sane.

gilles-peskine-arm: Yes, but not as is: `p_check_slot` might say “no, you can't create a key here because it's occupied”. I lean towards adding a `psa_key_creation_method_t` to `p_allocate` and `p_check_slot`, which would allow the driver to enforce constraints like “this slot is locked against import and can only contain internally-generated, non-exportable keys”.

gilles-peskine-arm:
> What part of the PSA depends on the key type being PSA_KEY_TYPE_RAW_DATA to work properly with secure elements, if any?

I'm not sure I understand the question. A secure element doesn't have to support `PSA_KEY_TYPE_RAW_DATA`. I expect that most do: it's of very limited usefulness since the SE can't use that data, but it helps deployment to use a secure element slot to store a certificate for a private key that's in the secure element, and that slot would have the type `PSA_KEY_TYPE_RAW_DATA`.

include/psa/crypto_extra.h
```diff
@@ -154,6 +154,98 @@ typedef struct mbedtls_psa_stats_s
...
+ */
+static inline void psa_set_key_slot_number(
+    psa_key_attributes_t *attributes,
+    psa_key_slot_number_t slot_number)
+{
+    attributes->slot_number = slot_number;
+    attributes->has_slot_number = 1;
+}
+
+/** Retrieve the enrollment algorithm policy from key attributes.
```
athoelke: Cut and paste bug - description copied from `psa_get_key_policy`

include/psa/crypto_extra.h
```diff
@@ -154,6 +154,98 @@ typedef struct mbedtls_psa_stats_s
...
+ *
+ * \param[in] attributes        The key attribute structure to query.
+ *
+ * \return The slot containing the specified key.
+ *         If the key is not in a secure element,
+ *         or if the caller is not permitted to observe slot numbers,
+ *         the value is unspecified.
```
athoelke: Slightly 'euww!'

Can we define a symbolic name for `PSA_KEY_SLOT_NUMBER_INVALID` (or similar), and have the implementation define it as a constant value of type `psa_key_slot_number_t`? - this API will return that value if the key slot number is not set (or doesn't exist in this implementation).

That way we get implementation efficiency, precision in the spec, and still permit implementations to define the type of `psa_key_slot_number_t` and use `0` as a valid slot number.

include/psa/crypto_extra.h
```diff
@@ -154,6 +154,98 @@ typedef struct mbedtls_psa_stats_s
...
+ */
+static inline psa_key_slot_number_t psa_get_key_slot_number(
+    const psa_key_attributes_t *attributes)
+{
+    return( attributes->slot_number );
```
athoelke: Will compiler-error if `MBEDTLS_PSA_CRYPTO_SE_C` is not defined. Yes I know you say "undefined" but this is a painful way to inform developers that they misunderstand the API and how to fix their mistake.

include/psa/crypto_extra.h
```diff
@@ -154,6 +154,98 @@ typedef struct mbedtls_psa_stats_s
...
+ * \retval #PSA_SUCCESS
+ *         The key was successfully registered.
+ *         Note that depending on the driver's
+ * \retval #PSA_ERROR_ALREADY_EXISTS
+ *         There is already a key with thw identifier specified in
+ *         \p attributes.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         \p attributes specifies a lifetime which is not located
+ *         in a secure eleement.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
```
athoelke: Duplicate error code - this might be confusing in the API documentation? (or is this pattern used elsewhere)

include/psa/crypto_se_driver.h
```diff
@@ -114,10 +114,14 @@ typedef psa_status_t (*psa_drv_se_init_t)(psa_drv_se_context_t *drv_context,
                                           void *persistent_data,
                                           psa_key_lifetime_t lifetime);
 
+#ifdef __DOXYGEN_ONLY__
+/* Mbed Crypto defines this type in crypto_extra.h because it is also
+ * available to applications through implementation-specific extensions. */
```
athoelke: Is it worth a comment to indicate that this type is also defined in `crypto_types.h` - and the definitions need to be identical?

library/psa_crypto.c
```diff
@@ -1093,6 +1101,13 @@ psa_status_t psa_get_key_attributes( psa_key_handle_t handle,
     attributes->type = slot->type;
     attributes->bits = psa_get_key_slot_bits( slot );
 
+#if defined(MBEDTLS_PSA_CRYPTO_SE_C)
+    if( psa_get_se_driver( slot->lifetime, &drv, NULL ) )
+    {
+        attributes->slot_number = slot->data.se.slot_number;
```
athoelke: Should this also set `has_slot_number`?

include/psa/crypto_extra.h
gilles-peskine-arm: Right now, all 2^64 potential values for a slot number are considered valid. I'm not sure this is the best design. I'm sure every driver can afford one invalid value. For consistency with the rest of PSA crypto, 0 should be invalid. Making 0 invalid would also avoid the annoying extra `has_slot_number` bit. But typical secure elements number their slots from 0 to N-1, so it would require most drivers to do some shifting. So declare all-bits-one invalid (with a symbolic constant of course) and mandate implementations to return it when not applicable?

include/psa/crypto_extra.h
gilles-peskine-arm: I'll fix such errors in a batch when I add a build with `MBEDTLS_PSA_CRYPTO_SE_C` not defined.

include/psa/crypto_extra.h
gilles-peskine-arm: This pattern is used elsewhere: each `\retval` is an error reason, and some error reasons can map to the same error code. The alternative is something like
```
 * \retval #PSA_ERROR_INVALID_ARGUMENT
 *         \p attributes specifies a lifetime which is not located
 *         in a secure eleement.
 *
 *         Alternatively, …
```
I have no strong feelings either way, but I'd only change this if we change it elsewhere.

include/psa/crypto_se_driver.h
gilles-peskine-arm: I'd prefer to create a new header file. I made this hack to avoid having to mess with build scripts.

library/psa_crypto.c
gilles-peskine-arm: Yes. Fixed.

gilles-peskine-arm:
Thanks for the reviews. I'll fix the documentation mistakes in a batch later, probably rebased against a new version of #157.

gilles-peskine-arm:
I've rebased this work on top of the latest version of https://github.com/ARMmbed/mbed-crypto/pull/157 . I also made significant changes to the history, to address review comments, to avoid breaking tests, and also to make some design changes.

@athoelke @AndrzejKurek As per our habit I marked the review comments that I've handled with :rocket:. I'm still planning to make an improvement pass, but perhaps only after yet another rebase on top of #157 once I've addressed the review comments there.

gilles-peskine-arm:
Now that #157 is merged, I've rebased on top of that and switched the target branch to [psa-api-1.0-beta](https://github.com/ARMmbed/mbed-crypto/tree/psa-api-1.0-beta). I'm not planning any more rebases.

gilles-peskine-arm:
On second thoughts, I am planning to rebase and even split this PR. It does three things which would be better done separately and in the right sequence:

1. Store the bit size of a key in a secure element, and make sure that `psa_get_key_attributes` works for an SE key after a restart.
2. Support the slot number in key attributes.
3. Key registration.

include/psa/crypto_se_driver.h
gilles-peskine-arm: For the time being, I'm not adding a header file. I think I'll just wait until the header files are no longer used to generate the documentation so this becomes moot.

include/psa/crypto_extra.h
gilles-peskine-arm: I've changed `psa_get_key_slot_number` to return an error if there is no recorded slot number. See #201.

library/psa_crypto.c
gilles-peskine-arm: Done in the newest rebase. On a related note, I renamed `p_check_slot` to `p_validate_slot_number` in #202.

library/psa_crypto.c
```diff
@@ -1505,6 +1505,16 @@ static psa_status_t psa_validate_key_attributes(
     return( PSA_SUCCESS );
 }
 
+/** An enumeration indicating how a key is created.
```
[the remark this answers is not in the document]
gilles-peskine-arm: I'll fix that when I rebase (which will probably happen once the precursor PRs are reviewed).

include/psa/crypto_se_driver.h
```diff
@@ -831,14 +842,16 @@ typedef psa_status_t (*psa_drv_se_allocate_key_t)(
...
 
 /** \brief A function that determines whether a slot number is valid
  * for a key.
  *
- * \param[in,out] drv_context       The driver context structure.
- * \param[in] attributes    Attributes of the key.
- * \param[in] key_slot      Slot where the key is to be stored.
+ * \param[in,out] drv_context   The driver context structure.
+ * \param[in] attributes        Attributes of the key.
+ * \param method                The way in which the key is being created.
```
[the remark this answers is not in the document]
gilles-peskine-arm: It isn't a pointer. The in/out annotations are about the thing pointed to. A data argument is always "in" and we don't annotate them.

library/psa_crypto.c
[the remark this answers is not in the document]
k-stachowiak: Please note, that the commit message typo is still present after the recent rebase.

include/psa/crypto_se_driver.h
```diff
@@ -803,12 +810,44 @@ typedef struct {
...
+ */
+typedef enum
+{
+    PSA_KEY_CREATION_IMPORT, /**< During psa_import_key() */
+    PSA_KEY_CREATION_GENERATE, /**< During psa_generate_key() */
+    PSA_KEY_CREATION_DERIVE, /**< During psa_key_derivation_output_key() */
+    PSA_KEY_CREATION_COPY, /**< During psa_copy_key() */
+
+#ifndef __DOXYGEN_ONLY__
+    /** A key is being registered with mbedtls_psa_register_se_key().
```
AndrzejKurek: Why is this in the doxygen ifndef block?

gilles-peskine-arm: Because it's an Mbed Crypto extension, so it shouldn't be included in the PSA specification. We normally resolve this by using different files, but this is an enum so it can't be defined partly in one file and partly in another.  `__DOXYGEN_ONLY__` is not defined when generating the Mbed Crypto documentation.

gilles-peskine-arm:
I rebased again to get rid of the commits from preceding PRs, to resolve a trivial conflict in `crypto_se_driver.h` due to the late changes of the documentation of `psa_drv_se_allocate_key_t` in #202, and also to reorder commits (I moved "Pass the key creation method to drivers" before "New function mbedtls_psa_register_se_key") to avoid having an unbuildable commit in the middle.

PR #185, opened 2019-07-19
Remove method dispatch from md

gilles-peskine-arm:
Remove method dispatch from the md module. Do dispatch with a switch based on type.

This removes the ability for users to define their own `md_info` structures, which was possible in practice but only by using interfaces described in `md_internal.h`, which were not meant to be transparent. The main benefit is smaller code, easier to optimize. There's also very slight security advantage in having fewer function pointers lying around for exploits to leverage.

This is an ABI change and an internal-only API change.

~Once this PR is merged, update the headers and the submodule in Mbed TLS via https://github.com/ARMmbed/mbedtls/pull/2757.~


gilles-peskine-arm:
Crypto CI ok except for the API/ABI change as expected. TLS CI in https://github.com/ARMmbed/mbedtls/pull/2757

gilles-peskine-arm:
> I gather that would be both ABI and API break

Right, I wanted to avoid an outright API break.

I'll relaunch the CI now that the header files no longer need to be updated in TLS.

Patater:
https://jenkins-internal.mbed.com/blue/organizations/jenkins/mbed-crypto-pr/detail/PR-185-merge/20/pipeline/2604

Only failure in merge CI job is ABI change and TLS testing finding some generated files to be stale. Probably the generated files issue needs to be fixed before we merge this.

gilles-peskine-arm:
@Patater Doesn't fixing the `check_generated_files.sh` failure require a change on the TLS side, namely, running `generate_visualc_files.pl` after updating the submodule? I don't see how it could be related to this PR.

Patater:
> @Patater Doesn't fixing the `check_generated_files.sh` failure require a change on the TLS side, namely, running `generate_visualc_files.pl` after updating the submodule? I don't see how it could be related to this PR.

Correct. Removal of md_wrap.c would require running the generate files scripts in TLS after submodule updating. We could probably improve our TLS testing a bit...

gilles-peskine-arm:
Oh right, it's the removal of `md_wrap.c`. Which is indeed due to this PR, but can't be fixed in this PR.

PR #186, opened 2019-07-21
Add CMake option for explicitly link library to trusted_storage

moshe-shahar:
CMake option name: LINK_WITH_TRUSTED_STORAGE
Default value is `ON` since the default of mbedtls for `MBEDTLS_PSA_ITS_FILE_C` [is using ITS](https://github.com/ARMmbed/mbedtls/blob/ff645d9838f4797c726562d90c0a9ae5207ea86f/include/mbedtls/config.h#L2872).

This PR is based on https://github.com/ARMmbed/mbed-crypto/pull/180 with the addition of one commit. Please merge after it.

The new option assumes `trusted_storage` library exist and force `mbedcrypto` library to link with it. Any application/test uses emulated ITS (by defining `MBEDTLS_PSA_ITS_FILE_C`) should set this option to `OFF`.

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -1,7 +1,7 @@
 option(USE_STATIC_MBEDTLS_LIBRARY "Build mbed TLS static library." ON)
 option(USE_SHARED_MBEDTLS_LIBRARY "Build mbed TLS shared library." OFF)
 option(LINK_WITH_PTHREAD "Explicitly link mbed TLS library to pthread." OFF)
-option(LINK_WITH_TRUSTED_STORAGE "Explicitly link mbed TLS library to trusted_storage." ON)
+option(LINK_WITH_TRUSTED_STORAGE "Explicitly link mbed TLS library to trusted_storage." OFF)
 
 # Set the project root directory if it's not already defined, as may happen if
 # the library folder is included directly by a parent project, without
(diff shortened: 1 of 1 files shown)


moshe-shahar:
@Patater, please review.

@simonqhughes, FYI.

library/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 option(USE_STATIC_MBEDTLS_LIBRARY "Build mbed TLS static library." ON)
 option(USE_SHARED_MBEDTLS_LIBRARY "Build mbed TLS shared library." OFF)
 option(LINK_WITH_PTHREAD "Explicitly link mbed TLS library to pthread." OFF)
+option(LINK_WITH_TRUSTED_STORAGE "Explicitly link mbed TLS library to trusted_storage." ON)
```
Patater: This should default to OFF, as most of our Linux systems don't have a trusted_storage library available. On Mbed Linux, this can be changed to ON.

library/CMakeLists.txt
```diff
@@ -125,6 +126,10 @@ if(LINK_WITH_PTHREAD)
     set(libs ${libs} pthread)
 endif()
 
+if(LINK_WITH_TRUSTED_STORAGE)
+    set(libs ${libs} trusted_storage)
+endif()
```
Patater: Do we need to link with trusted_storage in Makefile builds as well? Seems odd to only add the possibility to link with trusted_stored for CMake.

Patater: I've filed https://github.com/ARMmbed/mbed-crypto/issues/188 for this

library/CMakeLists.txt
moshe-shahar: done

PR #192, opened 2019-07-30
Add some negative tests for policy checks

gilles-peskine-arm:
The goal of this PR is to get the work from #106 into the product (in the PSA branch, since this is PSA work). It only contains one commit to add test cases. The rest has already gone in through ARMmbed/mbedtls-psa#266.


gilles-peskine-arm:
The PR-head job had an unexplained failure (missing logs). I re-ran the PR-merge job, which has the benefit of an up-to-date merge target, and it passed except for the ABI check which doesn't matter on the API branch.

PR #187, opened 2019-07-23
Bring in changes from Mbed TLS as of 2019-07-22

Patater:
Suggested ways of reviewing:
- Don't use GitHub
- Compare Mbed TLS `development` files to the files changed in this PR
- Compare previous version of Mbed Crypto files with the files changed in this PR
  - Review [this diff](https://github.com/ARMmbed/mbed-crypto/files/3430701/20190722-changes.txt) to easily see this

PR #198, opened 2019-07-31
Merge mbed-crypto/development into psa-api-1.0-beta

gilles-peskine-arm:
Take development just after #187 was merged. Update to the 1.0 final ITS interface, allow declarations after statements, and many other changes.

This is one merge commit, plus a follow-up.


gilles-peskine-arm:
Crypto CI ok except for the ABI checking script (known bug: it fails on merges) which we don't care about in this branch anyway.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -712,9 +712,9 @@ exit:
     return( ok );
 }
 
-static int asn1_skip_integer( unsigned char **p, const unsigned char *end,
```
k-stachowiak: This change popped up for me during reproduction of the merge. It doesn't seem to be used elsewhere, hence could be left static.

library/psa_crypto_se.c
```diff
@@ -148,17 +148,23 @@ psa_status_t psa_load_se_persistent_data(
 {
     psa_status_t status;
     psa_storage_uid_t uid;
+    size_t length;
 
     status = psa_get_se_driver_its_file_uid( driver, &uid );
     if( status != PSA_SUCCESS )
         return( status );
 
+    /* Read the amount of persistent data that the driver requests.
```
k-stachowiak: While the section below explains the code, this one rather describes the function behavior. I find it fine, but instead of placing two such comment sections next to each other, the first one could be placed as a doxygen note?

tests/suites/test_suite_psa_crypto.function
gilles-peskine-arm: It's non-static because it's unused in some build configurations, and the compiler complains that a function is unused. It didn't use to be a problem because we used to compile tests with `-Wno-unused`, but since this merge, we've stopped doing that. This is explained in the merge commit message for `test_suite_psa_crypto_slot_management.function`. I didn't list it for `test_suite_psa_crypto.function` because there it's a change that already happened in the development branch, there was no conflict.

library/psa_crypto_se.c
gilles-peskine-arm: What do you mean by “placed as a doxygen note”? This implements the behavior documented in `crypto_se_driver.h` for the `persistent_data` of `psa_drv_se_context_t`, which `driver->internal.persistent_data` accesses. I think it's useful to explain what's going on at the point of implementation.

k-stachowiak: I may have misunderstood the comment to be information that may be interesting to the function's user and as such placed in the doxygen comment for the function. I found it suspicious for a function to have the same amount of comment lines that it has code, so I suggested revising it, which you kindly have. Thanks.

PR #201, opened 2019-08-02
Add slot number attribute

gilles-peskine-arm:
Add a slot number attribute for the sake of keys in a secure element. Report the slot number when relevant.

The main point of having a slot number attribute is to allow applications to choose the slot number when creating a key, which is done in the successor PR #202.

~Successor of https://github.com/ARMmbed/mbed-crypto/pull/197. The first new commit is "Add infrastructure for key attribute flags".~

### DIFF
--- a/include/psa/crypto_extra.h
+++ b/include/psa/crypto_extra.h
@@ -104,6 +104,67 @@ static inline psa_algorithm_t psa_get_key_enrollment_algorithm(
     return( attributes->core.policy.alg2 );
 }
 
+#if defined(MBEDTLS_PSA_CRYPTO_SE_C)
+
+/** Retrieve the slot number where a key is stored.
+ *
+ * A slot number is only defined for keys that are stored in a secure
+ * element.
+ *
+ * This information is only useful if the secure element is not entirely
+ * managed through the PSA Cryptography API. It is up to the secure
+ * element driver to decide how PSA slot numbers map to any other interface
+ * that the secure element may have.
+ *
+ * \param[in] attributes        The key attribute structure to query.
+ * \param[out] slot_number      On success, the slot number containing the key.
+ *
+ * \retval #PSA_SUCCESS
+ *         The key is located in a secure element, and \p *slot_number
+ *         indicates the slot number that contains it.
+ * \retval #PSA_ERROR_NOT_PERMITTED
+ *         The caller is not permitted to query the slot number.
+ *         Mbed Crypto currently does not return this error.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         The key is not located in a secure element.
+ */
+psa_status_t psa_get_key_slot_number(
+    const psa_key_attributes_t *attributes,
+    psa_key_slot_number_t *slot_number );
+
+/** Choose the slot number where a key is stored.
+ *
+ * This function declares a slot number in the specified attribute
+ * structure.
+ *
+ * A slot number is only meaningful for keys that are stored in a secure
+ * element. It is up to the secure element driver to decide how PSA slot
+ * numbers map to any other interface that the secure element may have.
+ *
+ * \note Setting a slot number in key attributes for a key creation can
+ *       cause the following errors when creating the key:
+ *       - #PSA_ERROR_NOT_SUPPORTED if the selected secure element does
+ *         not support choosing a specific slot number.
+ *       - #PSA_ERROR_NOT_PERMITTED if the caller is not permitted to
+ *         choose slot numbers in general or to choose this specific slot.
+ *       - #PSA_ERROR_INVALID_ARGUMENT if the chosen slot number is not
+ *         valid in general or not valid for this specific key.
+ *       - #PSA_ERROR_ALREADY_EXISTS if there is already a key in the
+ *         selected slot.
+ *
+ * \param[out] attributes       The attribute structure to write to.
+ * \param slot_number           The slot number to set.
+ */
+static inline void psa_set_key_slot_number(
+    psa_key_attributes_t *attributes,
+    psa_key_slot_number_t slot_number )
+{
+    attributes->core.flags |= MBEDTLS_PSA_KA_FLAG_HAS_SLOT_NUMBER;
+    attributes->slot_number = slot_number;
+}
+
+#endif /* MBEDTLS_PSA_CRYPTO_SE_C */
+
 /**@}*/
 
 /**
--- a/include/psa/crypto_extra.h
+++ b/include/psa/crypto_extra.h
@@ -163,6 +163,18 @@ static inline void psa_set_key_slot_number(
     attributes->slot_number = slot_number;
 }
 
+/** Remove the slot number attribute from a key attribute structure.
+ *
+ * This function undoes the action of psa_set_key_slot_number().
+ *
+ * \param[out] attributes       The attribute structure to write to.
+ */
+static inline void psa_clear_key_slot_number(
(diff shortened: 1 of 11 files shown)


library/psa_crypto_core.h
```diff
@@ -68,7 +68,7 @@ typedef struct
 
 /* A mask of key attribute flags used only internally.
  * Currently there aren't any. */
-#define MBEDTLS_PSA_KA_MASK_INTERNAL_ONLY (     \
+#define PSA_KA_MASK_INTERNAL_ONLY (     \
```
AndrzejKurek: Is there any policy for PSA macros similar to this one with MbedTLS ones?

include/psa/crypto_extra.h
```diff
@@ -163,6 +163,18 @@ static inline void psa_set_key_slot_number(
...
+/** Remove the slot number attribute from a key attribute structure.
+ *
+ * This function undoes the action of psa_set_key_slot_number().
+ *
+ * \param[out] attributes       The attribute structure to write to.
+ */
+static inline void psa_clear_key_slot_number(
+    psa_key_attributes_t *attributes )
+{
+    attributes->core.flags &= ~MBEDTLS_PSA_KA_FLAG_HAS_SLOT_NUMBER;
```
AndrzejKurek: Wouldn't it be good to reset the slot number to an invalid number here?

library/psa_crypto_core.h
gilles-peskine-arm: No. Mbed TLS had no internal headers (`library/*.h`), but Mbed Crypto does in its PSA parts (`library/psa_*.h`).

include/psa/crypto_extra.h
gilles-peskine-arm: Yes, but there's no invalid slot number at the moment. I'm considering introducing some though, if only to be future-proof. A slot number is 64 bits, maybe we should say that a slot number is 62 bits, reserve values 0xc000000000000000..0xfffffffffffffffe and declare 0xffffffffffffffff to be permanently invalid? This would not remove the need for the HAS_SLOT_NUMBER flag, because 0 is valid.

include/psa/crypto_struct.h
```diff
@@ -331,11 +331,13 @@ typedef uint16_t psa_key_bits_t;
...
 typedef uint16_t psa_key_attributes_flag_t;
 
-#define MBEDLTS_PSA_KA_FLAG_SLOT_NUMBER ( (psa_key_attributes_flag_t) 0x0001 )
+#define MBEDTLS_PSA_KA_FLAG_HAS_SLOT_NUMBER     \
+    ( (psa_key_attributes_flag_t) 0x0001 )
 
 /* A mask of key attribute flags used externally only.
  * Only meant for internal checks inside the library. */
 #define MBEDTLS_PSA_KA_MASK_EXTERNAL_ONLY (      \
+        MBEDTLS_PSA_KA_FLAG_HAS_SLOT_NUMBER |    \
```
AndrzejKurek: Why or it with zero?

gilles-peskine-arm: So we can add/remove flags just by editing their line and not have to modify the surrounding punctuation.

library/psa_crypto.c
```diff
@@ -1407,6 +1407,15 @@ psa_status_t psa_export_public_key( psa_key_handle_t handle,
                                      data_length, 1 ) );
 }
 
+#if defined(static_assert)
+static_assert( ( MBEDTLS_PSA_KA_MASK_EXTERNAL_ONLY & MBEDTLS_PSA_KA_MASK_DUAL_USE ) == 0,
+               "One or more key attribute flag is listed as both external-only and dual-use" );
+static_assert( ( MBEDTLS_PSA_KA_MASK_INTERNAL_ONLY & MBEDTLS_PSA_KA_MASK_DUAL_USE ) == 0,
+               "One or more key attribute flag is listed as both external-only and dual-use" );
```
AndrzejKurek: typo: `external-only` -> `internal-only`

library/psa_crypto.c
```diff
@@ -714,20 +714,24 @@ static inline size_t psa_get_key_slot_bits( const psa_key_slot_t *slot )
  *
  * \return The key size in bits, calculated from the key data.
  */
-static size_t psa_calculate_key_bits( const psa_key_slot_t *slot )
+static psa_key_bits_t psa_calculate_key_bits( const psa_key_slot_t *slot )
```
AndrzejKurek: Is it possible for a key to have `( bits % 8 ) != 0`?

gilles-peskine-arm: Yes, for some key types. Currently in Mbed Crypto this is only supported with two elliptic curves: Curve25519 (255) and SECP521R1 (521).

library/psa_crypto_core.h
```diff
@@ -64,9 +64,19 @@ typedef struct
     } data;
 } psa_key_slot_t;
 
-/** Flag for psa_key_slot_t::attr::core::flags indicating that the
- * slot is in use. */
-#define PSA_KEY_SLOT_FLAG_ALLOCATED ( (uint16_t) 0x0001 )
+/** Test whether a key slot is occupied.
+ *
+ * A key slot is occupied iff the key type is nonzero. This works because
```
AndrzejKurek: `iff` -> `if`

gilles-peskine-arm: No. Iff = “if and only if”

AndrzejKurek: Oh, right - Chris has reminded me of this acronym - if and only if. I really don't have it memorized.

library/psa_crypto_slot_management.c
```diff
@@ -111,12 +111,8 @@ psa_status_t psa_internal_allocate_key_slot( psa_key_handle_t *handle,
     for( *handle = PSA_KEY_SLOT_COUNT; *handle != 0; --( *handle ) )
     {
         *p_slot = &global_data.key_slots[*handle - 1];
-        if( ! psa_key_slot_get_flags( *p_slot, PSA_KEY_SLOT_FLAG_ALLOCATED ) )
-        {
-            psa_key_slot_set_bits_in_flags( *p_slot,
-                                            PSA_KEY_SLOT_FLAG_ALLOCATED );
+        if( ! psa_is_key_slot_occupied( *p_slot ) )
```
AndrzejKurek: This approach has one flaw: `psa_internal_allocate_key_slot` can be called twice and succeed.

gilles-peskine-arm: Yes, but does it matter? If the caller hasn't done anything with the slot yet, the slot is still free. Callers don't “release” the slot if they decide not to write anything after all. I do see how this could be error-prone though. How about renaming the function to `psa_get_empty_key_slot` (which used to exist)? I think that would make the new behavior of the function intuitive.

AndrzejKurek: Yes, this would make sense. Although this function is used internally, after seeing just its name, imagine the following situation: 
I want to allocate a batch (let's say, 3) keys and then set the same attributes / perform operations with all of them - I would receive the same key, three times.

gilles-peskine-arm:
I've rebased on top of the target branch now that #197 is merged. The rebase shows a large diff because other things were also merged, including `development` There were no conflicts. The previous version is in https://github.com/gilles-peskine-arm/mbed-crypto/tree/psa-se_driver-set_key_slot_number-1

gilles-peskine-arm:
Crypto CI ok except for API/ABI check.

PR #202, opened 2019-08-05
Let applications create a key in a specific secure element slot

gilles-peskine-arm:
Successor of #201. The first new commit is "Improve documentation of the allocate method".

### DIFF
--- a/include/psa/crypto_se_driver.h
+++ b/include/psa/crypto_se_driver.h
@@ -833,6 +833,30 @@ typedef psa_status_t (*psa_drv_se_allocate_key_t)(
     const psa_key_attributes_t *attributes,
     psa_key_slot_number_t *key_slot);
 
+/** \brief A function that determines whether a slot number is valid
+ * for a key.
+ *
+ * \param[in,out] drv_context       The driver context structure.
+ * \param[in] attributes    Attributes of the key.
+ * \param[in] key_slot      Slot where the key is to be stored.
+ *
+ * \retval #PSA_SUCCESS
+ *         The given slot number is valid for a key with the given
+ *         attributes.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         The given slot number is not valid for a key with the
+ *         given attributes. This includes the case where the slot
+ *         number is not valid at all.
+ * \retval #PSA_ERROR_ALREADY_EXISTS
+ *         There is already a key with the specified slot number.
+ *         Drivers may choose to return this error from the key
+ *         creation function instead.
+ */
+typedef psa_status_t (*psa_drv_se_validate_slot_number_t)(
+    psa_drv_se_context_t *drv_context,
+    const psa_key_attributes_t *attributes,
+    psa_key_slot_number_t key_slot);
+
 /** \brief A function that imports a key into a secure element in binary format
  *
  * This function can support any output from psa_export_key(). Refer to the
@@ -993,6 +1017,16 @@ typedef struct {
      * last call to this function had not taken place.
      */
     psa_drv_se_allocate_key_t   p_allocate;
+    /** Function that checks the validity of a slot for a key.
+     *
+     * The core calls this function instead of
+     * psa_drv_se_key_management_t::p_allocate to create
+     * a key in a specific slot. It then calls the actual creation function
+     * (such as psa_drv_se_key_management_t::p_import or
+     * psa_drv_se_key_management_t::p_generate) or
+     * psa_drv_se_key_management_t::p_destroy.
+     */
+    psa_drv_se_validate_slot_number_t p_validate_slot_number;
     /** Function that performs a key import operation */
     psa_drv_se_import_key_t     p_import;
     /** Function that performs a generation */
--- a/include/psa/crypto_se_driver.h
+++ b/include/psa/crypto_se_driver.h
@@ -134,10 +134,17 @@ typedef psa_status_t (*psa_drv_se_init_t)(psa_drv_se_context_t *drv_context,
                                           void *persistent_data,
                                           psa_key_lifetime_t lifetime);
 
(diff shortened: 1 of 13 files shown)


tests/suites/test_suite_psa_crypto_se_driver_hal.function
```diff
@@ -536,6 +548,64 @@ exit:
...
+    psa_key_handle_t handle = 0;
+    psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+    const uint8_t key_material[3] = {0xfa, 0xca, 0xde};
+
+    memset( &driver, 0, sizeof( driver ) );
+    memset( &key_management, 0, sizeof( key_management ) );
+    driver.hal_version = PSA_DRV_SE_HAL_VERSION;
+    driver.key_management = &key_management;
+    driver.persistent_data_size = sizeof( ram_slot_usage_t );
+    key_management.p_validate_slot_number = ram_validate_slot_number;
```
k-stachowiak: Since the allocation function remains null here, maybe the test could be extended with an attempt to use that (expected to fail), before the wanted slot attribute is set. Similarly, the test, that only sets the allocation function could perform a negative test with a wanted slot set.

tests/suites/test_suite_psa_crypto_se_driver_hal.function
```diff
@@ -587,17 +588,26 @@ void key_creation_in_chosen_slot( int slot_arg,
                              &handle );
     TEST_EQUAL( status, expected_status );
 
-    if( status == PSA_SUCCESS )
+    if( status != PSA_SUCCESS )
+        goto exit;
+
+    /* Maybe restart, to check that the information is saved correctly. */
+    if( restart )
```
k-stachowiak: I don't know if it matters, but if this passes, we won't know, if the key was fine before the restart.

gilles-peskine-arm: The test data generally calls this function first with `restart=0`, then again with the same parameters except for `restart=1`. So the first call checks that everything works in memory and the second call checks that the storage part works.

tests/suites/test_suite_psa_crypto_se_driver_hal.function
gilles-peskine-arm: Indeed, but that's a bit out of scope: in this series of PRs, a lot of functionality isn't tested.

include/psa/crypto_se_driver.h
```diff
@@ -977,7 +977,21 @@ typedef psa_status_t (*psa_drv_se_generate_key_t)(psa_drv_se_context_t *drv_cont
...
+     * psa_drv_se_key_management_t::p_import or
+     * psa_drv_se_key_management_t::p_generate).
+     *
+     * If this function succeeds, the next call that the core makes to the
+     * driver is either the creation function or
+     * psa_drv_se_key_management_t::p_destroy. Note that
+     * if the platform is reset after this function returns, the core
+     * may either subsequently call
+     * psa_drv_se_key_management_t::p_destroy or may behave as if the
+     * last call to this function had not taken place.
```
AndrzejKurek: What do you mean by "behave as if the last call to this function had not taken place"? Should the driver store persistent data about allocation separately? Should it be able to work in the following situation:
- `p_allocate` called, slot 1 returned;
- hardware resets;
- `p_generate` called with slot 1?

include/psa/crypto_se_driver.h
```diff
@@ -977,7 +977,21 @@ typedef psa_status_t (*psa_drv_se_generate_key_t)(psa_drv_se_context_t *drv_cont
  * If one of the functions is not implemented, it should be set to NULL.
  */
 typedef struct {
-    /** Function that allocates a slot. */
+    /** Function that allocates a slot for a key.
+     *
+     * The core calls this function to determine a slot number, then
```
AndrzejKurek: Why is this documentation here, not next to `psa_drv_se_allocate_key_t`?

gilles-peskine-arm: I'm not sure how to structure the documentation between the type and the function pointer. I think it makes more sense to document in the documentation of the type, which is where the parameters and return type are spelled out. But I don't think it makes sense to document the context in which the function is called there, and especially not information like “this function is called after that function” — from a type, it doesn't really make sense to reference a specific function rather than another type.

I'm open to suggestions to structure the documentation better.

include/psa/crypto_se_driver.h
gilles-peskine-arm: I'm actually not sure about that case. I'm going to try to come up with clearer explanations, but first I need to figure out in more detail which behaviors are possible.

I'd like to merge this PR without necessarily getting all the corner cases right (and there's a precedent: this function has been there for several weeks with incomplete documentation). But I'm planning to make at least one update here.

gilles-peskine-arm:
I've rebased on top of the target branch now that #197 is merged. The rebase shows a large diff because other things were also merged, including `development` There were no conflicts except for "Rename internal macro to pass check-names.sh" which had been done both in #201 and here and which I therefore skipped in the rebase. The previous version is in https://github.com/gilles-peskine-arm/mbed-crypto/tree/psa-se_driver-choose_key_slot_number-2.

include/psa/crypto_se_driver.h
AndrzejKurek: I would propose leaving just short notes here, like `Function that allocates a slot for a key.`, and provide all the documentation next to `psa_drv_se_allocate_key_t`. With the current structure, if someone wants to use this function, he/she has to read about it in both of these places, and that's confusing.

include/psa/crypto_se_driver.h
gilles-peskine-arm: I've reworked the documentation. I don't think it covers all the subtleties that can arise if an error or a reset happens, but I think it's good enough for now.

PR #204, opened 2019-08-06
Bring in changes from Mbed TLS as of 2019-08-06

gilles-peskine-arm:
Merge mbedtls development branch into mbed-crypto. Main changes:

* Fix for `abi_check.py` always failing on CI.
* `TEST_ASSUME` facility to skip unit tests at runtime.

Reviewers: all you need to review is the merge commit.


gilles-peskine-arm:
The Travis Clang job failed, but due to a system error that's unrelated to this problem. The Travis GCC job succeeded, and everything that's tested on Travis is also tested on Jenkins anyway (perhaps not with the exact same set of build settings, but close enough). So I'm merging this PR.

PR #210, opened 2019-08-07
Remove extra mbedtls_ecp_group_free()

krizhanovsky:
The call `mbedtls_ecp_group_free()` is unneccessary since the `grp` is already freed at the top of the function.

### DIFF
--- a/library/ecp_curves.c
+++ b/library/ecp_curves.c
@@ -836,7 +836,6 @@ int mbedtls_ecp_group_load( mbedtls_ecp_group *grp, mbedtls_ecp_group_id id )
 #endif /* MBEDTLS_ECP_DP_CURVE448_ENABLED */
 
         default:
-            mbedtls_ecp_group_free( grp );
             return( MBEDTLS_ERR_ECP_FEATURE_UNAVAILABLE );
     }
 }
--- a/library/ecp_curves.c
+++ b/library/ecp_curves.c
@@ -836,7 +836,7 @@ int mbedtls_ecp_group_load( mbedtls_ecp_group *grp, mbedtls_ecp_group_id id )
 #endif /* MBEDTLS_ECP_DP_CURVE448_ENABLED */
 
         default:
-            grp->id = 0;
+            grp->id = MBEDTLS_ECP_DP_NONE;
             return( MBEDTLS_ERR_ECP_FEATURE_UNAVAILABLE );
     }
 }
--- a/library/ecp_curves.c
+++ b/library/ecp_curves.c
@@ -836,6 +836,7 @@ int mbedtls_ecp_group_load( mbedtls_ecp_group *grp, mbedtls_ecp_group_id id )
 #endif /* MBEDTLS_ECP_DP_CURVE448_ENABLED */
 
         default:
+            grp->id = id;
             return( MBEDTLS_ERR_ECP_FEATURE_UNAVAILABLE );
     }
 }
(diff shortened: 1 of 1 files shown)


library/ecp_curves.c
```diff
@@ -836,7 +836,6 @@ int mbedtls_ecp_group_load( mbedtls_ecp_group *grp, mbedtls_ecp_group_id id )
 #endif /* MBEDTLS_ECP_DP_CURVE448_ENABLED */
 
         default:
-            mbedtls_ecp_group_free( grp );
```
gilles-peskine-arm: This leaves `grp->id` set to the invalid value `id`. Code that's using this function should stop using `grp` if the function returns a failure status, but it's more robust if `grp` stays in a valid state, with `id=0`.

krizhanovsky: @gilles-peskine-arm thank you for the review! Fixed.

library/ecp_curves.c
```diff
@@ -836,7 +836,7 @@ int mbedtls_ecp_group_load( mbedtls_ecp_group *grp, mbedtls_ecp_group_id id )
...
+            grp->id = id;
```
Patater: I don't understand why we need to set grp->id to id in the default case. Isn't that already its value?

krizhanovsky: Hi @Patater ,

this is just zeroing, instead of full call of `mbedtls_ecp_group_free()` which does much more logic and calls slow and unnecessary `mbedtls_platform_zeroize()`. As @gilles-peskine-arm  note the `id` might be changed above in `grp->id = id;`, so the zeroing is required.

This particular call isn't a big problem, but we observed significant performance problems due to extra memory allocations, zeroing and copying https://github.com/tempesta-tech/tempesta/issues/1064#issuecomment-519182986 . We're still in progress of optimizations and will be able to backport the changes later.

This single-liner is just a small cleanup which can be easitly done.

gilles-peskine-arm: This shouldn't be `grp->id = id`, it should be `grp->id = 0`. The point is that in the error case, `grp` should to a blank state. (Remaining in a blank state is not practical since `grp->id` must be set to `id` in all other cases.)

But I'm not convinced that it's better to use `grp->id = 0` than `mbedtls_ecp_group_free`. Sure, calling this function takes a few more cycles, but it also makes the code harder to maintain.

gilles-peskine-arm: And regarding performance in general, Mbed TLS does call malloc more often than it should. But the main culprit is the bignum module. No allocations are involved here.

krizhanovsky: Guys, I'm very sorry for the misprint - it's a big shame to make the single-line change only with 3 tries.

In our profiling we also observed memory zeroing among the most hot spots, but if you feel that the change is too minor improvement and doesn't make much sense, then feel free just to close the pull request.

Patater:
IAR compiler is not OK with this change, unfortunately.

```
[ 33%] Building C object library/CMakeFiles/mbedcrypto.dir/ecp_curves.c.obj

   IAR ANSI C/C++ Compiler V8.20.1.14183/W32 for ARM
   Copyright 1999-2017 IAR Systems AB.
   Network license: 10.10.12.86 (STD)

              grp->id = 0;
                      ^
"C:\builds\ws\-210-merge-FMBLXQN76JRI42LQEMOXU\src\library\ecp_curves.c",839  Error[Pe188]: 
          enumerated type mixed with another type
```

Patater:
To fix, set `id` to `MBEDTLS_ECP_DP_NONE` instead of to `0`.

krizhanovsky:
> To fix, set `id` to `MBEDTLS_ECP_DP_NONE` instead of to `0`.

Done.

Patater:
@krizhanovsky Thanks for working with us to get this in! Appreciated.

Issue #218, opened 2019-08-13
Implement transaction recovery for SE key creation/destruction


gilles-peskine-arm:
### Description

If a power failure (or more generally a reset of the cryptography subsystem) happens during the creation or destruction of a key, on the next start, a transaction file may be present in storage. Recovering the interrupted transaction (in `psa_crypto_recover_transaction`) is currently not implemented, which prevents `psa_crypto_init` from succeeding.

### Issue request type

    [ ] Question
    [ ] Enhancement
    [x] Bug

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTCRYPT-862*

gilles-peskine-arm:
This only concerns keys in a secure element that has its own key storage, not transparent keys or keys in a secure element whose keys are stored in wrapped form in the internal storage.

This issue only tracks dynamically registered secure element drivers enabled with `MBEDTLS_PSA_CRYPTO_SE_C`, which are deprecated. We are not going to implement transaction recovery for dynamic secure element drivers.

See https://github.com/Mbed-TLS/mbedtls/pull/7646 and follow-up issues for the same problem with the new kind of secure element driver (“unified driver model”).

gilles-peskine-arm:
Now tracked (for new-style drivers) at https://github.com/Mbed-TLS/mbedtls/issues/8431

Issue #217, opened 2019-08-13
Proper recovery if creating a key in a secure element fails


gilles-peskine-arm:
### Description

When creating a key in a secure element, if the call to the secure element's create method succeeds, but an error happens later (e.g. saving the driver's persistent data to storage), the API function reports a failure, but the key remains in the secure element. This is a resource leak and can be a data leak (when importing a key).

There may be other key creation errors that are not handled properly. This should be reviewed and tested.

### Issue request type

    [ ] Question
    [ ] Enhancement
    [x] Bug

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTCRYPT-863*

gilles-peskine-arm:
This issue only concerns dynamically registered secure element drivers enabled with `MBEDTLS_PSA_CRYPTO_SE_C`, which are deprecated. We do not intend to do any further work on that feature, even bug fixes, unless we have reports of it being used in the wild.

Issue #216, opened 2019-08-13
Implement custom RSA public exponent for keys in secure elements


gilles-peskine-arm:
### Description

In `psa_get_key_attributes`, reporting the domain parameters for a key in a secure element is not implemented. Specifically, we're supposed to report the public exponent of an RSA key, so that `psa_get_key_attributes` on a key followed by `psa_generate_key` creates a key with the same public exponent as the first key.

### Issue request type

    [ ] Question
    [x] Enhancement
    [ ] Bug

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTCRYPT-864*

gilles-peskine-arm:
The secure element driver interface enabled by `MBEDTLS_PSA_CRYPTO_SE_C` is deprecated. We will add new features to it.

The same issue applies to the new “unified” driver interface, but also to the PSA API, since we have removed the ability to randomly generate an RSA key with a custom public exponent. Since the driver interface now follows the API, if we add this possibility to the API, we will also add it to the driver interface.

Issue #215, opened 2019-08-13
Better error handling when destroying a key in a secure element


gilles-peskine-arm:
### Description

Consider `psa_destroy_key` on a key which is located in a secure element. The normal process is:

1. Create a transaction file.
2. Call the driver to destroy the key in the secure element.
3. Destroy the key metadata in internal storage.
4. Remove the transaction file.

If an error occurs, we normally do as much as we can anyway. But if creating the transaction file fails, we do nothing. This is undesirable because `psa_destroy_key` should make the best effort to destroy all key material. 

In particular, it's currently impossible to destroy a key if the storage is full, because creating the transaction file fails.

### Issue request type

    [ ] Question
    [x] Enhancement
    [ ] Bug

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTCRYPT-865*

gilles-peskine-arm:
The secure element driver interface enabled by `MBEDTLS_PSA_CRYPTO_SE_C` is deprecated. We will not work on it any further.

PR #221, opened 2019-08-13
SE keys: fix psa_destroy_key; add Github issue numbers for missing code

gilles-peskine-arm:
I went through FIXME/TODO comments in the code and added the corresponding issue numbers:
* https://github.com/ARMmbed/mbed-crypto/issues/214
* https://github.com/ARMmbed/mbed-crypto/issues/215
* https://github.com/ARMmbed/mbed-crypto/issues/216
* https://github.com/ARMmbed/mbed-crypto/issues/217
* https://github.com/ARMmbed/mbed-crypto/issues/218

I also modified a comment relating to https://github.com/ARMmbed/mbed-crypto/issues/86 (not related to secure elements) but did not add an issue number because this is not exactly a bug and may even be resolved differently.

While doing this, I found two bugs in `psa_destroy_key` which I fixed and added a test for (in one case, I actually fixed a test bug that was hidden by the implementation bug).

I'm making this PR to `psa-api-1.0-beta` to be on top of the latest SE work. Once this PR is merged, we should do another merge from `psa-api-1.0-beta` following #212.

### DIFF
--- a/include/psa/crypto_struct.h
+++ b/include/psa/crypto_struct.h
@@ -12,6 +12,26 @@
  * In implementations with isolation between the application and the
  * cryptography module, it is expected that the front-end and the back-end
  * would have different versions of this file.
+ *
+ * <h3>Design notes about multipart operation structures</h3>
+ *
+ * Each multipart operation structure contains a `psa_algorithm_t alg`
+ * field which indicates which specific algorithm the structure is for.
+ * When the structure is not in use, `alg` is 0. Most of the structure
+ * consists of a union which is discriminated by `alg`.
+ *
+ * Note that when `alg` is 0, the content of other fields is undefined.
+ * In particular, it is not guaranteed that a freshly-initialized structure
+ * is all-zero: we initialize structures to something like `{0, 0}`, which
+ * is only guaranteed to initializes the first member of the union;
+ * GCC and Clang initialize the whole structure to 0 (at the time of writing),
+ * but MSVC and CompCert don't.
+ *
+ * In Mbed Crypto, multipart operation structures live independently from
+ * the key. This allows Mbed Crypto to free the key objects when destroying
+ * a key slot. If a multipart operation needs to remember the key after
+ * the setup function returns, the operation structure needs to contain a
+ * copy of the key.
  */
 /*
  *  Copyright (C) 2018, ARM Limited, All Rights Reserved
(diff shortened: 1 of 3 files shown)


include/psa/crypto_struct.h
```diff
@@ -12,6 +12,26 @@
...
+ * In particular, it is not guaranteed that a freshly-initialized structure
+ * is all-zero: we initialize structures to something like `{0, 0}`, which
+ * is only guaranteed to initializes the first member of the union;
+ * GCC and Clang initialize the whole structure to 0 (at the time of writing),
+ * but MSVC and CompCert don't.
+ *
+ * In Mbed Crypto, multipart operation structures live independently from
+ * the key. This allows Mbed Crypto to free the key objects when destroying
+ * a key slot. If a multipart operation needs to remember the key after
+ * the setup function returns, the operation structure needs to contain a
```
yanesca: What does this mean for opaque keys?

library/psa_crypto.c
```diff
@@ -1050,7 +1050,7 @@ psa_status_t psa_destroy_key( psa_key_handle_t handle )
 #endif /* MBEDTLS_PSA_CRYPTO_SE_C */
 
 #if defined(MBEDTLS_PSA_CRYPTO_STORAGE_C)
-    if( slot->attr.lifetime == PSA_KEY_LIFETIME_PERSISTENT )
+    if( slot->attr.lifetime != PSA_KEY_LIFETIME_VOLATILE )
```
yanesca: Could you please explain what difference this change makes?

library/psa_crypto.c
```diff
@@ -1041,18 +1041,30 @@ psa_status_t psa_destroy_key( psa_key_handle_t handle )
         if( status != PSA_SUCCESS )
         {
             (void) psa_crypto_stop_transaction( );
-            /* TOnogrepDO: destroy what can be destroyed anyway */
-            return( status );
+            /* We should still try to destroy the key in the secure
```
yanesca: Did you omit the TODO label on purpose here?

library/psa_crypto.c
gilles-peskine-arm: As the log message says:

> SE keys: make psa_destroy_key remove the key from storage

Before, for a key in a secure element (`lifetime` is neither `PERSISTENT` nor `VOLATILE`), the metadata wasn't removed, so the key object kept existing even though the key material was destroyed.

library/psa_crypto.c
gilles-peskine-arm: That's more of a lack of robustness rather than an outright bug or missing functionality. It is a blurry line though.

include/psa/crypto_struct.h
gilles-peskine-arm: We haven't yet implemented any multipart operation with opaque keys. When we do, we'll need to update this description, quite possibly with a different relationship between the operation object and the key object.

Issue #49, opened 2019-02-11
Undefined behavior in mbedtls_platform_zeroize()


Patater:
### Description

Calling `memset()` with a non-NULL buffer argument is undefined in C. mbedtls_platform_zeroize() can end up calling `memset()` directly (via a function pointer). It might be good to make `mbedtls_platform_zeroize()` have defined behavior when the buffer argument in NULL and size is 0.

### Issue request type

    [ ] Question
    [ ] Enhancement
    [x] Bug

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTCRYPT-672*

PR #222, opened 2019-08-15
Check for zero length and NULL buffer pointer.

jainvikas8:
In reference to issue https://github.com/ARMmbed/mbed-crypto/issues/49

Backport to 2.16: https://github.com/ARMmbed/mbedtls/pull/2793

After [analysis](https://github.com/ARMmbed/mbedtls/pull/2794) this turned out not to be applicable to 2.7.

### DIFF
--- a/library/platform_util.c
+++ b/library/platform_util.c
@@ -72,7 +72,10 @@ static void * (* const volatile memset_func)( void *, int, size_t ) = memset;
 
 void mbedtls_platform_zeroize( void *buf, size_t len )
 {
-    memset_func( buf, 0, len );
+    MBEDTLS_INTERNAL_VALIDATE( len == 0 || buf != NULL );
+
+    if( len > 0 )
+        memset_func( buf, 0, len );
 }
 #endif /* MBEDTLS_PLATFORM_ZEROIZE_ALT */
 


Patater:
Should this have a ChangeLog entry, like https://github.com/ARMmbed/mbedtls/pull/2793/commits/e5c9e796b2c80fc12093891c94027f8c7b0e89c9 ?

Patater:
> Should this have a ChangeLog entry, like [ARMmbed/mbedtls@e5c9e79](https://github.com/ARMmbed/mbedtls/commit/e5c9e796b2c80fc12093891c94027f8c7b0e89c9) ?

Nevermind, this is to Mbed Crypto which doesn't have a Changelog. Misread it as going to Mbed TLS.

Patater:
Will merge at same time as https://github.com/ARMmbed/mbedtls/pull/2793 and https://github.com/ARMmbed/mbedtls/pull/2794

athoelke:
Surely if `buf == NULL` and `len > 0` there is a serious programming error somewhere? This change silently ignores the problem and carries on. Is that an appropriate option for a security service?

library/platform_util.c
```diff
@@ -72,7 +72,8 @@ static void * (* const volatile memset_func)( void *, int, size_t ) = memset;
 
 void mbedtls_platform_zeroize( void *buf, size_t len )
 {
-    memset_func( buf, 0, len );
+    if ( buf != NULL && len > 0 )
```
Patater: Given the discussion in https://github.com/ARMmbed/mbed-crypto/pull/239#discussion_r320234363, we should probably change this check to just `if ( len > 0 )` and avoid the NULL check. This will allow static analyzers to see that `buf` is not allowed to be NULL for this function (and, as a very small bonus, we can save a tiny amount of code space).

gilles-peskine-arm: PSA and mbedtls APIs work differently here. In PSA Crypto, a null pointer dereference is supposed to invoke the platform's equivalent of a segmentation fault. In mbedtls, in low-level modules, a null pointer dereference is supposed to invoke `MBEDTLS_PARAM_FAILED` if `MBEDTLS_CHECK_PARAMS` is enabled (and segfaults otherwise), and many high-level modules (e.g. md, cipher and pk) return an error code on null pointers.

Here, I think the correct behavior is
```
MBEDTLS_INTERNAL_VALIDATE( len == 0 || buf != NULL );
if( len > 0 )
    memset_func( buf, 0, len );
```

jainvikas8:
Forced push to accommodate reviewer change request.

library/platform_util.c
```diff
@@ -72,7 +72,10 @@ static void * (* const volatile memset_func)( void *, int, size_t ) = memset;
 
 void mbedtls_platform_zeroize( void *buf, size_t len )
 {
-    memset_func( buf, 0, len );
+    MBEDTLS_INTERNAL_VALIDATE( len == 0 || buf != NULL );
+    
+    if ( len > 0 )
```
gilles-peskine-arm: Style:
```suggestion
    if( len > 0 )
```

jainvikas8:
Forced push to correct style and trailing whitespaces.

PR #228, opened 2019-08-19
Init function variable to avoid compiler warning

jainvikas8:
Make it more robust.
Issue : https://github.com/ARMmbed/mbedtls/issues/2340

### DIFF
--- a/library/ecp.c
+++ b/library/ecp.c
@@ -2080,7 +2080,7 @@ static int ecp_mul_comb( mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
                          void *p_rng,
                          mbedtls_ecp_restart_ctx *rs_ctx )
 {
-    int ret;
+    int ret = MBEDTLS_ERR_ECP_BAD_INPUT_DATA;
     unsigned char w, p_eq_g, i;
     size_t d;
     unsigned char T_size, T_ok;


jainvikas8:
Forced pushed to fix typo commit message.

Patater:
Please copy-paste the compiler warning message into the commit message, including information about the specific toolchain and build options used, as well as explaining why the change is needed.

yanesca:
We couldn't locate the issue in the code, the original reporter didn't come back to Ron and he couldn't give the details and we couldn't reproduce. We are making this change to make the code more robust, avoid introducing an uninitialised read if it is not present and fix it if it is and we missed it.

k-stachowiak:
I can only see two main ways of getting to `return( ret )` in the function:

1. Through:

    if( T == NULL )
    {
        ret = MBEDTLS_ERR_ECP_ALLOC_FAILED;
        goto cleanup;
    }

which sets the `ret` variable.

2. directly, which unconditionally (both in the compile time and in runtime) expands:

    MBEDTLS_MPI_CHK( ecp_mul_comb_after_precomp(
        grp, R, m,
        T, T_size, w, d,
        f_rng, p_rng, rs_ctx ) );

, which, in turn, unconditionally sets the `ret` variable as well.

If we aren't observing the warning anywhere (neither locally nor in the CI or Coverity), I suggest not introducing this change. The proposed initial value doesn't express the actual error, and prevents static analyzers detecting the real lack of assignment should it ever occurs.

Patater:
Closing as not needed and not the right fix. Re-open if you can reproduce the error and know how to fix the real error (as opposed to masking it with blind initialization).

k-stachowiak:
@Patater Please note, that two backports of this PR are still open: ARMmbed/mbedtls#2796 and ARMmbed/mbedtls#2797

PR #229, opened 2019-08-19
Remove legacy psa key derivation

k-stachowiak:
The legacy function psa_key_derivation in Mbed Crypto is from an early draft of the PSA Crypto API. It was superseded in API 1.0 beta1 by an API with multiple input steps: psa_key_derivation_setup and friends.

This PR removes the `psa_key_derivation()` function and the corresponding code. It continues work merged in #160.

**Note to reviewers** Please consider possible refactoring that this change enables. I failed to find any, but am willing to perform additional changes.


library/psa_crypto.c
```diff
@@ -5111,10 +4898,7 @@ static psa_status_t psa_key_derivation_input_internal(
                                  PSA_ALG_HKDF_GET_HASH( kdf_alg ),
                                  step, data, data_length );
     }
-    else
-#endif /* MBEDTLS_MD_C */
-#if defined(MBEDTLS_MD_C)
-    if( PSA_ALG_IS_TLS12_PRF( kdf_alg ) )
+    else if( PSA_ALG_IS_TLS12_PRF( kdf_alg ) )
```
Patater: Why did we need to make this change?

k-stachowiak: Actually it's not needed. I just stumbled upon the `endif` followed by an `if` for the same condition and decided to remove them. Afterwards, it turned out we have an orphaned `else` followed by an `if` statement, so I merged them.

If I have missed a reason for the code to be the way it was, I'll gladly revert the change.

yanesca: The unit tests are failing in some builds we return different error codes  than before. I am not certain that this causes the failures, but if it does, then we need to revert it.

yanesca:
I have looked at it and couldn't find any obvious ways to refactor it and I don't have any suggestions for improvement. I can approve after the CI is fixed.

k-stachowiak:
Not all the CI jobs give the details, but after going through a bunch I think I can only see one test failure appearing for some configurations.

In `test_suite_psa_crypto` test suite, the tests:

* PSA key derivation: invalid type (0),
* PSA key derivation: invalid type (PSA_KEY_TYPE_CATEGORY_MASK),

are expected to return `PSA_ERROR_NOT_SUPPORTED`, but instead they return `PSA_ERROR_INVALID_ARGUMENT`.
That error is returned from the `psa_generate_derived_key_internal()` after failing the test for `key_type_is_raw_bytes( slot->attr.type )` being false.

This does not contradict the test's intention, thus I suggest one of two solutions:
* changing the expected error type for the failing tests,
* reintroducing the `!PSA_PRE_1_0_KEY_DERIVATION` dependency on the failing tests (or maybe even revert all the dependency changes - maybe I was wrong to remove them?).

@yanesca, @Patater what do you think?

In the meantime I'll go all the way through the CI failures to see if anything else has popped up there.

Patater:
>     * changing the expected error type for the failing tests,

Yes please. It makes sense that the error back would be `PSA_ERROR_INVALID_ARGUMENT` when passed invalid parameters like `0` or `PSA_KEY_TYPE_CATEGORY_MASK`.

yanesca:
I think the right way to fix is to update the error codes and not the tests for the following reasons:
- The way it is implemented now we return the same error code whether it is not supported or invalid argument
- The documentation does not mention that this function can return `PSA_ERROR_INVALID_ARGUMENT`
- As a consequence we might fail the PSA compliance test if we change this.

Patater:
>     * The documentation does not mention that this function can return `PSA_ERROR_INVALID_ARGUMENT`

That should be fixed.

>     * As a consequence we might fail the PSA compliance test if we change this.

The compliance tests would need to take this into account. In general, the compliance tests should not be too picky about which exact errors are returned. PSA Crypto implementations have a good amount of flexibility when it comes to the errors they choose.

Patater:
CI failures are expected API breaks. No other failures.

Issue #171, opened 2019-07-12
Double-underscore in macro names is not valid in C++


gilles-peskine-arm:
### Description

Mbed Crypto headers define an identifier that is reserved in C++. Specifically `PSA__ALG_AEAD_WITH_DEFAULT_TAG_LENGTH__CASE` in `include/psa/crypto_values.h`. C++ reserves double underscores.

Note that it isn't enough to just change the double underscore to a single underscore: `test_psa_constant_names.py` picks up macros `PSA_ALG_.*` and would need to skip this one somehow. See https://github.com/ARMmbed/mbedtls-psa/pull/258 (private link).

Acceptance criteria:

* Add a check for double underscores. To `check-names.sh`, presumably. It already checks for trailing underscores, for whatever reason.
* Obviously, make the check pass without breaking anything else.

### Issue request type

    [ ] Question
    [ ] Enhancement
    [x] Bug

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTCRYPT-826*

AndrzejKurek:
PR created: https://github.com/ARMmbed/mbed-crypto/pull/230

gilles-peskine-arm:
Fixed by https://github.com/ARMmbed/mbed-crypto/pull/230

PR #230, opened 2019-08-21
Remove double underscores from macro and add a check for it

AndrzejKurek:
This PR addresses issue https://github.com/ARMmbed/mbed-crypto/issues/171.
It removes the double underscore from a macro, adds an exception for it in "test_psa_constant_names", and introduces a check for double underscores in check-names.sh.

### DIFF
--- a/include/psa/crypto_values.h
+++ b/include/psa/crypto_values.h
@@ -1015,15 +1015,15 @@
  * \return              The corresponding AEAD algorithm with the default
  *                      tag length for that algorithm.
  */
-#define PSA_ALG_AEAD_WITH_DEFAULT_TAG_LENGTH(aead_alg)                  \
-    (                                                                   \
-        PSA__ALG_AEAD_WITH_DEFAULT_TAG_LENGTH__CASE(aead_alg, PSA_ALG_CCM) \
-        PSA__ALG_AEAD_WITH_DEFAULT_TAG_LENGTH__CASE(aead_alg, PSA_ALG_GCM) \
-        PSA__ALG_AEAD_WITH_DEFAULT_TAG_LENGTH__CASE(aead_alg, PSA_ALG_CHACHA20_POLY1305) \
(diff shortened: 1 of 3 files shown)


athoelke:
This will now result in `PSA_ALG_AEAD_WITH_DEFAULT_TAG_LENGTH_CASE` appearing in the PSA Crypto specification as all symbols that begin with `PSA_ALG_` are considered part of the architectural specification rather than part of the mbed implementation that should not be part of the interface spec.

Although we could fix this with a matching change to the tools that extracts the spec documentation from the mbed implementation (which we still do for the moment), it is not an ideal situation.

Ideally, there needs to be some reserved name spaces defined for the PSA specification, such that implementations would be able to allocate symbols for their own internal and extension APIs without risk of collision with the specification. It may be late to request that mbed not use a `psa_` or `PSA_` prefix for its own APIs - although that is likely to be what that specs will recommend to avoid clashes with future version of the specification.

gilles-peskine-arm:
> Ideally, there needs to be some reserved name spaces defined for the PSA specification, such that implementations would be able to allocate symbols for their own internal and extension APIs without risk of collision with the specification.

I agree. Does the PSA framework define this?

>  It may be late to request that mbed not use a psa_ or PSA_ prefix for its own APIs - although that is likely to be what that specs will recommend to avoid clashes with future version of the specification.

We've been inconsistent about this. The first time I added an implementation-specific function to Mbed Crypto, I called it `mbedtls_psa_xxx`. But then I didn't follow this convention for other implementation-specific functions which I considered to be experimental, possibly to be made official at a later date. For macros, I never used `MBEDTLS_PSA_xxx` because Mbed Crypto has a very fragile script that checks for possible typos in preprocessor symbols that begin with `MBEDTLS_` and I wanted to minimize PSA-related modifications, but I think now we can move past that (https://github.com/ARMmbed/mbed-crypto/issues/241).

PR #238, opened 2019-08-27
HMAC DRBG: Split entropy-gathering requests to reduce request sizes

### LINKED ISSUE #237: HMAC DRBG requires more entropy than what Mbed Crypto's entropy implementation offers
This is a report on behalf of @artokin who discovered this issue.

__Summary:__ Mbed Crypto's HMAC DRBG implementation doesn't work together with Mbed Crypto's default entropy implementation if `MBEDTLS_SHA512_C` is disabled, due to a mismatch in the amount of entropy HMAC DRBG requires in this case, and what the entropy module can provide.

__Details:__ During seeding, the HMAC DRBG implementation queries the configured entropy callback for entropy, the amount of which depending on the digest function chosen: https://github.com/ARMmbed/mbed-crypto/blob/85c78b48a94a29cea682fa67ba743c8e89459e23/library/hmac_drbg.c#L236-L250
In particular, if the chosen hash is not SHA-1, this requires at least `24 * 3/2 = 36` Bytes to be 
provided by the underlying entropy callback `entropy()`. However, because of 
https://github.com/ARMmbed/mbed-crypto/blob/85c78b48a94a29cea682fa67ba743c8e89459e23/library/entropy.c#L326-L333
Mbed Crypto's entropy implementation rejects requests beyond `MBEDTLS_ENTROPY_BLOCK_SIZE` Bytes, that is, beyond 32 Bytes if `MBEDTLS_SHA512_C` is disabled: https://github.com/ARMmbed/mbed-crypto/blob/85c78b48a94a29cea682fa67ba743c8e89459e23/include/mbedtls/entropy.h#L77-L81

This leads to failures when trying to use HMAC DRBG with the default entropy module in SHA-256-only configurations.

Apart from the issue itself, this also highlights a test gap, because we should test this very natural combination. When considering the use of Mbed Crypto in Mbed TLS, for example, the `ssl_client2` and `ssl_server2` example application should be able to use HMAC DRBG instead of  CTR DRBG if the latter is disabled, which would trigger this bug. (Internal: See PR https://github.com/ARMmbed/mbedtls-restricted/pull/646).

hanno-arm:
__Summary:__ This fixes #237 by splitting the HMAC DRBG's entropy gathering in two steps to reduce the maximum amount of entropy requested at once.

__Note:__ This PR So far only attempts to fix the bug, but doesn't yet add tests which would have witnessed it. When integrated with Mbed TLS, such tests can be added by adding support for HMAC DRBG in the example applications, but for Mbed Crypto, it's likely that the HMAC DRBG test suite will need to be enhanced. 

__Details (from commit message)__: According to SP800-90A, the DRBG seeding process should use a nonce of length `security_strength / 2` bits as part of the DRBG seed. It further notes that this nonce may be drawn from the same source of entropy that is used for the first `security_strength` bits of the DRBG seed. The present HMAC DRBG implementation does that, requesting `security_strength * 3 / 2` bits of entropy from the configured entropy source in total to form the initial part of the DRBG seed.

However, some entropy sources may have thresholds in terms of how much entropy they can provide in a single call to their entropy gathering function which may be exceeded by the present HMAC DRBG implementation even if the threshold is not smaller than `security_strength` bits. Specifically, this is the case for our own entropy module implementation which only allows requesting at most 32 Bytes of entropy at a time in configurations disabling SHA-512, and this leads to runtime failure of HMAC DRBG when used with Mbed Crypto' own entropy callbacks in such configurations.

This commit fixes this by splitting the seed entropy acquisition into two calls, one requesting `security_strength` bits first, and another one requesting `security_strength / 2` bits for the nonce.

__Backports:__
- Mbed TLS 2.16 and 2.7?
- Mbed TLS baremetal branch: https://github.com/ARMmbed/mbedtls-restricted/pull/647

### DIFF
--- a/library/hmac_drbg.c
+++ b/library/hmac_drbg.c
@@ -149,20 +149,32 @@ int mbedtls_hmac_drbg_seed_buf( mbedtls_hmac_drbg_context *ctx,
 }
 
 /*
- * HMAC_DRBG reseeding: 10.1.2.4 (arabic) + 9.2 (Roman)
+ * Internal function used both for seeding and reseeding the DRBG.
+ * Comments starting with arabic numbers refer to section 10.1.2.4
+ * of SP800-90A, while roman numbers refer to section 9.2.
  */
-int mbedtls_hmac_drbg_reseed( mbedtls_hmac_drbg_context *ctx,
-                      const unsigned char *additional, size_t len )
+static int hmac_drbg_reseed_core( mbedtls_hmac_drbg_context *ctx,
+                                  const unsigned char *additional, size_t len,
+                                  int use_nonce )
 {
     unsigned char seed[MBEDTLS_HMAC_DRBG_MAX_SEED_INPUT];
-    size_t seedlen;
+    size_t seedlen = 0;
     int ret;
 
-    /* III. Check input length */
-    if( len > MBEDTLS_HMAC_DRBG_MAX_INPUT ||
-        ctx->entropy_len + len > MBEDTLS_HMAC_DRBG_MAX_SEED_INPUT )
     {
-        return( MBEDTLS_ERR_HMAC_DRBG_INPUT_TOO_BIG );
+        size_t total_entropy_len;
+
+        if( use_nonce == 0 )
+            total_entropy_len = ctx->entropy_len;
+        else
+            total_entropy_len = ctx->entropy_len * 3 / 2;
+
+        /* III. Check input length */
+        if( len > MBEDTLS_HMAC_DRBG_MAX_INPUT ||
+            total_entropy_len + len > MBEDTLS_HMAC_DRBG_MAX_SEED_INPUT )
+        {
+            return( MBEDTLS_ERR_HMAC_DRBG_INPUT_TOO_BIG );
+        }
     }
 
     memset( seed, 0, MBEDTLS_HMAC_DRBG_MAX_SEED_INPUT );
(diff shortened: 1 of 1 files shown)


mpg:
Adding "needs: backports" as bugs always need backports unless there's a specific reason they don't.

mpg:
Labeling as "needs: work" as there were some changes in the baremetal version that should probably be done here as well.

library/hmac_drbg.c
```diff
@@ -233,20 +280,15 @@ int mbedtls_hmac_drbg_seed( mbedtls_hmac_drbg_context *ctx,
...
+                       md_size <= 28 ? 24 : /* 224-bits hash -> 192 bits */
+                       32;  /* better (256+) -> 256 bits */
 
-    if( ( ret = mbedtls_hmac_drbg_reseed( ctx, custom, len ) ) != 0 )
+    if( ( ret = hmac_drbg_reseed_core( ctx, custom, len,
+                                       1 /* add nonce */ ) ) != 0 )
+    {
         return( ret );
-
-    ctx->entropy_len = entropy_len;
```
Patater: Previously, it appears `ctx->entropy_len` was set to a value that included the nonce (32 * 3 / 2 at most). Now, it only includes without nonce (just 32). I looked at `hmac_drbg_reseed_core()` as well to see if it would be setting `ctx->entropy_len`, but it doesn't add the nonce length to `ctx->entropy_len`.

Is it OK that ctx->entropy_len no longer has the nonce length included?

hanno-arm: @Patater `ctx->entropy_len` was previously set to `3/2 * security` just for the duration of the initial seeding, and was reset to `security` afterwards. Now this single use of factor `3/2` is controlled through the added parameter, while `ctr->entropy_len` is always `security`.

Patater: OK, I missed that it was reset to `security` length

Patater:
CI test failure in merge run is only missing Everest Crypto identifiers, which is a known issue until https://github.com/ARMmbed/mbedtls/pull/2799 is merged

Issue #203, opened 2019-08-06
Test "nonexistent" in psa its suite causes NULL p_data to propagate to fwrite in psa_its_set


nanokatze:
# Description

Mbed Crypto 1.1.1
Toolchain GCC 9.1.1
Target Linux 5.1.20 x86_64 (Fedora 30 5.1.20-300)

Test
https://github.com/ARMmbed/mbed-crypto/blob/mbedcrypto-1.1.1/tests/suites/test_suite_psa_its.function#L167
will propagate NULL p_data to fwrite
https://github.com/ARMmbed/mbed-crypto/blob/mbedcrypto-1.1.1/library/psa_its_file.c#L212
which must not be NULL.  This is reproduced with appending sanitizer flags -fsanitize=address,undefined flags to CFLAGS in library/Makefile and tests/Makefile and building and running the tests with sanitizers enabled.

### Issue request type

    [ ] Question
    [ ] Enhancement
    [X] Bug

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTCRYPT-850*

gilles-peskine-arm:
```
psa_its_set_wrap( uid, 0, NULL, 0 )
```
ends up calling `psa_its_set( uid, 0, NULL, 0 )`. Since the buffer size is 0, `NULL` is a valid buffer pointer.

The C standard doesn't see it that way (at least for `mem*` functions, NULL/0 is UB; for stdio functions, it doesn't seem to be specified). So we should indeed avoid calling `fwrite` in this case.

Oddly, we _do_ test with Asan and our CI is passing. But I can reproduce the failure locally with gcc5 and `make CFLAGS='-O2 -fsanitize=address,undefined'`. So we should figure out what options we're using, and tweak or add to the CI job.

gilles-peskine-arm:
Ok, I understand why CI is passing. I thought we always used Asan and UBsan together (`-fsanitize=address,undefined`), but in `CMakelists.txt`, we activate UBSan with Clang but not with GCC. And we only run that test with ASan on CI through cmake with GCC, so we don't run it with UBSan.

AndrzejKurek:
Created a PR here: https://github.com/ARMmbed/mbed-crypto/pull/239

Patater:
PR #239 merged with a fix for this.

PR #239, opened 2019-08-28
Remove a potential call to fwrite with null buffer. Add UBsan testing

AndrzejKurek:
This PR handles issue https://github.com/ARMmbed/mbed-crypto/issues/203.
First commit removes a potential call to fwrite with a null buffer, and second extends testing.

Tested locally on Ubuntu 16.04 with gcc 5.4 - after extending the testing in CMakeLists and before adding the null check Asan type build fails (and therefore all.sh too). With these two commits - everything passes.

Before merging, this PR will also need a changelog entry to credit nanokatze who has filed this issue.


library/psa_its_file.c
```diff
@@ -214,9 +214,12 @@ psa_status_t psa_its_set( psa_storage_uid_t uid,
     n = fwrite( &header, 1, sizeof( header ), stream );
     if( n != sizeof( header ) )
         goto exit;
-    n = fwrite( p_data, 1, data_length, stream );
-    if( n != data_length )
-        goto exit;
+    if( p_data != NULL )
```
k-stachowiak: Is this enough? The Doxygen for this function says, among others, it will return `PSA_ERROR_INVALID_ARGUMENT`, if:

> The operation failed because one of the provided pointers(`p_data`) is invalid, for example is `NULL` or references memory the caller cannot access.

If the docs are up to date, I suggest returning the above error.

athoelke: It is not required for the implementation to report this error code in the case of an invalid memory reference, as a memory access violation is also a reasonable outcome of such a call. But an implementation is permitted to detect and report invalid memory references using this error code.

It would be helpful if the spec provided guidance on how memory buffers are handled in the API, for comparison see the PSA-FF and PSA Crypto specifications.

In this API `p_data` should be permitted to have any value if `data_length==0`. In that specific case:
* `fwrite()` has fully defined behaviour (do nothing)
* `psa_its_set()` has agreed behaviour (but it is not explicit in the spec), when it will create a zero-length data asset.

However, I can see that avoiding the potential UB when `(data_length > 0 && p_data == NULL)` is important due to the risk of leaving a half-populated ITS storage file in the system. So maybe you need:

```
if (p_data == NULL && data_length > 0)
    return PSA_ERROR_INVALID_ARGUMENT;
```

further up the function, before creating/opening the file and storing any data.

AndrzejKurek: I assumed that creating a file with a good header but no data is a valid case. Adding such check will change behaviour - is this what we want?

gilles-peskine-arm: `data_length == 0 && p_data == NULL` is a valid use case. It creates an empty file. Crypto is unlikely to need it, but this implementation of ITS is meant to be generic.

`data_length != 0 && p_data == NULL` is not a valid input. In the PSA Crypto API, there is implementation guidance to use the normal handling of invalid pointer dereferences (typically terminating the application and if possible logging the fault), and not return an error from the function. (The rationale for this guidance is that returning an error has no benefit since it's a programmer error and not a recoverable error, it makes errors harder to debug, it confuses static analyzers that have heuristics around nullable pointers by making them think that a null input is ok, and the code to check for null is pure overhead with no benefit.) I recommend following this guidance elsewhere in PSA.

gilles-peskine-arm: This should be `if( data_length == 0 )`. Checking for null silently writes nothing instead of faulting if `p_data` is null but `data_length != 0`.

AndrzejKurek:
Updated the PR. Did not add an additional test to fill in the testing gap after removing MEMORY_BUFFER_ALLOC, since it's bound to be removed anyway here: https://github.com/ARMmbed/mbedtls/pull/2469

library/psa_its_file.c
```diff
@@ -195,6 +195,9 @@ psa_status_t psa_its_set( psa_storage_uid_t uid,
     psa_its_file_header_t header;
     size_t n;
 
+    if( p_data == NULL && data_length > 0 )
```
gilles-peskine-arm: I'd prefer not to have that check, mainly because it fools static analyzers into thinking that it's ok to call this function with `p_data == NULL`, also because it makes calls with NULL harder to debug. However, the ITS interface explicitly documents that the function returns `PSA_ERROR_INVALID_ARGUMENT` in that case. So we need to make a call between friendliness to static analysis (and ease of debugging) and compliance. I vote for static analysis. @Patater What's your call?

Patater: I think not being compliant will complicate compliance testing, which hurts maintainability as much as fooling static analyzers does. Can we change the standard?

gilles-peskine-arm: How important compliance is depends on what the objective of `psa_its.c` is. It started out as solely for testing Mbed Crypto without relying on any third-party code. But I believe @paulhowardarm now intends to use it in production on Linux.

PSA storage is now officially at 1.0, so we can't change it until some future release (1.1 or 2.0).

Patater: Ah, this is for PSA ITS File. We don't care about compliance with this code. I'm up for smaller code size and more effective static analysis: which means no NULL check here. We don't intend to do compliance testing on our PSA ITS file implementation.

@simonqhughes is working on implementing PSA ITS for Linux. He used our adapter as a starting point, but has built on it quite a bit. He intends his implementation to pass compliance. @paulhowardarm should get in touch and use that implementation.

gilles-peskine-arm: I filed a change request internally: https://github.com/ARMmbed/psa_trusted_storage_api/issues/86

athoelke: I don't think that it is _required_ that the API returns `INVALID_ARGUMENT` in case of an invalid buffer pointer (not all are easy to detect unless we are in a PSA-FF type environment). `buf==NULL` is an _example_ of a argument error can _can_ result in that error code.

The basic principle: as for almost any library API ever written, passing garbage pointers can have undefined behaviour, ranging from application termination to loss of control flow when overwriting the stack to bricking the device if writing to some flash control register.

The concern here is that because:
1. The standard explicitly states that passing `p_data==NULL && len>0` to `fwrite()` is `UNDEFINED BEHAVIOUR`, AND
2. Compilers build in knowledge of standard library functions, AND
3. Compilers have the ability to statically determine that _some_ calls to `fwrite()` will fall into this UB trap, AND
4. Aggressive optimisation exploits the standard definition of UB => program is meaningless => it doesn't matter what code is generated AT ALL, AND
5. The generated code might not fail at runtime in the way the developer expected: exception when NULL is written to. (It is unknown how bad the result could actually be - see 4)
6. So, UB sanitizers (which are often a good thing) can be conservative in identifying potential sites where UB is provoked, driving the developer to insert excessive NULL checks to keep the sanitizer quiet.

So can we manage without the NULL check in `psa_it_set()`:
- does a `data_len > 0` check eliminate the problem cases?
 - if UBSAN doesn't like it - are there specific call sites of `psa_its_set()` where it can see a potential data flow path which results in `p_data==NULL`? Can we ensure that doesn't happen?

---

_Side note: On the the whole UB is evil, it would have been much better if the ISO C standards committee had been less lazy and documented reasonable constraints on implementation behaviour in these circumstances_

gilles-peskine-arm: > does a data_len > 0 check eliminate the problem cases?

Wrapping the `fwrite` call in `if (data_len > 0)` (with no `else`) makes the function behave correctly when called with valid arguments (`p_data` is valid or `data_len` is zero).

> are there specific call sites of psa_its_set() where it can see a potential data flow path which results in p_data==NULL?

In PSA Crypto, at the moment, no. In the future, maybe, but of course only with `data_len==0`. Unless there's a bug in PSA crypto (or a compiler bug, or a runtime glitch…).

> it would have been much better if the ISO C standards committee had been less lazy and documented reasonable constraints on implementation behaviour in these circumstances

If they'd done that, they'd be called the ISO Rust standards committee. There's no realistic way to limit the behavior of `free(p); *p = 3;` on common hardware. Most processors don't even have a nasal demon peripheral.

athoelke: > > does a data_len > 0 check eliminate the problem cases?
> 
> Wrapping the `fwrite` call in `if (data_len > 0)` (with no `else`) makes the function behave correctly when called with valid arguments (`p_data` is valid or `data_len` is zero).

Sorry, I meant: does adding just the `data_len>0` check stop the UB sanitizer complaining about UB calls to `fwrite()`.

> > are there specific call sites of psa_its_set() where it can see a potential data flow path which results in p_data==NULL?
> 
> In PSA Crypto, at the moment, no. In the future, maybe, but of course only with `data_len==0`. Unless there's a bug in PSA crypto (or a compiler bug, or a runtime glitch…).

Well, those are the bugs we want UB Sanitizer to detect - definite cases of passing NULL pointers without zero length data.

> > it would have been much better if the ISO C standards committee had been less lazy and documented reasonable constraints on implementation behaviour in these circumstances
> 
> If they'd done that, they'd be called the ISO Rust standards committee. There's no realistic way to limit the behavior of `free(p); *p = 3;` on common hardware. Most processors don't even have a nasal demon peripheral.

By 'reasonable constraints' I meant that the _implementation_ doesn't do "anything surprising". In your scenario, causing an exception, or overwriting some other object with `3` is "reasonable behaviour" - even if the consequences are hard to enumerate. The *compiler* deciding to generate a program that computes PI to a million decimal places instead of attempting to do what the program appears to do is not "reasonable behaviour", but this is permitted by the standard. For more background see https://blog.regehr.org/archives/213

gilles-peskine-arm: Adding the check for `data_len > 0` fixes the actual problem and also causes UBsan not to complain. UBsan analyzed this correctly.

In practice, `free(p); *p = 3;` (or some variant like `free(p); p->foo = 3;`) corrupts data structures used by malloc and the behavior is truly undefined up to the capabilities of the hardware. Especially for us in the security business, where the choice to run this particular code is likely to have been made by an adversary whose objective is to do something that the defender considers unreasonable. Removing this undefined behavior hole means imposing constraints on memory management, and then C wouldn't be C anymore. I'm not saying that I like it, but C is what it is.

athoelke: > In practice, `free(p); *p = 3;` (or some variant like `free(p); p->foo = 3;`) corrupts data structures used by malloc and the behavior is truly undefined up to the capabilities of the hardware.

You underestimate the potential interpretation of the standard. The C standard considers that the compiler **and** runtime **and** system are all "the C abstract machine" that interprets the program.

We might expect that `free(p); *p = 3;` results in some machine code like:

    mov r0, r11
    bl free
    mov r0, #3 
    str [r11], r0

after which all hell could break loose in the system because we have just scribbled on allocator meta-data. This is only a _moderately bad_ result.

The C standard says that because writing to memory that has been freed is undefined behaviour, the **program** is incorrect and the "C abstract machine" can do whatever it wants. In practice the compiler detecting this might eliminate output code that only serves to implement scenarios that exhibit undefined behaviour - so here it might choose to make **the entire function** containing the above code a NOP.

It is not possible for the developer to reason about the impact that such unexpected transformations will have on the operation of their code.

gilles-peskine-arm: I'm well aware of what undefined behavior is, thank you. My remark was about how to approach UB when thinking about security. As a defender, the worst result is that the attacker gets to execute arbitrary code in your execution space.

And the secondmost reliable way to avoid UB is to root it out with static analysis. (The most reliable is of course to program in a language that doesn't have UB.) Hence my push towards 1. actually using static analysis tools and 2. making the code friendly to these tools.

athoelke: Sorry, I should have not started my reply with "you" - this wasn't meant to be a personal response.

I totally agree in the importance of using static tooling to identify UB, and ensuring that the code is both friendly to such tools and that all reports of issues from the tool are followed up.

My comment was attempting to dispel the myth that the results of UB in source code are merely the side effects of memory scribbles or numeric overflows (or similar) that result from the execution of the **expected** machine code. UB makes these expectations invalid in the long run (new versions or different toolchains), if not at present.

Which just reinforces the point about the need for using static tools that help identify and eliminate UB.

AndrzejKurek: Ok, so do we agree on removing this check?

AndrzejKurek: Done.

PR #240, opened 2019-09-02
Enable 64-bit bignum limbs and add optimized multiplication for Aarch64

Ko-:
This was ARMmbed/mbedtls#1964. The CLA was approved and @yanesca and @Patater reviewed this.


Patater:
Thanks for your contribution, @Ko- !

PR #244, opened 2019-09-04
Update function return codes

adrianlshaw:
New PR containing code from #213 

Consistent ordering of listed errors will be fixed in a separate fix in the future.


Patater:
This still has a merge commit in it. :(

Patater:
We prefer to keep the review comments contained in one PR thread where that makes most sense to do. For PR https://github.com/ARMmbed/mbed-crypto/pull/213, updating the branch there makes sense so we can keep the review history all together. Closing in favor of https://github.com/ARMmbed/mbed-crypto/pull/213 being updated.

PR #245, opened 2019-09-04
Corrected mpi_uint_bigendian_to_host_c for big endian platforms.

RobbieDubya:
Original version was reversing the input number. Note that this function is a last resort which is used by mpi_unt_bigendian_to_host when it cannot find a better way to convert, and uses pure code, not compiler defines / features.

### DIFF
--- a/library/bignum.c
+++ b/library/bignum.c
@@ -743,9 +743,20 @@ static mbedtls_mpi_uint mpi_uint_bigendian_to_host_c( mbedtls_mpi_uint x )
 {
     uint8_t i;
     mbedtls_mpi_uint tmp = 0;
-    /* This works regardless of the endianness. */
-    for( i = 0; i < ciL; i++, x >>= 8 )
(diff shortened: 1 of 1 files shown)


hanno-arm:
Note: This is a duplicate of https://github.com/ARMmbed/mbedtls/pull/2623 which hasn't yet been side-ported to Mbed Crypto.

RobbieDubya:
Cool - did not look in mbedtls. 

That version still does redundant work when the platform is bigendian though.

Patater:
I've merged https://github.com/ARMmbed/mbedtls/pull/2623 to the Mbed Crypto development branch.

@RobbieDubya Thanks for your contribution and poking us to fix the issue. Let us know if the now merged fix doesn't work for you

Patater:
@RobbieDubya If your platform sets the expected endian macros, you shouldn't have to use `mpi_uint_bigendian_to_host_c()` https://github.com/ARMmbed/mbed-crypto/blob/826907736f33136bf2209a784d332c11b43e5636/library/bignum.c#L761-L763

PR #247, opened 2019-09-05
Update the behavior of key handles

athoelke:
* open outputs distinct key handles
* each handle must be closed
* destroying a key does not invalidate other handles
* closing a key can/might fail an active operation (but not required)

This aligns the API documentation with the overview documentation (private link ARMmbed/psa-crypto#224).

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -414,7 +414,7 @@ psa_status_t psa_open_key(psa_key_id_t id,
  * Closing the key handle makes the handle invalid, and the key handle
  * must not be used again by the application.
  *
- * \note If the key handle was used to setup an active
+ * \note If the key handle was used to set up an active
  * :ref:\`multipart operation <multipart-operations>\`, then closing the
  * key handle can cause the multipart operation to fail. Applications should
  * maintain the key handle until after the multipart operation has finished.
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -361,15 +361,18 @@ void psa_reset_key_attributes(psa_key_attributes_t *attributes);
  * keys that can be opened with psa_open_key(). Such keys have a key identifier
  * in the vendor range, as documented in the description of #psa_key_id_t.
  *
- * The application must eventually close the handle with psa_close_key()
- * to release associated resources. If the application dies without calling
- * psa_close_key(), the implementation should perform the equivalent of a
- * call to psa_close_key().
+ * The application must eventually close the handle with psa_close_key() or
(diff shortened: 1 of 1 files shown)


include/psa/crypto.h
```diff
@@ -411,8 +414,10 @@ psa_status_t psa_open_key(psa_key_id_t id,
  * Closing the key handle makes the handle invalid, and the key handle
  * must not be used again by the application.
  *
- * If the key is currently in use in a multipart operation, then closing the
- * last remaining handle to the key will abort the multipart operation.
+ * \note If the key handle was used to setup an active
```
gilles-peskine-arm: Nit: I don't think “setup” is accepted as a verb yet: it should be “set up”.

In fact https://english.stackexchange.com/questions/93243/login-and-payoff-are-nouns-but-can-they-be-used-as-verbs makes some good points: if you replace “an active multipart operation” by “it”, the sentence becomes “… to set it up”, not “\*to setup it”, so there's still a space when the complement isn't a pronoun.

athoelke: Thanks. That briefly crossed my mind while writing it, but didn't return to re-read it.

include/psa/crypto.h
```diff
@@ -361,15 +361,18 @@ void psa_reset_key_attributes(psa_key_attributes_t *attributes);
...
+ * The application must eventually close the handle with psa_close_key() or
+ * psa_destroy_key() to release associated resources. If the application dies
+ * without calling one of these functions, the implementation should perform
+ * the equivalent of a call to psa_close_key().
  *
  * Some implementations permit an application to open the same key multiple
- * times. Applications that rely on this behavior will not be portable to
- * implementations that only permit a single key handle to be opened. See
- * also :ref:\`key-handles\`.
+ * times. If this is successful, each call to psa_open_key() will return a
```
k-stachowiak: This change seems to obsolete the first sentence in this paragraph. The case of opening the same key multiple times has now been moved to a separate note below. Maybe the first sentence in this paragraph should also be moved there?

athoelke: There is an intended difference in the two paragraphs. The first paragraph is trying to state that:
1. An implementation is permitted to open a key more than once, but is not required to support this.
2. If it does, it must return distinct keys handles for each call.

The following paragraph is an informative note for application developers regarding the implication of depending on opening a key multiple times. It can be deduced from the rules above - as an implementation is permitted to refuse to open a key more than once - but the note just makes this effect clear from the app perspective, to reduce the risk of accidentally non-portable code.

Patater:
CI passing on merge result, which is good enough for merging.

PR #246, opened 2019-09-04
Secure element driver interface test strategy

gilles-peskine-arm:
I wrote a test strategy document describing what I think we should test about the secure element driver interface (the Mbed Crypto code — this is not about how to test drivers).

PR #251, opened 2019-09-10
Invasive testing strategy

gilles-peskine-arm:
Documentation only. Explore potential strategies for invasive testing. Formulate some rules and guideline. Succinctly describe some solutions.

This is a proposal submitted for discussion.

I deliberately included solutions with “TODO” in them for things that we should do, but aren't doing yet.

I've changed headers to define macros mentioned explicitly in the strategy document so that one can start implementing parts of the strategy based on this PR. But implementing the proposed strategy is out of scope of this PR.

### DIFF
--- a/docs/architecture/testing/mbed-crypto-invasive-testing.md
+++ b/docs/architecture/testing/mbed-crypto-invasive-testing.md
@@ -20,7 +20,9 @@ Do not add test-specific interfaces if there's a practical way of doing it anoth
 
 ### Reliance on internal details
 
-In unit tests and in test programs, it's ok to include header files from `library/`. In contrast, sample programs must not include header files from `library/`.
+In unit tests and in test programs, it's ok to include header files from `library/`. Do not define non-public interfaces in public headers (`include/mbedtls` has `*_internal.h` headers for legacy reasons, but this approach is deprecated). In contrast, sample programs must not include header files from `library/`.
+
+Sometimes it makes sense to have unit tests on functions that aren't part of the public API. Declare such functions in `library/*.h` and include the corresponding header in the test code. If the function should be `static` for optimization but can't be `static` for testing, declare it as `MBEDTLS_STATIC_TESTABLE`, and make the tests that use it depend on `MBEDTLS_TEST_HOOKS` (see [“rules for compile-time options”](#rules-for-compile-time-options)).
 
 If test code or test data depends on internal details of the library and not just on its documented behavior, add a comment in the code that explains the dependency. For example:
 
@@ -53,7 +55,7 @@ The code calls a function `mbedtls_foo()`. Usually this a macro defined to be a 
 
 Sometimes the substitutable function is a `static inline` function that does nothing (not a macro, to avoid accidentally skipping side effects in its parameters), to provide a hook for test code; such functions should have a name that starts with the prefix `mbedtls_test_hook_`. In such cases, the function should generally not modify its parameters, so any pointer argument should be const. The function should return void.
 
-With `MBEDTLS_TEST_HOOKS` set, `mbedtls_foo` is a global variable of function pointer type. This global variable is initialized to the system function, or to a function that does nothing. The global variable is defined in a header in `library.h` such as `psa_crypto_invasive.h`.
+With `MBEDTLS_TEST_HOOKS` set, `mbedtls_foo` is a global variable of function pointer type. This global variable is initialized to the system function, or to a function that does nothing. The global variable is defined in a header in the `library` directory such as `psa_crypto_invasive.h`.
 
 In test code that needs to modify the internal behavior:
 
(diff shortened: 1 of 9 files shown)


docs/architecture/testing/mbed-crypto-invasive-testing.md
```diff
@@ -0,0 +1,352 @@
...
+
+Solution: TODO. Currently the various test suites do it differently.
+
+#### PSA storage resilience
+
+Goal: test the resilience of PSA storage against power failures.
+
+Solution: TODO.
+
+See the [secure element driver interface test strategy](driver-interface-test-strategy.html) for more information.
```
gilles-peskine-arm: Note: currently in https://github.com/ARMmbed/mbed-crypto/pull/246. [Rendered HTML](https://github.com/gilles-peskine-arm/mbed-crypto/blob/psa-se_driver-interface_test_strategy/docs/architecture/testing/driver-interface-test-strategy.md).

mpg:
I read the document and it looks very good.

I'd just like to check if I understood something right on a concrete example. Sometimes we have internal functions that we'd like to unit-test because that's the easiest and most efficient way. For example, `mbedtls_ssl_dtls_replay_update()` and `mbedtls_ssl_dlts_replay_check()`. Currently they're unconditonaly non-static (visible by the linker) and declared in `ssl_internal.h`, which is included from the file defining the test functions, but is not supposed to be included by applications.

That's mostly satisfying except that making the functions static would allow more compiler optimizations and possibly reduce the code size. This is a serious concern, as in another instance in the baremetal branch we observed significant reductions in code size (several hundred bytes) just by making a bunch of functions static (see https://github.com/ARMmbed/mbedtls-restricted/pull/652 for details).

IIUC, the proposed rules would encourage us to:

* in non- `MBEDTLS_TEST_HOOKS` builds: make the functions static (allowing the compiler to optimize as fully as it can), but then skip the associated unit tests in the test suite.
* in `MBEDTLS_TEST_HOOKS`-enabled builds: make the functions non-static (potentially increasing the code size) and build and run the associated unit tests as we do now.

Is that correct?

gilles-peskine-arm:
@mpg Thanks for the feedback. This is a use case that I hadn't thought of. I agree with your solution. I'll add a paragraph about this to the document.

If a function isn't part of the public API, I encourage declaring it in a header in `library`, not in a public header. We do that for PSA (see `library/psa_*.h`); most of these are exposed for cross-module references, but `psa_crypto_invasive.h` is intended for test only and would be wrapped in a `MBEDTLS_TEST_HOOKS` guard with the new rules.

mpg:
> If a function isn't part of the public API, I encourage declaring it in a header in library, not in a public header.

docs/architecture/testing/mbed-crypto-invasive-testing.md
```diff
@@ -0,0 +1,354 @@
...
+
+We need to balance the following goals, which are sometimes contradictory.
+
+* Coverage: we need to test behaviors which are not easy to trigger by using the API or which cannot be triggered deterministically, for example I/O failures.
+* Correctness: we want to test the actual product, not a modified version, since conclusions drawn from a test of a modified product may not apply to the real product.
+* Effacement: the product should not include features that are solely present for test purposes, since these increase the attack surface and the code size.
+* Portability: tests should work on every platform. Skipping tests on certain platforms may hide errors that are only apparent on such platforms.
+* Maintenability: tests should only enforce the documented behavior of the product, to avoid extra work when the product's internal or implementation-specific behavior changes. We should also not give the impression that whatever the tests check is guaranteed behavior of the product which cannot change in future versions.
+
+Where those goals conflit, we should at least mitigate the goals that cannot be fulfilled, and document the architectural choices and their rationale.
```
dgreen-arm: Typo: `conflit` -> `conflict`

docs/architecture/testing/mbed-crypto-invasive-testing.md
```diff
@@ -0,0 +1,354 @@
...
+
+### General goals
+
+We need to balance the following goals, which are sometimes contradictory.
+
+* Coverage: we need to test behaviors which are not easy to trigger by using the API or which cannot be triggered deterministically, for example I/O failures.
+* Correctness: we want to test the actual product, not a modified version, since conclusions drawn from a test of a modified product may not apply to the real product.
+* Effacement: the product should not include features that are solely present for test purposes, since these increase the attack surface and the code size.
+* Portability: tests should work on every platform. Skipping tests on certain platforms may hide errors that are only apparent on such platforms.
+* Maintenability: tests should only enforce the documented behavior of the product, to avoid extra work when the product's internal or implementation-specific behavior changes. We should also not give the impression that whatever the tests check is guaranteed behavior of the product which cannot change in future versions.
```
dgreen-arm: Typo: `Maintenability` -> `Maintainability`

Also appears in a few other places.

gilles-peskine-arm:
I rebased on top of `development` to resolve the conflict in `docs/architecture/Makefile` and also fixed the typos that Darryl pointed out. No semantic change.

gilles-peskine-arm:
Moved to https://github.com/ARMmbed/mbedtls/pull/3121

PR #255, opened 2019-09-11
Add mbed os conf tests

RonEld:
Add a test in `all.sh` that configures the default configuration in `Mbed OS` and fix test dependencies discovered from it.


tests/scripts/all.sh
```diff
@@ -1120,6 +1120,70 @@ component_test_cmake_out_of_source () {
...
 
+component_test_basic_mbed_os_conf () {
+msg "build+test: Mbed OS default configuration"
+    scripts/config.pl unset MBEDTLS_NET_C
+    scripts/config.pl unset MBEDTLS_TIMING_C
+    scripts/config.pl unset MBEDTLS_HAVE_TIME_DATE
+    scripts/config.pl unset MBEDTLS_FS_IO
+    scripts/config.pl unset MBEDTLS_PSA_ITS_FILE_C # requires a filesystem
+    scripts/config.pl unset MBEDTLS_PSA_CRYPTO_STORAGE_C # requires PSA ITS
+    #scripts/config.pl set MBEDTLS_NO_PLATFORM_ENTROPY Need some entropy
```
gilles-peskine-arm: Why is this line commented out?

tests/scripts/all.sh
```diff
@@ -1120,6 +1120,70 @@ component_test_cmake_out_of_source () {
...
+    scripts/config.pl unset MBEDTLS_KEY_EXCHANGE_ECDH_ECDSA_ENABLED
+    scripts/config.pl unset MBEDTLS_KEY_EXCHANGE_ECDH_RSA_ENABLED
+    scripts/config.pl unset MBEDTLS_SSL_FALLBACK_SCSV
+    scripts/config.pl unset MBEDTLS_SSL_CBC_RECORD_SPLITTING
+    scripts/config.pl unset MBEDTLS_SSL_PROTO_TLS1
+    scripts/config.pl unset MBEDTLS_SSL_PROTO_TLS1_1
+    scripts/config.pl unset MBEDTLS_SSL_TRUNCATED_HMAC
+    scripts/config.pl unset MBEDTLS_PLATFORM_TIME_TYPE_MACRO
+    scripts/config.pl set MBEDTLS_MPI_MAX_SIZE     512
+    make CC=gcc CFLAGS='-Werror -O1' all test
```
gilles-peskine-arm: Maybe do a 32-bit build to be slightly closer to Mbed OS?

tests/scripts/all.sh
```diff
@@ -1120,6 +1120,70 @@ component_test_cmake_out_of_source () {
     unset MBEDTLS_ROOT_DIR
 }
 
+component_test_basic_mbed_os_conf () {
+msg "build+test: Mbed OS default configuration"
+    scripts/config.pl unset MBEDTLS_NET_C
```
gilles-peskine-arm: Rather than call `config.pl` a bunch of times, how about defining the Mbed OS configuration in a file in `configs`? Or as a filter in `config.pl`, if we want to define it by comparison to the default configuration?

gilles-peskine-arm:
General question: where is the Mbed OS configuration defined?

tests/scripts/all.sh
RonEld: Because we need some entropy sources, otherwise `psa_crypto_init()` will fail

tests/scripts/all.sh
RonEld: ok.
The idea was to check Mbed OS configuration, but this won't harm

RonEld:
>General question: where is the Mbed OS configuration defined?

When Mbed TLS is imported to Mbed OS, the import script is run, adjusting the Mbed TLS config, in the [`adjust_config.sh`](https://github.com/ARMmbed/mbed-os/blob/master/features/mbedtls/importer/adjust-config.sh) script.

gilles-peskine-arm:
> When Mbed TLS is imported to Mbed OS, the import script is run, adjusting the Mbed TLS config, in the `adjust_config.sh` script.

Thanks. Could we move that script to the Mbed TLS repository, and use that from `all.sh`? Looking at its commit history, it's mostly modified by Mbed TLS team members.

RonEld:
As Gilles mentioned,
If we have a reference configuration for Mbed OS, that will be used by the importer script, it means that every new configuration in the main `config.h` file, will be need to be added to the reference configuration.
In my opinion we should have the importer script take the main configuration file, and script modify it to Mbed OS related features, like it does now, otherwise we will encounter numerous maintenance issues.

Patater:
> As Gilles mentioned,
> If we have a reference configuration for Mbed OS, that will be used by the importer script, it means that every new configuration in the main `config.h` file, will be need to be added to the reference configuration.
> In my opinion we should have the importer script take the main configuration file, and script modify it to Mbed OS related features, like it does now, otherwise we will encounter numerous maintenance issues.

What's the best way to ship a differential reference configuration?
- Is it a header file which would be included via `MBEDTLS_USER_CONFIG_FILE`?
- Is it a shell script that modifies the default configuration via a series of `config.py` calls?

I feel like the first option would be more portable and slightly easier to maintain, but interested to hear your thoughts as well.

tests/scripts/all.sh
Patater: Should we also add `-Wstack-usage=2048 -Werror` to ensure we don't overflow the Mbed OS default main thread stack?

We should also use `-Os`, if that's what Mbed OS will default to using.

Patater: Gah, seems we can't enforce that stack limit until we can fit within that stack limit.

```
$ make CFLAGS:="-Os -Wstack-usage=2048 -Werror" check
  CC    bignum.c
bignum.c: In function ‘mbedtls_mpi_exp_mod’:
bignum.c:1926:5: error: stack usage is 3312 bytes [-Werror=stack-usage=]
 int mbedtls_mpi_exp_mod( mbedtls_mpi *X, const mbedtls_mpi *A,
     ^~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
Makefile:150: recipe for target 'bignum.o' failed
make[2]: *** [bignum.o] Error 1
Makefile:158: recipe for target 'libmbedcrypto.a' failed
make[1]: *** [libmbedcrypto.a] Error 2
Makefile:19: recipe for target 'lib' failed
make: *** [lib] Error 2
```
What stack size should we test that we don't exceed and recommend Mbed OS users configure if they want to use Mbed TLS? We appear to fit within 6000 bytes.

```
  Gen   test_suite_version.c
  CC    test_suite_version.c
suites/host_test.function: In function ‘execute_tests’:
suites/host_test.function:494:5: error: stack usage might be 5968 bytes [-Werror=stack-usage=]
 int execute_tests( int argc , const char ** argv )
     ^~~~~~~~~~~~~
```

gilles-peskine-arm: Stack consumption in `mbedtls_mpi_exp_mod` increases exponentially with `MBEDTLS_MPI_WINDOW_SIZE`. It's a space/time compromise, with ~1.5kB difference in stack usage between the min and max value on a 32-bit machine. Maybe we could reduce this value in Mbed OS: currently it uses the best-performance value except on `TARGET_STM32F439xI`.

The library and the test suite are written in a very different style when it comes to stack consumption. `execute_tests` uses a `char buf[5000]` to store a line of the `.data` file. I think we should `-Wstack-usage` only for the library, not for the test suite.

Patater: I agree we should run with `-Wstack-usage` only for the library for this PR.

In the future, we should build the tests as they'd be used on target (so, likely excluding host_test.function): this enables us to catch when on-target testing would run into stack issues. Some test suites allocate unnecessarily large buffers on the stack. We should have a regression test in place to ensure issues like what https://github.com/ARMmbed/mbed-crypto/pull/278 fixes don't rear their heads again, but adding such tests in this PR is out of scope.

mpg:
It seems to me that with Mbed TLS recently becoming more independent from Mbed OS, it's no longer clear if this test should live in the mbedtls tree, especially if that requires having (an equivalent of) the Mbed OS config living in the Mbed TLS tree. @Patater @danh-arm wdyt?

mpg:
Considering this PR does not appear to be close to be ready (still requires design discussions) or recently active, and we moving away from the mbed-crypto repo for new work, I'm closing this PR and encouraging you to re-open a similar one in mbedtls if you're still interested (perhaps after discussing if this is still appropriate now that Mbed TLS is less strongly connected to Mbed OS).

PR #257, opened 2019-09-12
Forbid zero-length keys

### LINKED ISSUE #50: Zero length keys not supported when generating persistent keys
### Description

`psa_save_generated_persistent_key` does not support zero-length keys on all platforms. This has been missed due to not having a test for zero-length raw keys. 

### Issue request type

    [ ] Question
    [ ] Enhancement
    [x] Bug

### LINKED ISSUE #258: psa_asymmetric_verify accepts RSA signatures with trailing garbage
`psa_asymmetric_verify` with an RSA signature algorithm returns a success status not only if the provided signature is correct, but also if the provided signature consists of a correct signature followed by trailing junk.

This is of limited concern since it does not allow forgery attacks: there's no way to leverage this into getting a message accepted if the private key holder didn't sign it. However it violates the expectation that RSA PKCS#1v1.5 signatures are unique (if S1 and S2 are two valid signatures of the same message then S1 is bitwise identical to S2), and the expectation that a valid RSA signature (whatever the padding method) has a given size.

gilles-peskine-arm:
Keys of size 0 generally don't make sense: a key is supposed to be secret. There is one edge case which is "raw data" keys, which are useful to store non-key objects in the same storage location as keys. However those are also problematic because they involve a zero-length buffer. Manipulating zero-length buffers in C requires special cases with functions like malloc() and memcpy(). Additionally, 0 as a key size already has a meaning "unspecified", which does not always overlap seamlessly with the meaning "0".
    
Therefore, forbid keys of size 0. No PSA implementation may accept them. Reject them in Mbed Crypto.

* Fixes #50 (by making it moot).
* Fixes https://github.com/ARMmbed/psa-crypto/issues/168
* Addresses #137, but only for key objects and in `psa_asymmetric_sign`, not for buffers containing other things.
* Fix #258


library/psa_crypto.c
```diff
@@ -451,13 +451,6 @@ static psa_status_t prepare_raw_data_slot( psa_key_type_t type,
     switch( type )
     {
         case PSA_KEY_TYPE_RAW_DATA:
-            if( bits == 0 )
-            {
-                raw->bytes = 0;
-                raw->data = NULL;
-                return( PSA_SUCCESS );
-            }
-            break;
```
Patater: Don't we want to keep `break` here? or is it correct to `/* fallthrough */`?

library/psa_crypto.c
```diff
@@ -3191,8 +3191,8 @@ static psa_status_t psa_rsa_verify( mbedtls_rsa_context *rsa,
     if( status != PSA_SUCCESS )
         return( status );
 
-    if( signature_length < mbedtls_rsa_get_len( rsa ) )
-        return( PSA_ERROR_BUFFER_TOO_SMALL );
```
Patater: In commit "Fix signature size checks in psa_asymmetric_verify for RSA", `If the signature doesn't have the correct signature` is more easily understood when written as `If the signature doesn't have the correct length`, although it is less cute.

Patater:
Is there any code from https://github.com/ARMmbed/mbed-crypto/pull/48/ that we'd like to remove now that we disallow zero-length keys?

library/psa_crypto.c
gilles-peskine-arm: `RAW_DATA`, `HMAC` and `DERIVE` have the same treatment. No fallthrough, just several values with the same handling.

gilles-peskine-arm:
Regarding #48, I didn't look at it specifically, but “Remove special handling for zero-length keys” effectively reverts it as well as removing similar checks in other functions.

library/psa_crypto.c
gilles-peskine-arm: I amended the commit message.

athoelke:
I'm happy with the updates to the documentation..

Patater:
When this is merged, should we also close https://github.com/ARMmbed/mbed-crypto/issues/50 now that it's a feature and no longer a bug?

Patater:
> When this is merged, should we also close #50 now that it's a feature and no longer a bug?

Yes. I see you've address this in your PR description. Steps ahead of me. :)

PR #259, opened 2019-09-13
Check `len` against buffers size upper bound in PSA tests

k-stachowiak:
In `asn1_skip_integer()`, in `test_suite_psa_crypto.function` suite, a buffer is accessed at an index based on `len` - taken from itself. There is a lower bound check done on the `len` value, but none for the upper bound.

This PR adds the upper bound check for the value used for the lookup.


gilles-peskine-arm:
The test failures are unrelated and don't obscure coverage of the changes in this PR. Good to merge.

PR #260, opened 2019-09-13
Update multipart operation documentation

athoelke:
This aligns the API documentation for multipart operations to match the general pattern for these object types. In particular:

* After setup, an error return from other functions results in the object entering an _error_ state. The relevant `psa_xxx_abort` function must be called to reset the object.
* Tidy up the use of `PSA_ERROR_BAD_STATE` to only refer to mismatched object state and function call. invalid objects can be reported using `PSA_ERROR_CORRUPTION_DETECTED`.

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -1318,7 +1318,7 @@ static psa_mac_operation_t psa_mac_operation_init(void);
  * -# Allocate an operation object which will be passed to all the functions
  *    listed here.
  * -# Initialize the operation object with one of the methods described in the
- *    documentation for #psa_mac_operation_t, e.g. PSA_MAC_OPERATION_INIT.
+ *    documentation for #psa_mac_operation_t, e.g. #PSA_MAC_OPERATION_INIT.
  * -# Call psa_mac_sign_setup() to specify the algorithm and key.
  * -# Call psa_mac_update() zero, one or more times, passing a fragment
  *    of the message each time. The MAC that is calculated is the MAC
@@ -1326,13 +1326,15 @@ static psa_mac_operation_t psa_mac_operation_init(void);
  * -# At the end of the message, call psa_mac_sign_finish() to finish
  *    calculating the MAC value and retrieve it.
  *
- * The application may call psa_mac_abort() at any time after the operation
+ * If an error occurs at any step after a call to psa_mac_sign_setup(), the
+ * operation will need to be reset by a call to psa_mac_abort(). The
+ * application may call psa_mac_abort() at any time after the operation
  * has been initialized.
  *
  * After a successful call to psa_mac_sign_setup(), the application must
  * eventually terminate the operation through one of the following methods:
- * - A failed call to psa_mac_update().
- * - A call to psa_mac_sign_finish() or psa_mac_abort().
+ * - A successful call to psa_mac_sign_finish().
+ * - A call to psa_mac_abort().
  *
  * \param[in,out] operation The operation object to set up. It must have
  *                          been initialized as per the documentation for
@@ -1358,8 +1360,7 @@ static psa_mac_operation_t psa_mac_operation_init(void);
  * \retval #PSA_ERROR_STORAGE_FAILURE
  *         The key could not be retrieved from storage.
  * \retval #PSA_ERROR_BAD_STATE
- *         The operation state is not valid (already set up and not
- *         subsequently completed).
+ *         The operation state is not valid (it must be inactive).
  * \retval #PSA_ERROR_BAD_STATE
  *         The library has not been previously initialized by psa_crypto_init().
  *         It is implementation-dependent whether a failure to initialize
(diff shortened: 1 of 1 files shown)


athoelke:
Some of the key derivation operation functions specify specific behaviour for certain error conditions - for example setting the operation capacity to zero. I haven't changed that text yet, should these scenarios:
1. Continue to result in a non-error-state derivation operation with capacity zero, OR
2. Align with the general error flow and result in the operation entering an error state that will return `PSA_ERROR_BAD_STATE` in subsequent calls, OR
3. Be propagated to other error conditions, effectievly making "zero capacity" equivalent to "error state" for these objects (except for the error return code).

My preference is for (2) - as a zero capacity derivation object is as useful as one that is in an *error* state, and unifying the outcome on error makes the API simpler.

include/psa/crypto.h
```diff
@@ -1607,7 +1593,8 @@ psa_status_t psa_mac_abort(psa_mac_operation_t *operation);
 /** Encrypt a message using a symmetric cipher.
  *
  * This function encrypts a message with a random IV (initialization
- * vector).
+ * vector). Use the multipart #psa_cipher_operation_t object to provide
```
gilles-peskine-arm: “the multipart object” doesn't make sense: which multipart object? Maybe something like

> Use the multipart operation interface with a #psa_cipher_operation_t object

include/psa/crypto.h
```diff
@@ -2471,8 +2474,8 @@ psa_status_t psa_aead_set_nonce(psa_aead_operation_t *operation,
  * \retval #PSA_SUCCESS
  *         Success.
  * \retval #PSA_ERROR_BAD_STATE
- *         The operation state is not valid (not set up, already completed,
- *         or psa_aead_update_ad() or psa_aead_update() already called).
+ *         The operation state is not valid (it must be active, but before
```
gilles-peskine-arm: Hiatus in the parentheses. Suggestion:

> (it must be active, and psa_aead_update_ad() and psa_aead_update() must not have been called yet)

include/psa/crypto.h
```diff
@@ -2517,8 +2521,8 @@ psa_status_t psa_aead_set_lengths(psa_aead_operation_t *operation,
  * \retval #PSA_SUCCESS
  *         Success.
  * \retval #PSA_ERROR_BAD_STATE
- *         The operation state is not valid (not set up, nonce not set,
- *         psa_aead_update() already called, or operation already completed).
+ *         The operation state is not valid (it must be active and ready to
```
gilles-peskine-arm: Saying that this function fails when the operation is not ready for this function to be called doesn't explain much.

> (it must be active, have a nonce set, have lengths set if required by the algorithm, and psa_aead_update() must not have been called yet)

include/psa/crypto.h
```diff
@@ -2585,8 +2590,8 @@ psa_status_t psa_aead_update_ad(psa_aead_operation_t *operation,
  * \retval #PSA_SUCCESS
  *         Success.
  * \retval #PSA_ERROR_BAD_STATE
- *         The operation state is not valid (not set up, nonce not set
- *         or already completed).
+ *         The operation state is not valid (it must be active and ready to
```
gilles-peskine-arm: Saying that this function fails when the operation is not ready for this function to be called doesn't explain much.

> (it must be active, have a nonce set, and have lengths set if required by the algorithm)

include/psa/crypto.h
```diff
@@ -2656,8 +2663,8 @@ psa_status_t psa_aead_update(psa_aead_operation_t *operation,
  * \retval #PSA_SUCCESS
  *         Success.
  * \retval #PSA_ERROR_BAD_STATE
- *         The operation state is not valid (not set up, nonce not set,
- *         decryption, or already completed).
+ *         The operation state is not valid (it must be an active aead encrypt
```
gilles-peskine-arm: > (it must be an active decryption operation with a nonce set)

include/psa/crypto.h
```diff
@@ -2721,8 +2730,8 @@ psa_status_t psa_aead_finish(psa_aead_operation_t *operation,
  * \retval #PSA_SUCCESS
  *         Success.
  * \retval #PSA_ERROR_BAD_STATE
- *         The operation state is not valid (not set up, nonce not set,
- *         encryption, or already completed).
+ *         The operation state is not valid (it must be an active aead decrypt
```
gilles-peskine-arm: > (it must be an active decryption operation with a nonce set)

include/psa/crypto.h
```diff
@@ -3338,7 +3357,10 @@ psa_status_t psa_key_derivation_key_agreement(
  * stream.
  * The operation's capacity decreases by the number of bytes read.
  *
- * \param[in,out] operation The key derivation operation object to read from.
+ * If this function returns an error status, the operation enters an error
+ * state and must be aborted by calling psa_key_derivation_abort().
+ *
+* \param[in,out] operation The key derivation operation object to read from.
```
gilles-peskine-arm: Blocker: continuation comment lines must start with exactly one space and a `*`, otherwise this confuses one of our scripts (`tests/scripts/check-names.sh`).

gilles-peskine-arm:
Regarding key derivation operations that have reached their capacity, I think attempting to get output from them should keep returning `INSUFFICIENT_CAPACITY`. What other error cases are not described properly?

include/psa/crypto.h
athoelke: yes, but "encryption"

athoelke:
>Regarding key derivation operations that have reached their capacity, I think attempting to get output from them should keep returning INSUFFICIENT_CAPACITY. What other error cases are not described properly?

So this requires amendments to the general "If this functoin returns an error, then the operation enter an _error_ state and requires you to call `abort`" for the `psa_key_derivation_output_*` functions - to exclude the `INSUFFICIENT_DATA` error status.

Additionally, the first sentence of description of `psa_key_derivation_abort` seems redundant, now that:
- the rest of the text states that the operation object is reset to _inactive_.
- the documentation for the extraction functions states that they will return `BAD_STATE` if the operation is not _active_.

I will post a commit to make those changes.

gilles-peskine-arm:
This only changes the PSA part of the documentation. CI is green except for some random failures in DTLS tests which is a known unrelated issue. Merging.

PR #263, opened 2019-09-17
Introduce ASN.1 SEQUENCE traversal API

hanno-arm:
__Summary:__ This PR is a spin-out from the X.509 on demand parsing work https://github.com/ARMmbed/mbedtls/pull/2478, one of the main drivers of which is to avoid dynamically allocated presentations of ASN.1 SEQUENCEs by traversing the raw ASN.1 in-place whenever needed. To this end, this PR introduces the ASN.1 API `mbedtls_asn1_traverse_sequence_of()` which can be used to traverse an ASN.1 SEQUENCE container, triggering a user-specified callback for each entry found. 
The PR also introduces two helper macros `MBEDTLS_ASN1_IS_STRING_TAG` and `MBEDTLS_OID_CMP_RAW`. Those are technically unrelated to the SEQUENCE traversal API and are included here solely to keep the number of dependencies for the on demand parsing PR low.

### DIFF
--- a/include/mbedtls/asn1.h
+++ b/include/mbedtls/asn1.h
@@ -132,6 +132,10 @@
         ( ( MBEDTLS_OID_SIZE(oid_str) != (oid_buf)->len ) ||                \
           memcmp( (oid_str), (oid_buf)->p, (oid_buf)->len) != 0 )
 
+#define MBEDTLS_OID_CMP_RAW(oid_str, oid_buf, oid_buf_len)              \
+        ( ( MBEDTLS_OID_SIZE(oid_str) != (oid_buf_len) ) ||             \
+          memcmp( (oid_str), (oid_buf), (oid_buf_len) ) != 0 )
+
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/include/mbedtls/asn1.h
+++ b/include/mbedtls/asn1.h
@@ -343,6 +343,9 @@ int mbedtls_asn1_get_bitstring_null( unsigned char **p,
  * \brief       Parses and splits an ASN.1 "SEQUENCE OF <tag>".
  *              Updates the pointer to immediately behind the full sequence tag.
  *
+ * This function allocates memory for the sequence elements. You can free
+ * the allocated memory with mbedtls_asn1_sequence_free().
+ *
  * \note        On error, this function may return a partial list in \p cur.
  *              You must set `cur->next = NULL` before calling this function!
  *              Otherwise it is impossible to distinguish a previously non-null
@@ -384,6 +387,28 @@ int mbedtls_asn1_get_sequence_of( unsigned char **p,
                                   const unsigned char *end,
                                   mbedtls_asn1_sequence *cur,
                                   int tag );
+/**
+ * \brief          Free a heap-allocated linked list presentation of
+ *                 an ASN.1 sequence, including the first element.
+ *
+ * There are two common ways to manage the memory used for the representation
+ * of a parsed ASN.1 sequence:
+ * - Allocate a head node `mbedtls_asn1_sequence *head` with mbedtls_calloc().
+ *   Pass this node as the `cur` argument to mbedtls_asn1_get_sequence_of().
+ *   When you have finished processing the sequence,
+ *   call mbedtls_asn1_sequence_free() on `head`.
+ * - Allocate a head node `mbedtls_asn1_sequence *head` in any manner,
+ *   for example on the stack. Make sure that `head->next == NULL`.
+ *   Pass `head` as the `cur` argument to mbedtls_asn1_get_sequence_of().
+ *   When you have finished processing the sequence,
+ *   call mbedtls_asn1_sequence_free() on `head->cur`,
+ *   then free `head` itself in the appropriate manner.
+ *
+ * \param seq      The address of the first sequence component. This may
+ *                 be \c NULL, in which case this functions returns
+ *                 immediately.
+ */
+void mbedtls_asn1_sequence_free( mbedtls_asn1_sequence *seq );
 
 #if defined(MBEDTLS_BIGNUM_C)
 /**
--- a/include/mbedtls/asn1.h
+++ b/include/mbedtls/asn1.h
@@ -90,6 +90,18 @@
 #define MBEDTLS_ASN1_CONSTRUCTED             0x20
 #define MBEDTLS_ASN1_CONTEXT_SPECIFIC        0x80
 
+/* Slightly smaller way to check if tag is a string tag
+ * compared to canonical implementation. */
+#define MBEDTLS_ASN1_IS_STRING_TAG( tag )                                     \
+    ( ( tag ) < 32u && (                                                      \
+        ( ( 1u << ( tag ) ) & ( ( 1u << MBEDTLS_ASN1_BMP_STRING )       |     \
+                                ( 1u << MBEDTLS_ASN1_UTF8_STRING )      |     \
+                                ( 1u << MBEDTLS_ASN1_T61_STRING )       |     \
+                                ( 1u << MBEDTLS_ASN1_IA5_STRING )       |     \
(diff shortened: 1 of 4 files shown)


gilles-peskine-arm:
https://github.com/ARMmbed/mbed-crypto/pull/75 fixes some edge cases in the ASN.1 library and adds tests. Can you please make sure that your X.509 code and tests are compatible with these fixes? This PR doesn't have tests; I'm willing to accept it without tests but the lack of tests needs to be filed as a separate bug.

hanno-arm:
@gilles-peskine-arm It's not fully satisfactory, but because the PR reimplements `mbedtls_asn1_get_sequence_of()` in terms of the new traversal API, any tests for the former will also test the latter.

include/mbedtls/asn1.h
```diff
@@ -316,6 +316,58 @@ int mbedtls_asn1_get_sequence_of( unsigned char **p,
                           mbedtls_asn1_sequence *cur,
                           int tag);
 
+/**
+ * \brief                Traverse an ASN.1 SEQUENCE container and
+ *                       call a callback for each entry.
+ *
+ * \warning              This function is still experimental and may change
+ *                       at any time.
```
Patater: Why should this be experimental? When can it be non-experimental? Could we make asn1_internal.h and place it there if it's only for internal use for now?

include/mbedtls/asn1.h
```diff
@@ -316,6 +316,58 @@ int mbedtls_asn1_get_sequence_of( unsigned char **p,
...
+ * \param ctx            The context to be passed to the callback \p cb.
+ *
+ * \return               \c 0 if successful the entire ASN.1 SEQUENCE
+ *                       was traversed without parsing or callback errors.
+ * \return               A negative ASN.1 error code on a parsing failure.
+ * \return               A non-zero error code forwarded from the callback
+ *                       \p cb in case the latter returns a non-zero value.
+ */
+int mbedtls_asn1_traverse_sequence_of(
+    unsigned char **p,
```
Patater: Can the `unsigned char` be `const unsigned char`? I don't think we'd be modifying `**p`, only `p` or `*p`.

include/mbedtls/asn1.h
```diff
@@ -316,6 +316,58 @@ int mbedtls_asn1_get_sequence_of( unsigned char **p,
...
+    const unsigned char *end,
+    uint8_t tag_must_mask, uint8_t tag_must_val,
+    uint8_t tag_may_mask, uint8_t tag_may_val,
+    int (*cb)( void *ctx, int tag,
+               unsigned char* start, size_t len ),
```
Patater: Can `start` be made const?

library/asn1parse.c
```diff
@@ -292,59 +292,56 @@ void mbedtls_asn1_sequence_free( mbedtls_asn1_sequence *seq )
...
+    cur->buf.tag = tag;
 
+    cb_ctx->cur = cur;
     return( 0 );
 }
 
+/*
+ *  Parses and splits an ASN.1 "SEQUENCE OF <tag>"
+ */
+int mbedtls_asn1_get_sequence_of( unsigned char **p,
```
Patater: Ah, too bad this is not const. I guess we can't have our callback with `p` const because of this.

gilles-peskine-arm:
I rebased this pull request on top of the current base branch. The previous content by Hanno is at https://github.com/gilles-peskine-arm/mbed-crypto/tree/asn1_traversal_api-1.

Compared with the previous content:

* I removed most updates to existing documentation because they were largely superseded by #75.
* I added tests for the new functions.
* I tweaked the documentation of the new functions.

include/mbedtls/asn1.h
gilles-peskine-arm: It isn't internal since it will be used by Mbed TLS.

include/mbedtls/asn1.h
gilles-peskine-arm: `const unsigned char **` and `unsigned char **` do not mix. If this function used a different type, it couldn't be used with the other asn1 parsing functions.

include/mbedtls/asn1.h
gilles-peskine-arm: I'm wondering about that. In 99% of the cases it would be const: parsing a read-only string. But it may be useful occasionally to modify an input in place.

gilles-peskine-arm:
New CI run now that the problems with the Cypress boards are fixed: https://jenkins-internal.mbed.com/job/mbed-crypto-pr/job/PR-263-merge/20/

include/mbedtls/asn1.h
hanno-arm: @gilles-peskine-arm @Patater What's your view on this? This function must be public, as @gilles-peskine-arm said, so what remains is to decide whether we want to declare it as experimental or not. I'm open either way. If you think the API is good as-is, I'm happy to remove the qualification.

gilles-peskine-arm: My preference is what you wrote: declare this function as experimental. We don't care about this function except as an internal detail of the implementation an X509 feature that we do care about. It has a complex API that I would prefer not to make official, but that I'm ok to include with the understanding that we're likely to modify it in the future.

include/mbedtls/asn1.h
hanno-arm: I agree with @Patater in that in principle the type of `p` should be `const unsigned char **` here. However, the rest of the ASN.1 parsing library unfortunately also uses `unsigned char **`, which we can't change and which forces the use of `unsigned char **` here, too (not just for consistency, but simply because this function calls e.g. `mbedtls_asn1_get_tag()` which has a `unsigned char **` argument). An alternative would be to extend the documentation of all ASN.1 parsing functions to explicitly mention that, despite the missing `const` qualification, the `**p` is never written to, and that it is hence safe to use the functions in a `const unsigned char **` context. If we did that, we could start improving the API by using `const unsigned char **` at least for this new API function, and doing a const-cast when calling the other parsing functions.

gilles-peskine-arm: We can tweak the API in Mbed TLS 3. Until we get there, let's keep the API consistent.

include/mbedtls/asn1.h
hanno-arm: I agree that it should be `const`. However, if we want to call other ASN.1 parsing functions from within the callback, then we can't have `const` here because those parsing functions don't have `const` qualification, as mentioned above.

include/mbedtls/asn1.h
hanno-arm: Ok, no change here, then.

library/asn1parse.c
hanno-arm: This way it would work, but what doesn't work is calling parsing functions from within `mbedtls_asn1_traverse_sequence_of()`.

hanno-arm:
Rebase on top of `development`.

mpg:
The CI found a problem, in `test-ref-configs`, with `config-symmetric-only.h`, the first error is:
```
/var/lib/build/include/mbedtls/asn1.h:504:5: error: unknown type name 'uint8_t'
     uint8_t tag_must_mask, uint8_t tag_must_val,
     ^
```
and the other errors look similar. I suggest including `stdint.h` directly in `asn1.h` - it probably happens to be included indirectly in some configurations including the default one, but apparently not all of them.

hanno-arm:
Thank you @mpg, I'll fix that.

hanno-arm:
@mpg @gilles-peskine-arm Fixed the issue by replacing `uint8_t` by `unsigned char`.

include/mbedtls/asn1.h
mpg: The situation should be temporary, since I think we're aiming at making the whole `asn1.h` private in Mbed TLS 3.0 anyway.

gilles-peskine-arm:
CI passes except for the Mbed OS tests which are failing for unrelated reasons.

PR #264, opened 2019-09-17
Test the library when malloc(0) returns NULL

### LINKED ISSUE #137: different behavior in `calloc` on different libraries when trying to allocate 0 bytes

gilles-peskine-arm:
On our main test platforms (Linux, FreeBSD), malloc(0) returns a non-NULL pointer. With `MBEDTLS_MEMORY_BUFFER_ALLOC_C`, `calloc(0,_)` does return NULL, but memory_buffer_alloc doesn't play well with UBSan. Add a test component with UBSan where `calloc` returns NULL when asked to allocate 0 bytes, wrapping around the platform `calloc`.

Fix #137: now it's tested.

Should be backported to mbedtls LTS branches.

### DIFF
--- a/programs/test/selftest.c
+++ b/programs/test/selftest.c
@@ -65,6 +65,8 @@
 #else
 #include <stdio.h>
 #include <stdlib.h>
+#define mbedtls_calloc     calloc
+#define mbedtls_free       free
 #define mbedtls_printf     printf
 #define mbedtls_snprintf   snprintf
 #define mbedtls_exit       exit
@@ -76,6 +78,86 @@
 #include "mbedtls/memory_buffer_alloc.h"
 #endif
 
+
+#if defined MBEDTLS_SELF_TEST
+/* Sanity check for malloc. This is not expected to fail, and is rather
+ * intended to display potentially useful information about the platform,
+ * in particular the behavior of malloc(0). */
+static int calloc_self_test( int verbose )
+{
+    int failures = 0;
+    void *empty1 = mbedtls_calloc( 0, 1 );
+    void *empty2 = mbedtls_calloc( 0, 1 );
+    void *buffer1 = mbedtls_calloc( 1, 1 );
+    void *buffer2 = mbedtls_calloc( 1, 1 );
+    uintptr_t old_buffer1;
+
+    if( empty1 == NULL && empty2 == NULL )
+    {
+        if( verbose )
+            mbedtls_printf( "  CALLOC(0): passed (NULL)\n" );
+    }
+    else if( empty1 == NULL || empty2 == NULL )
+    {
+        if( verbose )
+            mbedtls_printf( "  CALLOC(0): failed (mix of NULL and non-NULL)\n" );
+        ++failures;
+    }
+    else if( empty1 == empty2 )
+    {
+        if( verbose )
+            mbedtls_printf( "  CALLOC(0): passed (same non-null)\n" );
+    }
+    else
+    {
+        if( verbose )
+            mbedtls_printf( "  CALLOC(0): passed (distinct non-null)\n" );
+    }
+
+    if( buffer1 == NULL || buffer2 == NULL )
+    {
+        if( verbose )
+            mbedtls_printf( "  CALLOC(1): failed (NULL)\n" );
+        ++failures;
+    }
+    else if( buffer1 == buffer2 )
+    {
+        if( verbose )
+            mbedtls_printf( "  CALLOC(1): failed (same buffer twice)\n" );
+        ++failures;
+    }
+    else
+    {
+        if( verbose )
+            mbedtls_printf( "  CALLOC(1): passed\n" );
+    }
+
+    old_buffer1 = (uintptr_t) buffer1;
+    mbedtls_free( buffer1 );
+    buffer1 = mbedtls_calloc( 1, 1 );
+    if( buffer1 == NULL )
+    {
+        if( verbose )
+            mbedtls_printf( "  CALLOC(1 again): failed (NULL)\n" );
+        ++failures;
+    }
+    else
+    {
+        if( verbose )
+            mbedtls_printf( "  CALLOC(1 again): passed (%s address)\n",
+                            (uintptr_t) old_buffer1 == (uintptr_t) buffer1 ?
+                            "same" : "different" );
+    }
+
+    if( verbose )
+        mbedtls_printf( "\n" );
+    mbedtls_free( empty1 );
+    mbedtls_free( empty2 );
+    mbedtls_free( buffer1 );
+    mbedtls_free( buffer2 );
+    return( failures );
+}
+#endif /* MBEDTLS_SELF_TEST */
 
 static int test_snprintf( size_t n, const char ref_buf[10], int ref_ret )
 {
@@ -173,6 +255,7 @@ typedef struct
 
 const selftest_t selftests[] =
 {
+    {"calloc", calloc_self_test},
 #if defined(MBEDTLS_MD2_C)
     {"md2", mbedtls_md2_self_test},
 #endif
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -822,6 +822,21 @@ component_test_platform_calloc_macro () {
     make test
 }
 
+component_test_malloc_0_null () {
+    msg "build: malloc(0) returns NULL (ASan+UBSan build)"
+    scripts/config.pl full
+    scripts/config.pl unset MBEDTLS_MEMORY_BUFFER_ALLOC_C
+    make CC=gcc CFLAGS="'-DMBEDTLS_CONFIG_FILE=\"$PWD/tests/configs/config-wrapper-malloc-0-null.h\"' -O -Werror -Wall -Wextra -fsanitize=address,undefined" LDFLAGS='-fsanitize=address,undefined'
+
+    msg "test: malloc(0) returns NULL (ASan+UBSan build)"
+    make test
+
+    msg "selftest: malloc(0) returns NULL (ASan+UBSan build)"
+    # Just the calloc selftest. "make test" ran the others as part of the
+    # test suites.
+    if_build_succeeded programs/test/selftest calloc
+}
+
 component_test_aes_fewer_tables () {
     msg "build: default config with AES_FEWER_TABLES enabled"
     scripts/config.pl set MBEDTLS_AES_FEWER_TABLES
--- a/tests/configs/config-wrapper-malloc-0-null.h
+++ b/tests/configs/config-wrapper-malloc-0-null.h
@@ -1,0 +1,39 @@
+/* config.h wrapper that forces calloc(0) to return NULL.
+ * Used for testing.
+ */
+/*
+ *  Copyright (C) 2019, ARM Limited, All Rights Reserved
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *  not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  This file is part of mbed TLS (https://tls.mbed.org)
+ */
+
+#ifndef MBEDTLS_CONFIG_H
+/* Don't #define MBEDTLS_CONFIG_H, let config.h do it. */
+
+#include "mbedtls/config.h"
+
+#include <stdlib.h>
+static inline void *custom_calloc( size_t nmemb, size_t size )
+{
+    if( nmemb == 0 || size == 0 )
+        return( NULL );
+    return( calloc( nmemb, size ) );
+}
+
+#define MBEDTLS_PLATFORM_MEMORY
+#define MBEDTLS_PLATFORM_STD_CALLOC custom_calloc
+
+#endif /* MBEDTLS_CONFIG_H */


Patater:
#257 is merged. Restarting CI.

programs/test/selftest.c
```diff
@@ -77,6 +79,103 @@
...
+    }
+
+    if( verbose )
+        mbedtls_printf( "\n" );
+    mbedtls_free( empty1 );
+    mbedtls_free( empty2 );
+    mbedtls_free( buffer1 );
+    mbedtls_free( buffer2 );
+    mbedtls_free( overflow );
+    return( failures );
```
Patater: What sorts of platforms would this fail on? Why are we checking exactly? What should we or our users do if we find a platform where this self-test fails? (Implement a calloc wrapper to make it work?)

> The main objective is to report whether calloc returns NULL for a size
of 0.

Yes, but so what? What is to be done with this knowledge?

tests/config-wrapper-malloc-0-null.h
```diff
@@ -0,0 +1,39 @@
+/* config.h wrapper that forces calloc(0) to return NULL.
```
Patater: In commit message for "Add a test component with malloc(0) returning NULL".

> `Make this a separte job with UBSan (and ASan) to detect potential.`

s/separte/separate

and is there supposed to be a word after "potential" in this sentence?

tests/config-wrapper-malloc-0-null.h
```diff
@@ -0,0 +1,39 @@
...
+ * Used for testing.
```
Patater: Why not put this file in `configs/` or, if that's not acceptable, `tests/configs/`? Having it at the `tests/` root is a bit messy, IMO.

tests/config-wrapper-malloc-0-null.h
```diff
@@ -0,0 +1,39 @@
...
+ *  This file is part of mbed TLS (https://tls.mbed.org)
+ */
+
+#ifndef MBEDTLS_CONFIG_H
+/* Don't #define MBEDTLS_CONFIG_H, let config.h do it. */
+
+#include "mbedtls/config.h"
+
+#include <stdlib.h>
+static inline void *custom_calloc( size_t nmemb, size_t size )
```
Patater: This file implements a custom calloc. However, the commit message says "malloc". The all.sh test also says "malloc".

gilles-peskine-arm:
CI reported failures. I'll fix these and also rebase on top of `development` now that the #257 has been merged.

programs/test/selftest.c
gilles-peskine-arm: It's a self-test. Self-tests don't check much that's useful. We don't expect the AES self-test to fail either — if it did, the unit tests would fail. The reason I added it here was to confirm that the test job that's supposed to test `calloc(0)=NULL` indeed has `calloc(0)=NULL`.

If the test fails, you know your `calloc` is broken. You know your platform better than we do, so we can't help you to fix it.

tests/config-wrapper-malloc-0-null.h
gilles-peskine-arm: This is not a maintained configuration, or a meaningful configuration for users. So it should be under `tests`. I can move it to a subdirectory if you prefer but I don't think this makes `tests` any less messy.

tests/config-wrapper-malloc-0-null.h
gilles-peskine-arm: `calloc` is in all but standardese a wrapper around `malloc`.

Patater: Looks a bit weird that it doesn't match names.

tests/config-wrapper-malloc-0-null.h
Patater: Subdirectory would be good.

Scope creep: Do we have any other configs in the `configs` folder that are really just for testing and wouldn't really be meaningful configurations for users?

programs/test/selftest.c
Patater: What if we get `CALLOC(0): passed (distinct non-null)`? That is a valid and passing self-test result, isn't it? Does our other test fail to test what we want when that happens?

tests/config-wrapper-malloc-0-null.h
gilles-peskine-arm: Fixed in rebase.

tests/config-wrapper-malloc-0-null.h
gilles-peskine-arm: The other configurations are useful for users. I'll create a `test/configs`.

tests/config-wrapper-malloc-0-null.h
gilles-peskine-arm: I changed the commit message, but kept some uses of malloc because “malloc” is a likely search term here.

programs/test/selftest.c
gilles-peskine-arm: `CALLOC(0): passed (distinct non-null)` is what you get on most if not all the platforms we test on today. I don't understand your question. The goal of this PR is to test on a platform with `CALLOC(0): passed (distinct non-null)`.

I didn't add a test run for `CALLOC(0): passed (same non-null)` because I don't think it matters to our code, whereas null vs non-null does.

Patater: > The reason I added it here was to confirm that the test job that's supposed to test calloc(0)=NULL indeed has calloc(0)=NULL.

What if the platform doesn't have `calloc(0)=NULL`? Is the test job that's supposed to test `calloc(0)=NULL` invalid or shouldn't be run?

programs/test/selftest.c
```diff
@@ -77,6 +79,105 @@
...
+{
+    int failures = 0;
+    void *empty1 = mbedtls_calloc( 0, 1 );
+    void *empty2 = mbedtls_calloc( 0, 1 );
+    void *buffer1 = mbedtls_calloc( 1, 1 );
+    void *buffer2 = mbedtls_calloc( 1, 1 );
+    /* Attempt to allocate a size that's larger than size_t. This
+     * should fail. With AddressSanitizer, this crashes the program by
+     * default; to avoid this, set the environment variable
+     * ASAN_OPTIONS=allocator_may_return_null=1 when running this program.
```
gilles-peskine-arm: On the one hand, I don't like having the self-test program fail when run with ASan.

On the other hand, this is the one aspect of this self-test that might be plausibly broken on an embedded platform with a custom `calloc`, and it wouldn't cause functional bugs but could be a major security issue (e.g. convince the memory allocator that you want to allocate 4 bytes, but make mbedtls code think it's allocated SIZE_MAX+4 bytes).

programs/test/selftest.c
gilles-peskine-arm: I don't understand the question. The platform may or may not return NULL for malloc(0): both are valid. The objective of this PR is to make sure that we test the library on a platform where malloc(0) returns NULL. Since we only use platforms where malloc(0) is non-null, I simulate a platform where malloc(0) returns NULL by defining a wrapper function for `calloc` (we only use `calloc`, never `malloc`) that forcibly returns NULL when told to allocate 0 bytes. So now there is one component of `all.sh` that runs the unit tests with `calloc` returning NULL for 0 bytes. This component also runs the selftest and I confirm that I didn't mess up the way I set up the wrapper by visually inspecting that the selftest prints `NULL`. I didn't test that in `all.sh` because it's not worth the trouble to do it with our current tooling.

programs/test/selftest.c
gilles-peskine-arm: And in fact GCC's warning covers the case where the product of the parameters to `calloc` overflows. So if we perform this test, we need to do it without using values that the compiler can see. Since this is out of scope of this PR, I'll just remove this test.

tests/config-wrapper-malloc-0-null.h
gilles-peskine-arm: Done in force-push.

gilles-peskine-arm:
The remaining CI failures are unrelated to this PR (random DTLS failures, which is a known thing, and "export keys functionality" in `ssl-opt.sh` which is unrelated but I wouldn't have expected to fail here since the crypto CI uses the latest mbedtls).

programs/test/selftest.c
Patater: > make sure that we test the library on a platform where malloc(0) returns NULL.

This is checked only manually? or is there a test that will fail if the platform's malloc(0) doesn't return NULL?

gilles-peskine-arm: Yes, I checked this manually. As a rule we don't have meta-tests that test that the tests are testing what we think they're testing, so I didn't do it here. It's like checking that a non-regression test fails once, and not keeping a CI run with a buggy product to check that it keeps catching the bug as the test code evolves.

tests/scripts/all.sh
```diff
@@ -822,6 +822,22 @@ component_test_platform_calloc_macro () {
     make test
 }
 
+component_test_malloc_0_null () {
+    msg "build: malloc(0) returns NULL (ASan+UBSan build)"
+    scripts/config.pl full
+    scripts/config.pl unset MBEDTLS_MEMORY_BUFFER_ALLOC_C
```
AndrzejKurek: Since https://github.com/ARMmbed/mbedtls/pull/2469 you don't have to do that.

gilles-peskine-arm: I know, but that's only in mbedtls. The change isn't in mbed-crypto yet and this PR is for mbed-crypto.

AndrzejKurek: Right, sorry!

tests/scripts/all.sh
```diff
@@ -822,6 +822,22 @@ component_test_platform_calloc_macro () {
...
+    make CC=gcc CFLAGS="'-DMBEDTLS_CONFIG_FILE=\"$PWD/tests/configs/config-wrapper-malloc-0-null.h\"' -O -Werror -Wall -Wextra -fsanitize=address,undefined" LDFLAGS='-fsanitize=address,undefined'
+
+    msg "test: malloc(0) returns NULL (ASan+UBSan build)"
+    make test
+
+    msg "selftest: malloc(0) returns NULL (ASan+UBSan build)"
+    # By default ASan terminates the process if called with a size that's
+    # insanely large. We do this deliberately in selftest, so disable this
+    # behavior.
```
Patater: We don't call calloc with a size that is extremely large anymore, so this comment needs updating.

programs/test/selftest.c
Patater: I wasn't requesting any meta tests. I was looking to ensure we weren't adding tests that had specific platform requirements and were running them on platforms that don't meet the requirements, whether self tests, test suites or otherwise.

I understand now what we are doing with the information from the self test, and that it is manual and not automatic. Thanks for explaining.

tests/scripts/all.sh
gilles-peskine-arm: I amended the offending commit (which was the last one).

tests/scripts/all.sh
```diff
@@ -822,6 +822,21 @@ component_test_platform_calloc_macro () {
...
+    scripts/config.pl full
+    scripts/config.pl unset MBEDTLS_MEMORY_BUFFER_ALLOC_C
+    make CC=gcc CFLAGS="'-DMBEDTLS_CONFIG_FILE=\"$PWD/tests/configs/config-wrapper-malloc-0-null.h\"' -O -Werror -Wall -Wextra -fsanitize=address,undefined" LDFLAGS='-fsanitize=address,undefined'
+
+    msg "test: malloc(0) returns NULL (ASan+UBSan build)"
+    make test
+
+    msg "selftest: malloc(0) returns NULL (ASan+UBSan build)"
+    # Just the calloc selftest. "make test" ran the others as part of the
+    # test suites.
```
Patater: How does "make test" avoid running the "calloc" selftest? I'm questioning if we need to call `make test`. I'm also questioning if we need to call calloc selftest separately if we decide we still want to run the whole `make test`

gilles-peskine-arm: Each test suite includes a call to the corresponding selftest function. This is why we don't bother to run `selftest` much in `all.sh`, just once to verify that the program works. This only misses the parts of `selftest.c` that aren't calling selftest functions from the various library modules.

We do need `make test` because the goal is to check that there isn't any null dereference (or similar undefined behavior) if `malloc(0)` returns null. This is very unlikely to come up in a crypto selftest. The places where it's likely to come up is in edge cases (empty input buffer, error recovery code).

Patater: OK, so `make test` is not calling the calloc self test, because we have no test suites that call the calloc self test.

Patater:
CI failures are export keys, to be fixed by other PRs.

PR #265, opened 2019-09-18
Add size asserts to psa_crypto test suite

AndrzejKurek:
Add a signature size check in the psa test suites.

### DIFF
--- a/tests/suites/test_suite_psa_crypto.function
+++ b/tests/suites/test_suite_psa_crypto.function
@@ -3779,6 +3779,8 @@ void sign_fail( int key_type_arg, data_t *key_data,
     size_t signature_length = 0xdeadbeef;
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
 
+    TEST_ASSERT( signature_size != 0 );
+    TEST_ASSERT( signature_size <= PSA_ASYMMETRIC_SIGNATURE_MAX_SIZE );
     ASSERT_ALLOC( signature, signature_size );
 
     PSA_ASSERT( psa_crypto_init( ) );


tests/suites/test_suite_psa_crypto.function
```diff
@@ -3668,6 +3668,7 @@ void aead_decrypt( int key_type_arg, data_t *key_data,
     if( expected_result != PSA_ERROR_INVALID_ARGUMENT )
         TEST_EQUAL( output_size,
                     PSA_AEAD_DECRYPT_OUTPUT_SIZE( alg, input_data->len ) );
+    TEST_ASSERT( output_size > 0 );
```
gilles-peskine-arm: No. If we're decrypting an empty plaintext, `output_size` should be 0, and `output_data` will be `NULL` as desired.

AndrzejKurek: Ok, thanks for clarifying.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -3779,6 +3779,8 @@ void sign_fail( int key_type_arg, data_t *key_data,
     size_t signature_length = 0xdeadbeef;
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
 
+    TEST_ASSERT( signature_size != 0 );
```
gilles-peskine-arm: Why these assertions? If we wanted to test that `psa_asymmetric_sign` behaves properly when given an output buffer with a bad size, this test function should be the right thing to call. If something fails with `signature_size == 0` or `signature_size > PSA_ASYMMETRIC_SIGNATURE_MAX_SIZE`, the something in question should be fixed.

AndrzejKurek: This is due to coverity complaining on the unknown signature size. I added the same check that is made in other tests, to unify it.

gilles-peskine-arm: Ok, but in this particular test, `signature_size` is legitimate input. If Coverity is complaining about something, we need to look into why it's complaining, not silence it by pretending this can't happen.

AndrzejKurek:
Closing, since the Coverity issue turned out to be a false positive.

PR #268, opened 2019-09-19
EC keys written with as DER could be incorrectly formatted

joelpalsson74:
When writing EC keys using mbedtls_pk_write_key_der() the private
part of the key got incorrectly pre-pended with 0 if MSB of the
key was 1.

See issue:
https://github.com/ARMmbed/mbedtls/issues/1028

### DIFF
--- a/include/mbedtls/asn1write.h
+++ b/include/mbedtls/asn1write.h
@@ -106,6 +106,23 @@ int mbedtls_asn1_write_raw_buffer( unsigned char **p, unsigned char *start,
  */
 int mbedtls_asn1_write_mpi( unsigned char **p, unsigned char *start,
                             const mbedtls_mpi *X );
+
+/**
+ * \brief           Write a arbitrary-precision number as an
+ *                  octet string (#MBEDTLS_ASN1_OCTET_STRING)
+ *                  in ASN.1 format.
+ *
+ * \note            This function works backwards in data buffer.
+ *
+ * \param p         The reference to the current position pointer.
+ * \param start     The start of the buffer, for bounds-checking.
+ * \param X         The MPI to write.
+ *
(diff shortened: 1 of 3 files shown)


joelpalsson74:
Gilles, thanks for you comments.
Unfortunately I got no idea how to make the unit tests.

/Joel

joelpalsson74:
Gilles,
It seems as one of the unit test fails:
Failed test suite: MBEDTLS_ECP_DP_SECP192R1_ENABLED

I don't understand how to check if the test is correct or not.
Can you please help out?

Best regards,
Joel

gilles-peskine-arm:
So… this is embarrassing… it turns out that we had already fixed that bug, but the fix slipped through the net. It was applied to the long-time support branches, but it was never released in the development branch of Mbed TLS. We missed this and one other bug fix when we split the crypto code between Mbed TLS and Mbed Crypto.

In the interest of simplifying our maintenance, we're going to apply the same patch to Mbed Crypto that we made to the Mbed TLS LTS branches: https://github.com/ARMmbed/mbed-crypto/pull/314 . It fixes ARMmbed/mbedtls#1028, but does not add `mbedtls_asn1_write_mpi_to_octet_string`.

Do you have a need for `mbedtls_asn1_write_mpi_to_octet_string` beyond fixing pkwrite on EC keys? If so, please rebase this pull request on top of the new code after https://github.com/ARMmbed/mbed-crypto/pull/314 is merged. If not, I think this pull request can be closed.

Please accept our apologies for the inconvenience.

joelpalsson74:
Gilles,
Thanks for the information.
We have no other use case for mbedtls_asn1_write_mpi_to_octet_string(), so this pull request can be closed.

PR #269, opened 2019-09-19
Add API versioning

adrianlshaw:
This adds the implemented specification version to the header file. This fixes ARMmbed/psa-crypto#175. The macros follows the general convention of the PSA Dev APIs.


gilles-peskine-arm:
PSA header changes only. so ok to merge with a single mbedtls team approval. CI is ok except for some TLS tests that failed due to an unrelated issue.

PR #274, opened 2019-09-21
Perform check before initialization

irwir:
The issue https://github.com/ARMmbed/mbedtls/issues/1869 became stale, and was closed.
Yet, one of the mentioned checks remained intact, but it should be made earlier.

### DIFF
--- a/library/ecp.c
+++ b/library/ecp.c
@@ -1945,12 +1945,12 @@ static int ecp_comb_recode_scalar( const mbedtls_ecp_group *grp,
     int ret;
     mbedtls_mpi M, mm;
 
+    /* N is always odd (see above), just make extra sure */
+    if( mbedtls_mpi_get_bit( &grp->N, 0 ) == 0 )
+        return( MBEDTLS_ERR_ECP_BAD_INPUT_DATA );
+
     mbedtls_mpi_init( &M );
     mbedtls_mpi_init( &mm );
-
-    /* N is always odd (see above), just make extra sure */
-    if( mbedtls_mpi_get_bit( &grp->N, 0 ) != 1 )
-        return( MBEDTLS_ERR_ECP_BAD_INPUT_DATA );
 
     /* do we need the parity trick? */
     *parity_trick = ( mbedtls_mpi_get_bit( m, 0 ) == 0 );


irwir:
Should the availability of test results to be raised as an issue?
There is no way to see logs neither for success, nor for failure, because connection invariably times out.

library/ecp.c
```diff
@@ -1945,13 +1945,13 @@ static int ecp_comb_recode_scalar( const mbedtls_ecp_group *grp,
...
 
-    mbedtls_mpi_init( &M );
-    mbedtls_mpi_init( &mm );
-
     /* N is always odd (see above), just make extra sure */
-    if( mbedtls_mpi_get_bit( &grp->N, 0 ) != 1 )
+    if( mbedtls_mpi_get_bit( &grp->N, 0 ) == 0 )
         return( MBEDTLS_ERR_ECP_BAD_INPUT_DATA );
 
+    mbedtls_mpi_init( &M );
```
gilles-peskine-arm: I don't understand why the initialization of `M` and `mm` was moved. It makes the error case very slightly faster, but why do we care about the error case?

In general we prefer a style where objects are initialized as close as possible to the beginning of the function. That makes it easier to check that cleanup code never runs on uninitialized objects.

irwir: > I don't understand why the initialization of `M` and `mm` was moved.

As the title says, the check was moved up because it does input parameter validation and early return in case of error.
Validation checks typically are performed before any other useful actions (in C++ it might be placed even before declarations).

> In general we prefer a style where objects are initialized as close as possible to the beginning of the function. That makes it easier to check that cleanup code never runs on uninitialized objects.

In that same general sense, cleanup code must be executed before the return, but it is skipped because as an obvious waste of time.
Similarly, in case of bad parameters. initialization would be also a waste of resources.
For example, `mbedtls_chacha20_crypt` uses a handful of `*_VALIDATE()` macros for parameter checking _before_ calling initialization function.

irwir:
> The test failures on TLS jobs are an unrelated issue which we're working on fixing.

Thanks for the information.

library/ecp.c
mpg: Honestly I think both versions are fine in this case. There are pros and cons for each. In terms of readability, I like seeing the call to `mbedtls_mpi_init()` close to the declarations (which would mean not moving them), but OTOH as you point out since this check skips the cleanup code by symmetry it can also skip init.

Overall I think I'd still prefer not to move the init code. (I'd also like to point out that the error path is never taken, the check is really there only to make sure we remember about that assumption if we ever add short weierstrass curves with non-prime order in the future, which seems very unlikely at this point.)

irwir:
> I disagree with moving the initialization away from the declaration. I can't think of a good reason to favor performance in the error case over ease of maintenance.

The current code relies on implicit and undocumented assumption that _mbedtls_mpi_init()_ does nothing that makes necessary an explicit cleanup, and never will do. Like memory allocation, for example.
Hence the mentioned _ease of maintenance_ requires that cleanup code must be executed if initialization was perfomed.
The reason to move initialization might be this: to skip initialization correctly.

gilles-peskine-arm:
> The current code relies on implicit and undocumented assumption that mbedtls_mpi_init() does nothing that makes necessary an explicit cleanup, and never will do.

Ok, this is a good reason.

We do rely on this assumption in many other places, and I'd prefer to guarantee it and document it, so that we can minimize the lifetime of uninitialize objects. But it's also ok to avoid relying on it. This change makes the code more fragile to local changes, but less fragile to changes in another module. I don't have strong feelings on this specific case. @yanesca, @mpg how do you see this?

irwir:
> This change makes the code more fragile to local changes

Maybe not.
Sanity checks are commonly performed in the very first lines of code (unfortunately, in C this cannot be placed before declarations).
For example, should the initialization routine require parameters with dereferenced pointers, checks for null would be a necessity.

mpg:
> mpg how do you see this?

I previously stated that I though both versions were fine, with a slight inclination towards keeping the code as it is. But I think @irwir 's point that the current code relies on an undocumented assumption is a pretty strong one, so I now prefer moving the init calls, as done by this PR.

> Sanity checks are commonly performed in the very first lines of code (unfortunately, in C this cannot be placed before declarations).

Well, with C99 it can. And unless I'm mistaken, we are using C99 now in the development branch. So perhaps we could have the best of both worlds by ordering things like this: input check, variable declarations, calls to init. Wdyt?

gilles-peskine-arm:
> we are using C99 now in the development branch

In theory yes, but in practice our CI still hasn't fully caught up (IIRC IAR and VS _still_ don't accept declarations after statements with the settings we're using on the CI).

irwir:
Thanks for merging.
Would it be correct to assume that similar changes will be made in `mbedtls` repository too?

> In theory yes, but in practice our CI still hasn't fully caught up (IIRC IAR and VS _still_ don't accept declarations after statements with the settings we're using on the CI).

I tried to place the check before declarations; it compiled fine with the current VS 2019. So later it might be possible to return to the question of reordering code lines.

mpg:
> Would it be correct to assume that similar changes will be made in mbedtls repository too?

Yes, we're actually going to merge mbed-crypto/development into mbedtls/development. We're just waiting for the last few PRs here before we do that. (By the way, if you want to contribute other improvements and fixes, your next PRs should be to the mbedtls repository.)

> I tried to place the check before declarations; it compiled fine with the current VS 2019. So later it might be possible to return to the question of reordering code lines.

Unfortunately we might still have to support older version. I think currently the minimum version we claim to support (and test in our CI) is VS 2013. Hopefully at some point we can move past that!

gilles-peskine-arm:
Yes, we made the change in both mbed-crypto/development and mbedtls/development.

We currently test with VS 2013 and up, which in theory should support this particular C99 feature. But I remember getting a CI failure when I put a declaration in the middle of a block. It was for an urgent feature so I didn't investigate and just moved the declaration.

We may well drop VS 2013, but requiring only VS 2019 would be too restrictive.

irwir:
This was not a blocking issue and it is fine to support older compilers.

> We currently test with VS 2013 and up, which in theory should support this particular C99 feature.

You could find in the net, that even VS 2017 provided incomplete C99 support.
A while ago I had errors with misplaced declarations, but cannot tell for sure the version. It might be VS 2015 or even one of the earlier versions of VS 2017. Now it would be too much effort to check again. The current versions of VS 2017/2019 do compile declarations after code.

gilles-peskine-arm:
> even VS 2017 provided incomplete C99 support

I know. For example https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=vs-2019#note_C states that variadic macros aren't fully correctly supported. And `restrict` is a keyword in VS 2013-2017 but means something completely different from standard C. But according to https://web.archive.org/web/20130701001452/http://blogs.msdn.com/b/vcblog/archive/2013/06/28/c-11-14-stl-features-fixes-and-breaking-changes-in-vs-2013.aspx, several C99 features including declarations after statements are supposed to be in VS 2013.

irwir:
This PR was not merged into the 

> Yes, we're actually going to merge mbed-crypto/development into mbedtls/development. We're just waiting for the last few PRs here before we do that.

Is there still anything to wait for, or this PR could be merged to mbedtls repository?

PR #276, opened 2019-09-23
Relax input restrictions for key derivation

gilles-peskine-arm:
Key derivation requires a `PSA_KEY_TYPE_DERIVE` key for a `PSA_KEY_DERIVATION_INPUT_SECRET` step. This is the common case, but sometimes applications have a buffer in memory which they'll only use for a single key derivation. Sometimes this is even an empty buffer, as when [deriving an IV in EAP-TLS](https://github.com/ARMmbed/mbed-crypto/issues/137#issuecomment-498607672). So allow the `SECRET` to be passed with `psa_key_derivation_input_bytes`.

Conversely, allow other inputs to be passed as `RAW_DATA` key objects with `psa_key_derivation_input_key`.

Fix ARMmbed/psa-crypto#257. Needed to repair the `ssl-opt.sh` test case `export keys functionality` which fails when using the latest version of Mbed Crypto (since https://github.com/ARMmbed/mbed-crypto/pull/257 was merged).


library/psa_crypto.c
```diff
@@ -5076,15 +5076,41 @@ static psa_status_t psa_tls12_prf_psk_to_ms_input(
...
 #endif /* MBEDTLS_MD_C */
 
+static int psa_key_derivation_check_input_type(
+    psa_key_derivation_step_t step,
+    psa_key_type_t key_type )
+{
+    switch( step )
+    {
+        case PSA_KEY_DERIVATION_INPUT_SECRET:
+            if( key_type == PSA_KEY_TYPE_DERIVE || key_type == 0 )
```
AndrzejKurek: What does a key type of zero mean? Why not use a define for anything it is?

gilles-peskine-arm: 0 is not a valid key type. Here it's used to mean that the input is a direct string rather than a key object.

AndrzejKurek: A separate define signalling that (to be used in test data, since it's ambiguous there too) would be useful. If not - at least a comment here and in tests.

yanesca:
This solves the problem, but I think it relaxes the API too much and makes it easier to misuse. (Of course whatever we do, a dedicated user can misuse it anyways, but I don't think we should make it any easier than it is necessary.)

This way the user can input any data as secret in a single call, potentially making all the derived keys insecure. (Before they had to import it into a key slot to do so, that should raise an alarm in most people and hopefully make them think twice.)

Performing an operation on a key that was originally intended for non-secret data can potentially leak information about the key. (In this particular case I don't think it does, but this interface should be able to serve legacy and future algorithms too, we shouldn't lay traps like this for them if we don't have to.)

We could avoid this if instead of relaxing the constraints, we added a third variant of TLS PRF for IV generation that does not take a secret input at all and throws an error if `psa_key_derivation_output_key()` is called on the operation. What do you think?

gilles-peskine-arm:
> We could avoid this if instead of relaxing the constraints, we added a third variant of TLS PRF for IV generation that does not take a secret input at all and throws an error if `psa_key_derivation_output_key()` is called on the operation. 

I don't like adding another algorithm. What do we do when the next protocol comes along and does the same with some other KDF?

On the other hand, I do like the idea of forbidding `psa_key_derivation_output_key()` if the SECRET input isn't a key. Do you think that's enough?

library/psa_crypto.c
gilles-peskine-arm: Done.

athoelke:
>

> On the other hand, I do like the idea of forbidding `psa_key_derivation_output_key()` if the SECRET input isn't a key. Do you think that's enough?

It's definitely cute, and makes some conceptual sense: "don't permit secret outputs if there were no secret inputs".

If the caller really wants to us the output as a key they then have to import the output bytes as a key - which probably alerts a reviewer to check what is happening in that code.

If the caller wants to deterministically generate a key from a non-secret input, then (as currently) they have to import the non-secret to a key-derivation key first. Which also triggers a reviewer to inspect the application behaviour.

yanesca:
> On the other hand, I do like the idea of forbidding psa_key_derivation_output_key() if the SECRET input isn't a key. Do you think that's enough?

Yes, I think it should be enough to make allowing non-key SECRET inputs.

On the other hand I still feel uncomfortable about allowing keys as a non-SECRET input. This relaxation is not necessary for the current fix, can't we just keep the restriction and hope that eventually we won't need to lift it?

gilles-peskine-arm:
Label-type inputs have to be RAW_DATA, not DERIVE. It makes sense to me to allow RAW_DATA for a label or info (a personalization string kept in persistent storage), or for a salt or seed (a device-specific value that may or may not be secret and is kept in persistent storage). @yanesca What makes you uncomfortable? The risk that an application will pass the wrong key (which would have to be of the wrong type)?

yanesca:
I am imagining here a RAW_DATA that is supposed to be secret and a key derivation algorithm that leaks information about for example the label. (Yes, such a key derivation algorithm is not safe in the first place.)

I admit that what I am afraid of is highly hypothetical, I just thought if we can avoid it then why not?

gilles-peskine-arm:
It wouldn't necessarily be bad for a KDF to leak the “label”, although I can't think of a common construction that comes anywhere near it.

RAW_DATA is often not supposed to be kept secret anyway. A common use case is to store a certificate next to the corresponding private key.

A RAW_DATA does need to have the DERIVE usage flag to be used as a label (etc) in a key derivation.

gilles-peskine-arm:
I implemented the constraint on output_key.

And now I need to resolve a merge conflict. I'm afraid I'm going to need to rebase this. The current state is in https://github.com/gilles-peskine-arm/mbed-crypto/tree/psa-key_derivation-relax_inputs-2.

gilles-peskine-arm:
I rebased on top of `development` plus one new commit https://github.com/ARMmbed/mbed-crypto/pull/276/commits/6ddb4d84341e739a9f245cca2168a1b16ae831da "Improve descriptions of derive test cases". Thanks to that commit, there were no conflicts.

include/psa/crypto_struct.h
```diff
@@ -255,6 +255,7 @@ typedef struct psa_tls12_prf_key_derivation_s
 struct psa_key_derivation_s
 {
     psa_algorithm_t alg;
+    unsigned int can_output_key : 1;
```
gilles-peskine-arm: This leaves 31 bits unused. https://github.com/ARMmbed/mbed-crypto/issues/281

yanesca:
The CI fails only on the ABI check and the known export key issue.

athoelke:
I'm happy with the documentation improvements. 👍

PR #277, opened 2019-09-23
Improve speed of PBKDF2 by caching the digest state of the passphrase

jack-fortanix:
mbedtls_md_hmac_starts computes a hash block consisting of H(passphrase xor ipad), but this operations is performed again and again during the PBKDF2 computation even though it always produces the same result. Instead call `mbedtls_md_hmac_starts` once at the start of computation and use `mbedtls_md_hmac_restart` to reuse the same key again.

In my measurements (x86-64 machine, gcc, using SHA-256, 1 million iterations) this is about 10% faster than the current code, which matters quite a lot eg on a server which regularly validates user logins against a hash database.

### DIFF
--- a/library/pkcs5.c
+++ b/library/pkcs5.c
@@ -243,13 +243,12 @@ int mbedtls_pkcs5_pbkdf2_hmac( mbedtls_md_context_t *ctx,
         return( MBEDTLS_ERR_PKCS5_BAD_INPUT_DATA );
 #endif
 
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm:
Thank you for your contribution! We'll review it as soon as we can.

The CI failures are a known issue which is unrelated to this PR.

gilles-peskine-arm:
CI is ok except for some TLS tests that failed due to an unrelated issue in mbedtls (since fixed).

PR #279, opened 2019-09-24
Include IANA reference in the definition of ECC curves and DH groups

athoelke:
Fixes ARMmbed/psa-crypto#262

### DIFF
--- a/include/psa/crypto_types.h
+++ b/include/psa/crypto_types.h
@@ -111,9 +111,35 @@ typedef uint16_t psa_ecc_curve_t;
  * 1. Select a ::psa_dh_group_t value in the range #PSA_DH_GROUP_VENDOR_MIN to
  *    #PSA_DH_GROUP_VENDOR_MAX, which is a subset of the IANA private use
  *    range.
- * 2. Use a ::psa_key_type_t value that is vendor-defined.
+ * 2. Select a ::psa_dh_group_t value from the named groups allocated for
+ *    GREASE in the IETF draft specification. The GREASE specification and
+ *    values are listed below.
+ * 3. Use a ::psa_key_type_t value that is vendor-defined.
  *
- * The first option is recommended.
+ * Option 1 or 2 are recommended.
+ *
+ * The current draft of the GREASE specification is
+ * https://datatracker.ietf.org/doc/draft-ietf-tls-grease
+ *
+ * The following GREASE values are allocated for named groups:
+ * \code
+ * 0x0A0A
+ * 0x1A1A
+ * 0x2A2A
+ * 0x3A3A
+ * 0x4A4A
+ * 0x5A5A
+ * 0x6A6A
+ * 0x7A7A
+ * 0x8A8A
+ * 0x9A9A
+ * 0xAAAA
+ * 0xBABA
+ * 0xCACA
+ * 0xDADA
+ * 0xEAEA
+ * 0xFAFA
+ * \endcode
  */
 typedef uint16_t psa_dh_group_t;
 
--- a/include/psa/crypto_types.h
+++ b/include/psa/crypto_types.h
@@ -65,10 +65,24 @@ typedef int32_t psa_status_t;
  */
 typedef uint32_t psa_key_type_t;
 
-/** The type of PSA elliptic curve identifiers. */
+/** The type of PSA elliptic curve identifiers.
+ *
+ * The encoding of curve identifiers is aligned with the
+ * TLS Supported Groups Registry (formerly known as the
+ * TLS EC Named Curve Registry)
+ * https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
+ * The values are defined by RFC 8422 and RFC 7027.
+ */
 typedef uint16_t psa_ecc_curve_t;
 
-/** The type of PSA Diffie-Hellman group identifiers. */
+/** The type of PSA Diffie-Hellman group identifiers.
+ *
+ * The encoding of group identifiers is aligned with the
+ * TLS Supported Groups Registry (formerly known as the
+ * TLS EC Named Curve Registry)
+ * https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
+ * The values are defined by RFC 7919.
+ */
 typedef uint16_t psa_dh_group_t;
 
 /** \brief Encoding of a cryptographic algorithm.
--- a/include/psa/crypto_types.h
+++ b/include/psa/crypto_types.h
@@ -67,21 +67,29 @@ typedef uint32_t psa_key_type_t;
 
 /** The type of PSA elliptic curve identifiers.
  *
- * The encoding of curve identifiers is aligned with the
+ * The encoding of curve identifiers is taken from the
  * TLS Supported Groups Registry (formerly known as the
  * TLS EC Named Curve Registry)
  * https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
  * The values are defined by RFC 8422 and RFC 7027.
+ *
+ * This specification defines identifiers for some of the curves in the IANA
+ * registry. Implementations that support other curves that are in the IANA
+ * registry should use the IANA value and a implementation-specific identifier.
  */
 typedef uint16_t psa_ecc_curve_t;
 
 /** The type of PSA Diffie-Hellman group identifiers.
  *
- * The encoding of group identifiers is aligned with the
+ * The encoding of group identifiers is taken from the
  * TLS Supported Groups Registry (formerly known as the
  * TLS EC Named Curve Registry)
  * https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
  * The values are defined by RFC 7919.
+ *
+ * This specification defines identifiers for some of the groups in the IANA
+ * registry. Implementations that support other groups that are in the IANA
+ * registry should use the IANA value and a implementation-specific identifier.
  */
 typedef uint16_t psa_dh_group_t;
 
--- a/include/psa/crypto_types.h
+++ b/include/psa/crypto_types.h
@@ -75,7 +75,6 @@ typedef uint32_t psa_key_type_t;
  * TLS Supported Groups Registry (formerly known as the
  * TLS EC Named Curve Registry)
  * https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
- * The values are defined by RFC 8422 and RFC 7027.
  *
  * This specification defines identifiers for some of the curves in the IANA
  * registry. Implementations that support other curves that are in the IANA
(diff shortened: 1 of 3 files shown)


include/psa/crypto_types.h
```diff
@@ -65,10 +65,24 @@ typedef int32_t psa_status_t;
...
 typedef uint32_t psa_key_type_t;
 
-/** The type of PSA elliptic curve identifiers. */
+/** The type of PSA elliptic curve identifiers.
+ *
+ * The encoding of curve identifiers is aligned with the
+ * TLS Supported Groups Registry (formerly known as the
+ * TLS EC Named Curve Registry)
+ * https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
+ * The values are defined by RFC 8422 and RFC 7027.
```
gilles-peskine-arm: The values defined in this specification, but the IANA registry has more and may add more in the future.

I think we should take this one step further and specify the IANA registry is included by reference with respect to numerical values. This specification only contributes the names, which are generally `PSA_ECC_CURVE_` + uppercase(IANA.description), but there are exceptions (additional underscore for Brainpool, `CURVE_` rather than `x` for 25519 and 448).

Thus vendors who want to support other curves that are in the IANA registry should use the IANA value and a vendor-specific name. Vendors who want to support non-IANA curves should use the vendor range.

And we should define a vendor range that is a subset of the IANA private use range (which is only 0x01fc..0x01ff and 0xfe00..0xfeff).

athoelke: Is it worth combining the `psa_dh_group_t` and `psa_ecc_curve_t` types, given that they are both extracted from the same enumeration in the IANA registry? Then we can provide a single overall description of the association with the registry regarding the symbols/values defined in this spec, the use of other values by vendors and the right way to declare non-IANA curves.

It might be helpful to also move the type(s) and the ECC and DH definitions into a separate subsection of the API reference, so they clearly are connected to these overall description?

gilles-peskine-arm: I prefer to keep them separate because there's no context in which you can pass either. Not that C enforces integral types anyway, but they're semantically distinct.

athoelke: > I think we should take this one step further and specify the IANA registry is included by reference with respect to numerical values.

Some specific examples to clarify your suggestion. Is the proposal that all defined IANA registry values are "in scope" of the specification when combined with the appropriate ECC or DH keys types - and therefore not "vendor defined"? For example:

- The key type for a public key of an IANA registered ECC curve, but which is not explicitily provided an identifier in this specification, would have the value `0x6003xxxx`. It would not require the vendor flag to be set (`0xE003xxxx`).

Would the key type for a non-IANA ECC curve, require:
1. Use of a value in the IANA private use range for the curve id.
2. Use of the vendor flag in the key type.
3. Either (1) or (2).
4. Both (1) and (2). Although, presumably if the vendor flag is used, the use of a private use id and the `0x6003xxxx` key type is just a recommendation.

gilles-peskine-arm: (3), I think. Curve encodings have their own existence independently of key type encodings.

athoelke: Just wrote that all up in some follow-up commits.

I've sliced off half of each of the private-use ranges for vendor-defined identifiers, leaving half for future potential use by the PSA specification. Let me know if that approach does not make sense...

include/psa/crypto_values.h
```diff
@@ -535,6 +582,25 @@
 #define PSA_DH_GROUP_FFDHE6144          ((psa_dh_group_t) 0x0103)
 #define PSA_DH_GROUP_FFDHE8192          ((psa_dh_group_t) 0x0104)
 
+/** Minimum value for a vendor-defined Diffie Hellman group identifier
+ *
+ * The range for vendor-defined group identifiers is a subset of the IANA
+ * registry private use range, `0x01fc` - `0x01ff`.
```
gilles-peskine-arm: Unfortunately, this is a ridiculously small range whereas an implementation may well wish to define many custom groups, since many groups have been common in the wild and specific systems or protocols may define their own groups.

The TLS registry puts EC and FF groups in the same 16-bit range, but we don't have to. I propose to use some values in the IANA EC range for vendor FF groups. We can in particular use the [GREASE](https://datatracker.ietf.org/doc/draft-ietf-tls-grease/) ranges, which TLS uses for the purpose of _not_ having a corresponding cryptographic algorithm.

include/psa/crypto_values.h
```diff
@@ -447,6 +455,12 @@
                         ((type) & PSA_KEY_TYPE_ECC_CURVE_MASK) : \
                         0))
 
+/**@}*/
+
+/** \defgroup iana_groups TLS Supported Groups
+* @{
```
gilles-peskine-arm: `*` needs to be in column 1, otherwise `check-names.sh` gags.

include/psa/crypto_types.h
athoelke: I am not an expert in this material, and happily defer to your judgement. Are you suggesting that I just remove the line with the RFC reference:

> * The values are defined by RFC 8422 and RFC 7027.

include/psa/crypto_values.h
```diff
@@ -447,6 +455,12 @@
                         ((type) & PSA_KEY_TYPE_ECC_CURVE_MASK) : \
                         0))
 
+/**@}*/
+
+/** \defgroup iana_groups TLS Supported Groups
```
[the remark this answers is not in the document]
athoelke: Because the documentation of algorithm and key identifiers is currently a big swirling mass without structure. This was an attempt to co-locate the type definition - with the documentation about other ECC values - and the defined values of that type.

A better alternative might be to provide more structure by splitting the single keys & algorithm bucket into multiple sections divided by key and algorithm types. This makes it make easier to reference the relevant algorithms from the documentation on the related functions. I'll revert this change and revisit it once #149 (migrate to reST) is done.

include/psa/crypto_values.h
athoelke: Fixed by removing the new defgroup.

include/psa/crypto_values.h
athoelke: From a quick glance, the proposed GREASE values that can be used in this way are:

```
0x0A0A, 0x1A1A, 0x2A2A, 0x3A3A, 0x4A4A, 0x5A5A, 0x6A6A, 0x7A7A,
0x8A8A, 0x9A9A, 0xAAAA, 0xBABA, 0xCACA, 0xDADA, 0xEAEA, 0xFAFA
```

This is still draft and subject to change. This might not affect this API element, but should modify the documentation of `psa_dh_group_t`.

1. Is the suggestion to directly list the values, as being permitted for vendor-defined additional DH groups, or indirectly by referencing the specification draft?
2. Do we want to slice up the GREASE identifiers, and reserve a set for PSA Crypto future use?
3. Is the [arbitrary] 50/50 split of the TLS support group private use range appropriate (for this and ECC)?

include/psa/crypto_types.h
athoelke: Is the same comment applicable to the RFC 7919 reference in the documenation for `psa_dh_group_t`?

gilles-peskine-arm: Yes, please remove this line and the similar line for `psa_dh_group_t`.

include/psa/crypto_values.h
gilles-peskine-arm: My proposal was to define the GREASE values (stated by value, with the GREASE spec given as an informative rationale) as reserved for vendors in `psa_dh_group_t`.

Though given that there are only 16 of them, that's not much. We can punt PSA crypto future use until later, but I do think we need more vendor FFDH values. I guess the only solution (short of moving to a bigger type) is to take values that are already assigned to elliptic curves and declare them as vendor-specific FFDH values.

athoelke: I've added a commit the extends the recommendations for vendor defined DH groups.

athoelke:
>The xxx_VENDOR_{MIN,MAX} macros need to be excluded from scripts/generate_psa_constants.py and tests/scripts/test_psa_constant_names.py to fix the build.

Can someone help with making the associated change to those scripts. There is no obviously correct place to exclude these definitions from the testing that is done here.

Why are these values not suitable for this test?

gilles-peskine-arm:
`generate_psa_constants.py` works out preprocessor_expression → numerical_value mappings, and generates C code that constructs a mapping in the other direction. If the mapping is not one-to-one,  it errors out. There are rules to exclude certain preprocessor expressions which are considered to be aliases of other expressions.

I'll update the scripts.

gilles-peskine-arm:
The problem was that the (for now) mbedcrypto-specific value `PSA_DH_GROUP_CUSTOM` had the same value as `PSA_DH_GROUP_VENDOR_MIN`. I've moved it to be somewhere else in the standard range, far from values that are currently in used. It's for a feature that isn't implemented yet, and we can see later whether this becomes a standard feature (perhaps with the same number, perhaps not) or stays implementation-specific (in which case we'll change the number to be in the vendor range).

gilles-peskine-arm:
CI report: the TLS head job failed in `test_suite_x509parse` because mbedtls needs a bug fix in mbed-crypto that is more recent than this PR. The TLS merge job succeeded so that's good. The crypto [merge job](https://jenkins-internal.mbed.com/blue/organizations/jenkins/mbed-crypto-pr/detail/PR-279-merge/11/pipeline) failed in the [Windows tests](https://jenkins-internal.mbed.com/job/mbed-crypto-pr/job/PR-279-merge/11/execution/node/3908/log/?consoleFull) due to what seems to have been platform instability (“error MSB4166: Child node "4" exited prematurely”). I'm rerunning the Windows crypto tests: https://jenkins-internal.mbed.com/view/mbed-crypto/job/mbedtls-psa-release-new/574/

gilles-peskine-arm:
https://jenkins-internal.mbed.com/view/mbed-crypto/job/mbedtls-psa-release-new/574/ passed so we have full CI approval. Since I made some of the changes, this needs approval from a team member other than me.

athoelke:
Is there any chance that this could be reviewed, approved and merged today (if there is no issue with the implementation, obviuosly? - please? :-)

include/psa/crypto_values.h
```diff
@@ -535,6 +564,19 @@
 #define PSA_DH_GROUP_FFDHE6144          ((psa_dh_group_t) 0x0103)
 #define PSA_DH_GROUP_FFDHE8192          ((psa_dh_group_t) 0x0104)
 
+/** Minimum value for a vendor-defined Diffie Hellman group identifier
+ *
+ * The range for vendor-defined group identifiers is a subset of the IANA
+ * registry private use range, `0x01fc` - `0x01ff`.
```
Patater: @athoelke To confirm, this range is still accurate, but the GREASE values are also available so the practical use of this range is limited.

athoelke: Yes, that is correct. An minor improvement might be to refer to the `psa_dh_group_t` definition and its description of other ways to allocate a vendor-defined group number.

However, I'd like this merged sooner so I can render the reST and separate the specification from the doxygen source code, which is a pre-requistite to completing the remaing API changes before publishing 1.0.0 of the spec. I can clarify/improve the text after the split if desirable.

PR #280, opened 2019-09-24
Include vendor-defined algorithm macros

athoelke:
Providing equivalent documentation to that for the vendor flag for key types.

Fixes ARMmbed/psa-crypto#264


gilles-peskine-arm:
The CI failures for TLS are a known issue that is unrelated to this documentation-only PR.

PR #285, opened 2019-10-01
SE driver: make persistent data work

gilles-peskine-arm:
Secure element interface changes:

* A driver without a `p_validate_slot_number` method no longer allows registering an existing key with `mbedtls_psa_register_se_key`.
* The `p_validate_slot_number` method now takes an extra read-write parameter `persistent_data`, like `p_allocate`.

New features and bug fixes:
* Call the `p_init` driver method during `psa_crypto_init`.
* Fix loading of the persistent data during init.
* Add some minimal testing (good cases only) for saved persistent data.

### DIFF
--- a/include/psa/crypto_se_driver.h
+++ b/include/psa/crypto_se_driver.h
@@ -927,7 +927,14 @@ typedef psa_status_t (*psa_drv_se_allocate_key_t)(
  * sake of initial device provisioning or onboarding. Such a mechanism may
  * be added to a future version of the PSA Cryptography API specification.
  *
+ * This function may update the driver's persistent data through
+ * \p persistent_data. The core will save the updated persistent data at the
+ * end of the key creation process. See the description of
+ * ::psa_drv_se_allocate_key_t for more information.
+ *
  * \param[in,out] drv_context   The driver context structure.
+ * \param[in,out] persistent_data   A pointer to the persistent data
+ *                                  that allows writing.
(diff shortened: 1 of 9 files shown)


include/psa/crypto_se_driver.h
```diff
@@ -927,7 +927,14 @@ typedef psa_status_t (*psa_drv_se_allocate_key_t)(
  * sake of initial device provisioning or onboarding. Such a mechanism may
  * be added to a future version of the PSA Cryptography API specification.
  *
+ * This function may update the driver's persistent data through
+ * \p persistent_data. The core will save the updated persistent data at the
+ * end of the key creation process. See the description of
+ * ::psa_drv_se_allocate_key_t for more information.
```
Patater: Should we have a test to make sure that the core saved the persistent data?

gilles-peskine-arm: Yes, but that's part of our tech debt. We don't have one for allocate either.

gilles-peskine-arm: I added some testing (good cases only). And fixed a bug that made it not work on deinit/reinit.

gilles-peskine-arm:
CI passed on -head. The failures on -merge are due to a recent CI instability. The PR is from a recent branch and we'd have been happy to merge with a CI job that had run on exactly the same content as the -head job. So CI is ok to merge.

PR #286, opened 2019-10-02
Fix defgroup syntax for API version section

athoelke:


gilles-peskine-arm:
This only affects the PSA documentation. CI passed. Merging with one approval.

PR #287, opened 2019-10-03
DRBG documentation improvements

gilles-peskine-arm:
Improve the documentation of the CTR_DRBG and HMAC_DRBG modules. This part is a straightfoward forward-port of https://github.com/ARMmbed/mbedtls/pull/2860: the 2.16 and 2.19 versions of `hmac_drbg.h` and `ctr_drbg.h` were near-identical and I changed them in identical ways.

```
git diff upstream-public/pr/2860  -- include/mbedtls/{ctr,hmac}_drbg.h
```

I am working on a separate pull request for the following changes (separate PR because it was getting really messy):

1. `mbedtls_hmac_drbg_seed()` and `mbedtls_ctr_drbg_seed()` overrode the entropy length in the DRBG context, so calling `mbedtls_hmac_drbg_set_entropy_len()` or `mbedtls_ctr_drbg_set_entropy_len()` before calling `xxx_seed()` had no effect. Change the `xxx_seed()` functions to respect the entropy length set with `xxx_set_entropy_len()`.
1. In CTR_DRBG, systematically grab a nonce from the entropy source. This removes the difficulty of using the module in a NIST-compliant way.

Backports: 2.16 in https://github.com/ARMmbed/mbedtls/pull/2860, 2.7 in https://github.com/ARMmbed/mbedtls/pull/2878.

Internal ref: IOTCRYPT-180, IOTCRYPT-457


gilles-peskine-arm:
CI failed in the Windows 2013 job. The failure is the CTR_DRBG selftest, The reason the selftest started failing is that I moved `MBEDTLS_CTR_DRBG_USE_128_BIT_KEY` to a different section in `config.pl`, so it got included in `config.pl full`. The `selftest` program has been failing when `MBEDTLS_CTR_DRBG_USE_128_BIT_KEY` is enabled ever since we introduced this option (#290) but we didn't realize until now because we were never enabling this option on the CI (#289).

I've excluded `MBEDTLS_CTR_DRBG_USE_128_BIT_KEY` from `config.pl full`. Fixing #289 and #290 is out of scope of this PR.

gilles-peskine-arm:
The merge job passed except for `check-generated-files.sh` in TLS which is a known issue with a fix pending.

The head job failed because of some Jenkins wonkiness that's unrelated to the PR.

include/mbedtls/ctr_drbg.h
```diff
@@ -56,9 +79,19 @@
 #define MBEDTLS_CTR_DRBG_BLOCKSIZE          16 /**< The block size used by the cipher. */
 
 #if defined(MBEDTLS_CTR_DRBG_USE_128_BIT_KEY)
-#define MBEDTLS_CTR_DRBG_KEYSIZE            16 /**< The key size used by the cipher (compile-time choice: 128 bits). */
+#define MBEDTLS_CTR_DRBG_KEYSIZE            16
+/**< The key size in bytes used by the cipher.
+ *
+ * Compile-time choice: 16 bytes (128 bits)
+ * because #MBEDTLS_CTR_DRBG_USE_128_BIT_KEY is enabled.
```
AndrzejKurek: What's the difference between `#MBEDTLS_CTR_DRBG_USE_128_BIT_KEY` and `\c MBEDTLS_CTR_DRBG_USE_128_BIT_KEY`? Since I can see that both of these versions are used.

include/mbedtls/ctr_drbg.h
```diff
@@ -209,18 +298,42 @@ void mbedtls_ctr_drbg_set_prediction_resistance( mbedtls_ctr_drbg_context *ctx,
...
+ * \note                The security strength of CTR_DRBG is bounded by the
+ *                      entropy length. Thus:
+ *                      - When using AES-256
+ *                        (\c MBEDTLS_CTR_DRBG_USE_128_BIT_KEY is disabled,
+ *                        which is the default),
+ *                        \p len must be at least 32 (in bytes)
+ *                        to achieve a 256-bit strength.
+ *                      - When using AES-128
+ *                        (\c MBEDTLS_CTR_DRBG_USE_128_BIT_KEY is enabled)
+ *                        \p len must be at least 16 (in bytes)
```
AndrzejKurek: Didn't you mention previously that in such case the entropy len has to be at least 24?

> 128 bits if AES-128 is used (\c MBEDTLS_CTR_DRBG_USE_128_BIT_KEY enabled)
>  *   and #MBEDTLS_CTR_DRBG_ENTROPY_LEN is set to 24 or more (which is
>  *   always the case unless it is explicitly set to a different value
>  *   in config.h).

include/mbedtls/ctr_drbg.h
gilles-peskine-arm: `#MACRO` typesets `MACRO` and links to the definition of `MACRO` if `MACRO` is defined, but typesets `#MACRO` if `MACRO` is undefined, so it isn't good for preprocessor symbols that may or may not be defined. Hence the use of `\c MACRO` for symbols that may or may not be defined.

include/mbedtls/ctr_drbg.h
gilles-peskine-arm: 24 for the initial seeding, 16 for subsequent reseeds. This function only influences subsequent reseeds. `MBEDTLS_CTR_DRBG_ENTROPY_LEN` influences both.

include/mbedtls/ctr_drbg.h
AndrzejKurek: Shouldn't it be `\c MBEDTLS_CTR_DRBG_USE_128_BIT_KEY` everywhere then? Or will doxygen ignore blocks that won't get compiled?

gilles-peskine-arm: It can be `#MBEDTLS_CTR_DRBG_USE_128_BIT_KEY` here because this comment is surrounded by `#ifdef MBEDTLS_CTR_DRBG_USE_128_BIT_KEY`.

gilles-peskine-arm:
[CI ok](https://github.com/ARMmbed/mbed-crypto/pull/287#issuecomment-539110867). PR approved, near-identical PR for 2.16 approved, 2.7 backport approved. This is good to merge.

PR #288, opened 2019-10-07
Add ECDSA tests with hash and key of different lengths

gilles-peskine-arm:
Add some ECDSA test cases where the hash is shorter or longer than the key length, to check that the API doesn't enforce a relationship between the two.


gilles-peskine-arm:
CI is flaky (failures of CY8CKIT boards, a Windows job failing due to what is probably a disk full error), but there's enough coverage between the -head job and the -merge job to see that everything succeeds.

PR #293, opened 2019-10-09
Always gather MBEDTLS_ENTROPY_BLOCK_SIZE bytes of entropy

gilles-peskine-arm:
Enforce that `mbedtls_entropy_func` gathers a total of `MBEDTLS_ENTROPY_BLOCK_SIZE` bytes or more from strong sources. In particular, in the default configuration, on a platform with a single entropy source, `mbedtls_entropy_func` will now grab 64 bytes from that source (due to the new lower bound of `MBEDTLS_ENTROPY_BLOCK_SIZE`), whereas it formerly only grabbed 32 bytes (due to the per-source threshold).

Backports: no — this is a security improvement but not a critical fix (32 bytes is enough if the entropy is of good quality) and it could be a problem on platforms where gathering more entropy would take an unacceptable amount of time.

Internal ref: IOTCRYPT-454


library/entropy.c
```diff
@@ -363,12 +364,17 @@ int mbedtls_entropy_func( void *data, unsigned char *output, size_t len )
...
 
-        done = 1;
+        thresholds_reached = 1;
+        strong_size = 0;
         for( i = 0; i < ctx->source_count; i++ )
+        {
             if( ctx->source[i].size < ctx->source[i].threshold )
-                done = 0;
+                thresholds_reached = 0;
+            if( ctx->source[i].strong == MBEDTLS_ENTROPY_SOURCE_STRONG )
```
AndrzejKurek: Is it possible to reach `ENTROPY_MAX_LOOP`  here if `ctx->source[i].strong == MBEDTLS_ENTROPY_SOURCE_STRONG ` is `false`? There's no way to increase `strong_size` in such case.

gilles-peskine-arm: Yes, but only if there is no strong source. And in this case the call to `entropy_gather_internal` would have returned `MBEDTLS_ERR_ENTROPY_NO_STRONG_SOURCE`.

gilles-peskine-arm:
The CI failures are not just due to unrelated problems that plagued the CI for the past few months. I'm investigating.

AndrzejKurek:
The CI seems to still be failing in strange ways. Are these fails related to the PR?

gilles-peskine-arm:
I think many jobs are failing on -head because it's an old branch that's broken on some boards and incompatible with the current version in mbedtls. Since -merge is passing, the CI is ok.

PR #298, opened 2019-10-10
Test a build without any asymmetric cryptography

### LINKED ISSUE #296: Test a build with only symmetric cryptography
### Description

We have no build with only symmetric cryptography anywhere in `all.sh`. In particular, we never build with `MBEDTLS_BIGNUM_C` disabled, which happens in the real world either when building with no asymmetric cryptography or when all the asymmetric cryptography comes from ALT implementations that don't use any bignum functions.

This could cause the mbedtls configuration [`config-ccm-psk-tls1_2.h`](https://github.com/ARMmbed/mbedtls/blob/development/configs/config-ccm-psk-tls1_2.h) to fail. As of 9ab7c07f1f370636fcaa8bc02e6f45035fab1596, mbed-crypto builds and passes TLS tests with `config-ccm-psk-tls1_2.h`, but the `test_suite_asn1parse` fails to build when `MBEDTLS_ASN1_PARSE_C` is enabled but `MBEDTLS_BIGNUM_C` is disabled.

We can't test every possible combination of build options, but we should test at least one configuration without the bignum module. I propose to test with all features that don't depend on `MBEDTLS_BIGNUM_C`.

### Issue request type

    [ ] Question
    [x] Enhancement
    [ ] Bug

gilles-peskine-arm:
Fix #296.

Also fix two bugs that the tests revealed: `component_test_se_full` wasn't using the full config, and there was a memory leak in an SE HAL test (fix duplicated in https://github.com/ARMmbed/mbed-crypto/pull/304).

### DIFF
--- a/configs/config-symmetric-only.h
+++ b/configs/config-symmetric-only.h
@@ -1,0 +1,99 @@
+/**
+ * \file config-symmetric-only.h
+ *
+ * \brief Configuration without any asymmetric cryptography.
+ */
+/*
+ *  Copyright (C) 2019, ARM Limited, All Rights Reserved
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *  not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *  This file is part of mbed TLS (https://tls.mbed.org)
+ */
+
+#ifndef MBEDTLS_CONFIG_H
+#define MBEDTLS_CONFIG_H
+
+/* System support */
+//#define MBEDTLS_HAVE_ASM
+#define MBEDTLS_HAVE_TIME
+#define MBEDTLS_HAVE_TIME_DATE
+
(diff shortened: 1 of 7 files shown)


configs/config-symmetric-only.h
```diff
@@ -0,0 +1,99 @@
...
+ *  limitations under the License.
+ *
+ *  This file is part of mbed TLS (https://tls.mbed.org)
+ */
+
+#ifndef MBEDTLS_CONFIG_H
+#define MBEDTLS_CONFIG_H
+
+/* System support */
+//#define MBEDTLS_HAVE_ASM
```
AndrzejKurek: Why is this line left in, even though it's commented?

gilles-peskine-arm: I left a couple of options commented out because I thought that they were options that should not be enabled in our test configuration, but could be useful to users who want to make a symmetric-only configuration tuned to their use case, and use this file as a starting point.

tests/suites/test_suite_psa_crypto_se_driver_hal.function
```diff
@@ -565,16 +565,17 @@ static int check_persistent_data( psa_key_lifetime_t lifetime,
     psa_storage_uid_t uid = file_uid_for_lifetime( lifetime );
     struct psa_storage_info_t info;
     uint8_t *loaded = NULL;
+    int ok = 0;
```
AndrzejKurek: In such cases I'd rather use "is_ok" to show that it's a boolean, but that's an opinion.

gilles-peskine-arm: `int ok` is what I've used in similar support functions before, so I used the same style here for consistency.

gilles-peskine-arm:
New CI run now that the problems with the Cypress boards are fixed: https://jenkins-internal.mbed.com/job/mbed-crypto-pr/job/PR-298-merge/15/

PR #299, opened 2019-10-11
Allow xxx_drbg_set_entropy_len before xxx_drbg_seed

gilles-peskine-arm:
The functions `mbedtls_ctr_drbg_set_entropy_len()` and `mbedtls_hmac_drbg_set_entropy_len()` were only supported after the call to `mbedtls_ctr_drbg_seed()` and `mbedtls_hmac_drbg_seed()`, so the initial seeding always grabbed the amount of entropy defined at compile time (`MBEDTLS_CTR_DRBG_ENTROPY_LEN` for CTR_DRBG, calculated from the hash algorithm for HMAC_DRBG). There was no good reason for that and the documentation was not clear prior to https://github.com/ARMmbed/mbed-crypto/pull/287. Change the API to be intuitive: you can call `set_entropy_len()` after `init()`, and it will influence both the initial seeding (if you call it before `seed()`) and subsequent reseeds.

A side benefit is that this removes the need for the test-only function `mbedtls_ctr_drbg_seed_entropy_len`.

Upon reflection, the ways this would break existing applications are:
* An application that called `xxx_set_entropy_len()` before `xxx_drbg_seed` and expected it to do nothing. That's no more likely than an application that called `xxx_set_entropy_len()` and expected this to set the entropy length throughout the lifetime of the DRBG object, and is currently unknowingly grabbing the default amount of entropy instead.
* An application that called `mbedtls_ctr_drbg_seed_entropy_len` despite it being marked as for internal use only.
So I think this should be backported, perhaps with a backward compatibility wrapper `mbedtls_ctr_drbg_seed_entropy_len` for the sake of linkers that may have performed some cross-module inlining across a shared library call (hopefully that doesn't happen, but maybe we can spare a few bytes of code size to play it safe).

A step on the path to IOTCRYPT-180.

### DIFF
--- a/library/ctr_drbg.c
+++ b/library/ctr_drbg.c
@@ -381,10 +381,6 @@ exit:
     return( ret );
 }
 
-/*
- * Non-public function wrapped by mbedtls_ctr_drbg_seed(). Necessary to allow
- * NIST tests to succeed (which require known length fixed entropy)
- */
 /* CTR_DRBG_Instantiate with derivation function (SP 800-90A &sect;10.2.1.3.2)
  * mbedtls_ctr_drbg_seed_entropy_len(ctx, f_entropy, p_entropy,
  *                                   custom, len, entropy_len)
@@ -397,13 +393,11 @@ exit:
  * and with outputs
  *   ctx = initial_working_state
  */
-int mbedtls_ctr_drbg_seed_entropy_len(
-                   mbedtls_ctr_drbg_context *ctx,
-                   int (*f_entropy)(void *, unsigned char *, size_t),
-                   void *p_entropy,
-                   const unsigned char *custom,
-                   size_t len,
-                   size_t entropy_len )
+int mbedtls_ctr_drbg_seed( mbedtls_ctr_drbg_context *ctx,
+                           int (*f_entropy)(void *, unsigned char *, size_t),
+                           void *p_entropy,
+                           const unsigned char *custom,
+                           size_t len )
 {
     int ret;
     unsigned char key[MBEDTLS_CTR_DRBG_KEYSIZE];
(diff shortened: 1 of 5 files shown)


gilles-peskine-arm:
[CI passed](https://jenkins-internal.mbed.com/blue/organizations/jenkins/mbed-crypto-pr/detail/PR-299-merge/6/pipeline) except for ABI checking ([as expected](https://github.com/ARMmbed/mbed-crypto/pull/299#issuecomment-543234790)) and `getting-started-CY8CKIT_062_WIFI_BT-IAR` which fails with the same error (looks like a stack overflow) on `development`. Ok to merge.

library/ctr_drbg.c
```diff
@@ -415,7 +409,8 @@ int mbedtls_ctr_drbg_seed_entropy_len(
     ctx->f_entropy = f_entropy;
     ctx->p_entropy = p_entropy;
 
-    ctx->entropy_len = entropy_len;
+    if( ctx->entropy_len == 0 )
```
gacquroff: Hi @gilles-peskine-arm, is there any reason you didin't use `ctr_drbg_set_entropy_len()` here? Maybe you wanted to keep some level of method independency.

gilles-peskine-arm: I can't think of a strong reason one way or the other. Semantically, I want to set the `entropy_len` field to a certain value. `mbedtls_ctr_drbg_set_entropy_len()` does that, but so does an assignment. An assignment is simpler than calling a function. Arguably, calling a function is better in case the function changes. But changes to do what? If it changes to add some validation, this validation might not be what is needed at this point inside `mbedtls_ctr_drbg_seed`. And if the meaning of the field changes, there'll be other changes needed in the code, so who knows what would need to change in `mbedtls_ctr_drbg_seed`. In terms of code size, the assignment is a couple of bytes smaller than a function call. So I have a very slight preference for the direct assignment. But only very slight. I don't remember thinking about it explicitly at the time.

gacquroff: Thanks for the perspective.

PR #300, opened 2019-10-13
Changelog assembly script

gilles-peskine-arm:
This is a first go at tooling to assemble a changelog file for Mbed Crypto.

`scripts/assemble_changelog.py` reads changelog entries from `ChangeLog.d/*.md` and merges them into `ChangeLog.md`. Currently the script is independent from git but I expect future versions to determine the merge order by analyzing the git history.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -18,11 +18,11 @@
 
 ### Performance improvements
 
-* Remove a useless call to mbedtls_ecp_group_free(). Contributed by Alexander Krizhanovsky in #210.
+* Remove a useless call to `mbedtls_ecp_group_free()`. Contributed by Alexander Krizhanovsky in #210.
 * Speed up PBKDF2 by caching the digest calculation. Contributed by Jack Lloyd and Fortanix Inc in #277.
 
 ### Other changes
 
-* Remove the technical possibility to define custom md_info structures, which was exposed only in an internal header.
+* Remove the technical possibility to define custom `mbedtls_md_info` structures, which was exposed only in an internal header.
 
 ## Mbed Crypto 2.0.0
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,0 +1,28 @@
+# Mbed Crypto change history
+
+## Unreleased changes
+
+### Interface changes
+
+* In the PSA API, forbid zero-length keys. To pass a zero-length input to a key derivation function, use a buffer instead (this is now always possible).
+
+### New features
+
+* Key derivation inputs in the PSA API can now either come from a key object or from a buffer regardless of the step type.
+
+### Bug fixes
+
+* Fix a buffer overflow in the PSA HMAC code when using a long key with an unsupported algorithm. Fixes #254.
+* Fix `mbedtls_asn1_get_int` to support any number of leading zeros.
+* Fix `mbedtls_asn1_get_bitstring_null` to correctly parse bitstrings of at most 2 bytes.
+
+### Performance improvements
+
+* Remove a useless call to mbedtls_ecp_group_free(). Contributed by Alexander Krizhanovsky in #210.
+* Speed up PBKDF2 by caching the digest calculation. Contributed by Jack Lloyd and Fortanix Inc in #277.
+
+### Other changes
+
+* Remove the technical possibility to define custom md_info structures, which was exposed only in an internal header.
+
+## Mbed Crypto 2.0.0
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,24 +5,40 @@
 ### Interface changes
 
 * In the PSA API, forbid zero-length keys. To pass a zero-length input to a key derivation function, use a buffer instead (this is now always possible).
+* Rename `psa_asymmetric_sign()` to `psa_sign_hash()` and `psa_asymmetric_verify()` to `psa_verify_hash()`.
+
+### Default behavior changes
+
+* The initial seeding of a CTR\_DRBG instance makes a second call to the entropy function to obtain entropy for a nonce if the entropy size is less than 3/2 times the key size. In case you want to disable the extra call to grab entropy, you can call `mbedtls_ctr_drbg_set_nonce_len()` to force the nonce length to 0.
 
 ### New features
 
 * Key derivation inputs in the PSA API can now either come from a key object or from a buffer regardless of the step type.
+* The CTR_DRBG module can grab a nonce from the entropy source during the initial seeding. The default nonce length is chosen based on the key size to achieve the security strength defined by NIST SP 800-90A. You can change it with `mbedtls_ctr_drbg_set_nonce_len()`.
+* Add ENUMERATED tag support to the ASN.1 module. Contributed by msopiha-linaro in #307.
+
+### Security
+
+* Enforce that `mbedtls_entropy_func()` gathers a total of `MBEDTLS_ENTROPY_BLOCK_SIZE` bytes or more from strong sources. In the default configuration, on a platform with a single entropy source, the entropy module formerly only grabbed 32 bytes, which is good enough for security if the source is genuinely strong, but less than the expected 64 bytes (size of the entropy accumulator).
 
 ### Bug fixes
 
 * Fix a buffer overflow in the PSA HMAC code when using a long key with an unsupported algorithm. Fixes #254.
-* Fix `mbedtls_asn1_get_int` to support any number of leading zeros.
+* Fix `mbedtls_asn1_get_int` to support any number of leading zeros. Credit to OSS-Fuzz for finding a bug in an intermediate version of the fix.
 * Fix `mbedtls_asn1_get_bitstring_null` to correctly parse bitstrings of at most 2 bytes.
+* `mbedtls_ctr_drbg_set_entropy_len()` and `mbedtls_hmac_drbg_set_entropy_len()` now work if you call them before `mbedtls_ctr_drbg_seed()` or `mbedtls_hmac_drbg_seed()`.
+* Fix some false-positive uninitialized variable warnings. Fix contributed by apple-ihack-geek in ARMmbed/mbedtls#2663.
 
 ### Performance improvements
 
 * Remove a useless call to `mbedtls_ecp_group_free()`. Contributed by Alexander Krizhanovsky in #210.
 * Speed up PBKDF2 by caching the digest calculation. Contributed by Jack Lloyd and Fortanix Inc in #277.
+* Small performance improvement of `mbedtls_mpi_div_mpi()`. Contributed by Alexander Krizhanovsky in #308.
 
 ### Other changes
 
 * Remove the technical possibility to define custom `mbedtls_md_info` structures, which was exposed only in an internal header.
+* `psa_close_key(0)` and `psa_destroy_key(0)` now succeed (doing nothing, as before).
+* Variables containing error codes are now initialized to an error code rather than success, so that coding mistakes or memory corruption tends to cause functions to return this error code rather than a success. There are no known instances where this changes the behavior of the library: this is merely a robustness improvement. #323
 
 ## Mbed Crypto 2.0.0
--- a/scripts/assemble_changelog.py
+++ b/scripts/assemble_changelog.py
@@ -28,11 +28,9 @@ import sys
 
 class InputFormatError(Exception):
     def __init__(self, filename, line_number, message, *args, **kwargs):
-        self.filename = filename
-        self.line_number = line_number
-        self.message = message.format(*args, **kwargs)
-    def __str__(self):
-        return '{}:{}: {}'.format(self.filename, self.line_number, self.message)
+        message = '{}:{}: {}'.format(filename, line_number,
+                                     message.format(*args, **kwargs))
+        super().__init__(message)
 
 STANDARD_SECTIONS = (
     b'Interface changes',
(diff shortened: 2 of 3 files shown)


scripts/assemble_changelog.py
```diff
@@ -0,0 +1,241 @@
...
+        options.output = options.input
+    if getattr(options, 'keep_entries', None) is None:
+        options.keep_entries = (output_file is not None)
+
+def main():
+    """Command line entry point."""
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument('--dir', '-d', metavar='DIR',
+                        default='ChangeLog.d',
+                        help='Directory to read entries from (default: ChangeLog.d)')
```
mpg: Typo: `.md`.

ChangeLog.md
```diff
@@ -0,0 +1,28 @@
...
+### Performance improvements
+
+* Remove a useless call to mbedtls_ecp_group_free(). Contributed by Alexander Krizhanovsky in #210.
+* Speed up PBKDF2 by caching the digest calculation. Contributed by Jack Lloyd and Fortanix Inc in #277.
+
+### Other changes
+
+* Remove the technical possibility to define custom md_info structures, which was exposed only in an internal header.
+
+## Mbed Crypto 2.0.0
```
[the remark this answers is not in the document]
gilles-peskine-arm: Yes. I just wrote a few entries as a starting point.

gilles-peskine-arm:
I used bytes to allow non-ASCII characters and be independent of any encoding issues. But now that you mention it, it may be a better idea to impose UTF-8.

I've only tested by visual inspection. The script isn't that complicated, I don't think it needs automated testing. Maybe I could add a runtime check that when a file gets removed, its content has been added to the destination file. That would take care of the most likely failure reason which is that some content gets dropped.

scripts/assemble_changelog.py
```diff
@@ -0,0 +1,241 @@
...
+    b'Removals',
+    b'New features',
+    b'Security',
+    b'Bug fixes',
+    b'Performance improvements',
+    b'Other changes',
+)
+
+class ChangeLog:
+    """An Mbed Crypto changelog.
```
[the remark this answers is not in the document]
gilles-peskine-arm: I prefer not to say “Mbed TLS” here until we do use this script in Mbed TLS.

scripts/assemble_changelog.py
gilles-peskine-arm: No: the directory containing inputs is `.d`, the output file is `.md`.

I know the two names are a bit too close for comfort. I'm open to better naming suggestions.

scripts/assemble_changelog.py
```diff
@@ -0,0 +1,241 @@
...
+        # The output is a non-regular file (e.g. pipe). Write to it directly.
+        output_temp = output_file
+    else:
+        # The output is a regular file. Write to a temporary file,
+        # then move it into place atomically.
+        output_temp = output_file + '.tmp'
+    changelog.write(output_temp)
+    for filename in files_to_remove:
+        sys.stderr.write('Removing ' + filename + '\n')
+        #os.remove(filename)
```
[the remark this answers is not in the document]
gilles-peskine-arm: Fixed.

gilles-peskine-arm:
I added a sanity check that no line is dropped from any of the input files. I propose to stop there for now and add more checks later if we think it's useful.

I also propose to stick to binary for now and possibly switch to utf8 later.

scripts/assemble_changelog.py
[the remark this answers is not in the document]
mpg: I agree, I was rather suggesting "A ChangeLog" (the project name would be implicit from the repository we got the script from).

scripts/assemble_changelog.py
mpg: Oh, I missed that. I don't have suggestions right know, I'll come back later if something comes up.

scripts/assemble_changelog.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        with open(filename, 'wb') as out:
+            for line in self.header:
+                out.write(line)
+            for section in self.section_list:
+                lines = self.section_content[section]
+                while lines and not lines[0].strip():
+                    del lines[0]
+                while lines and not lines[-1].strip():
+                    del lines[-1]
```
mpg: Suggestion: I think the above 5 lines could be shortened to
```
lines = [l for self.section_content[section] if l.strip()]
```

Also, I was under the impression that we eliminated blank lines when reading, but I don't have anything against a belt-and-braces approach here.

scripts/assemble_changelog.py
```diff
@@ -0,0 +1,284 @@
...
+    for filename in files_to_merge:
+        with open(filename, 'rb') as input_file:
+            changelog.add_file(input_file)
+    finish_output(changelog, options.output, options.input, files_to_merge)
+    if not options.keep_entries:
+        remove_merged_entries(files_to_merge)
+
+def set_defaults(options):
+    """Add default values for missing options."""
+    output_file = getattr(options, 'output', None)
```
mpg: Minor: It seems to me* that argparse will always populate this field thanks for `default=None`, so that `getattr()` seems redundant here (and three lines below). I generally don't mind playing it extra safe, but in this instance I think `getattr()` hurts readability a bit, so I'd prefer a version without it.

*I tested that by adding `print(options)` on the line above and then invoking the script from the command line with no arguments, the output included the following line:
```
Namespace(dir='ChangeLog.d', input='ChangeLog.md', keep_entries=None, output=None)
```

mpg:
Note: out of curiosity I ran `flake8`, and for information here are the results:
```
scripts/assemble_changelog.py:29:1: E302 expected 2 blank lines, found 1
scripts/assemble_changelog.py:35:1: E302 expected 2 blank lines, found 1
scripts/assemble_changelog.py:53:1: E302 expected 2 blank lines, found 1
scripts/assemble_changelog.py:70:5: E301 expected 1 blank line, found 0
scripts/assemble_changelog.py:125:56: E711 comparison to None should be 'if cond is not None:'
scripts/assemble_changelog.py:160:80: E501 line too long (96 > 79 characters)
scripts/assemble_changelog.py:164:80: E501 line too long (82 > 79 characters)
scripts/assemble_changelog.py:187:1: E302 expected 2 blank lines, found 1
scripts/assemble_changelog.py:206:1: E302 expected 2 blank lines, found 1
scripts/assemble_changelog.py:224:1: E302 expected 2 blank lines, found 1
scripts/assemble_changelog.py:228:1: E302 expected 2 blank lines, found 1
scripts/assemble_changelog.py:249:1: E302 expected 2 blank lines, found 1
scripts/assemble_changelog.py:257:1: E302 expected 2 blank lines, found 1
scripts/assemble_changelog.py:271:80: E501 line too long (87 > 79 characters)
scripts/assemble_changelog.py:274:80: E501 line too long (88 > 79 characters)
scripts/assemble_changelog.py:275:80: E501 line too long (87 > 79 characters)
```

* I checked all the "line too long" ones and I think they're more readable that way than if they were folded.
* I have no strong opinion regarding E302 "expected 2 blank lines", though perhaps there is some value in following pep8 more closely here as it's widely used
* The only thing I think we should really fix is using the `if cond is not None` idiom.

scripts/assemble_changelog.py
gilles-peskine-arm: I made this function a little more robust (and in fact I made it a separate function at all) to make it easier to use this code as a Python module from another script, rather than via the command line.

scripts/assemble_changelog.py
```diff
@@ -0,0 +1,284 @@
...
+    starts a version, and versions are sorted in reverse chronological
+    order. Lines with a level 2 section title must start with '##'.
+
+    Within a version, there are multiple sections, each devoted to a kind
+    of change: bug fix, feature request, etc. Section titles should match
+    entries in STANDARD_SECTIONS exactly.
+
+    Within each section, each separate change should be on a line starting
+    with a '*' bullet. There may be blank lines surrounding titles, but
+    there should not be any blank line inside a section.
```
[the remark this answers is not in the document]
gilles-peskine-arm: In the class description, I prefer to document stable behavior rather than the exact details that may change over time.

scripts/assemble_changelog.py
```diff
@@ -0,0 +1,284 @@
...
+        can be any generator returning the lines to read.
+        """
+        self.header = []
+        self.section_list = []
+        self.section_content = {}
+        self.add_sections(*STANDARD_SECTIONS)
+        self.trailer = []
+        self.read_main_file(input_stream)
+
+    def read_main_file(self, input_stream):
```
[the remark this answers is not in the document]
gilles-peskine-arm: I added a comment. Creating a new section for a new version is outside the scope of this function and I intended it to be in a follow-up pull request which, indeed, adds a command-line option that takes the version number as an argument. This script can be used at any time, not just to make a release.

scripts/assemble_changelog.py
gilles-peskine-arm: The code only strips leading and trailing blank line, not blank lines in the middle. However, since we strip blank lines from the input anyway, that doesn't matter. I'll simplify the code.

scripts/assemble_changelog.py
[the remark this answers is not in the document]
gilles-peskine-arm: I added some logic to create a level-2 section if it seems to be needed. This way, you can run the script with no argument at any time, and it will populate the “Unreleased changes” section, creating it if it doesn't exist. At release time, first run the script, then edit the title “Unreleased changes” to “Mbed Wossname 3.14”. I think it makes more sense this way, but I'm willing to be talked into a different workflow.

scripts/assemble_changelog.py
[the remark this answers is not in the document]
mpg: Ok.

scripts/assemble_changelog.py
mpg: Good point, I'm not sure how I missed that. Glad the simplification applied nonetheless.

scripts/assemble_changelog.py
mpg: Ok.

scripts/assemble_changelog.py
```diff
@@ -96,9 +87,9 @@ def __init__(self, input_stream):
         can be any generator returning the lines to read.
         """
         self.header = []
-        self.section_list = []
-        self.section_content = {}
-        self.add_sections(*STANDARD_SECTIONS)
+        self.section_content = OrderedDict()
+        for section in STANDARD_SECTIONS:
+            self.section_content[section] = []
```
mpg: Note: this could be shortened to 
```
self.section_content = OrderedDict((s, []) for s in STANDARD_SECTIONS)
```
But I leave it up to you to pick/keep the version you prefer.

scripts/assemble_changelog.py
```diff
@@ -100,21 +105,23 @@ def read_main_file(self, input_stream):
...
-                if level_2_seen <= 1:
-                    self.header.append(line)
-                else:
-                    self.trailer.append(line)
             elif level == 3 and level_2_seen == 1:
                 current_section = content
                 self.section_content.setdefault(content, [])
-            elif level_2_seen == 1 and current_section is not None:
-                if line.strip():
+            if level_2_seen == 1 and current_section is not None:
```
mpg: Minor: perhaps a blank line before this `if` would be useful to visually separate the two `if`-`elif` structures.

scripts/assemble_changelog.py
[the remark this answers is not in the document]
mpg: Thanks for adding the comments, I think that's much clearer now.

Regarding the logic, I think what you implemented might be more flexible than what we need (I don't see ourselves running that script too often between releases), but this flexibility comes as no cost in terms of code complexity or usage complexity, and might turn out to be needed contrary to my expectations, so I'm happy with it.

scripts/assemble_changelog.py
```diff
@@ -0,0 +1,314 @@
...
+        without leading or trailing whitespace. For a non-title line,
+        the level is 0.
+        """
+        level = re.match(self._title_re, line).end()
+        return level, line[level:].strip()
+
+    # Only accept dotted version numbers (e.g. "3.1", not "3").
+    # Refuse ".x" in a version number: this indicates a version that is
+    # not yet released.
+    _version_number_re = re.compile(br'[0-9]\.[0-9][0-9.]+([^.]|\.[^0-9x])')
```
mpg: Sorry, I'm looking at this again because I noticed we sometimes us `X` rather than `x` for marking unreleased version and I wanted to check if this was handled correctly, and I think there's a problem with this regex and the following function: the regex will happily match 2.20.x, 2.20.X, 2.20.Y so these will be considered "released" by the following function. Which I think is not at all what we want.

(Took me some time to understand why the regex matches `2.20.x` while the last part seems to explicitly exclude `x`, then I understood the culprit is the `[0-9.]+` with `.` included, which matches `20.`, and then the `[^.]` will accept the `x`.)

gilles-peskine-arm: Fixed, with two regexps rather than attempting to encode a negative condition in a regexp.

gilles-peskine-arm:
This has been approved. CI is irrelevant since the PR doesn't touch any file involved in the CI. So I'm merging.

The next step will happen in the mbedtls repository.

PR #291, opened 2019-10-07
Test MBEDTLS_CTR_DRBG_USE_128_BIT_KEY

gilles-peskine-arm:
Test `MBEDTLS_CTR_DRBG_USE_128_BIT_KEY` properly. Fix #289. Fix #290.

Follow-up of #287.

Needs backports: to 2.16 (almost identical); to 2.7 (only for the `config.h` and `all.sh` changes that are about `MBEDTLS_ENTROPY_FORCE_SHA256`).

PR #305, opened 2019-10-22
CTR_DRBG: grab a nonce from the entropy source if needed

gilles-peskine-arm:
CTR_DRBG grabs as many bytes of entropy for the initial seeding as for reseeds. This is not compliant with NIST SP 800-90A when the entropy length is too small to cover both the entropy itself and a nonce, unless the caller passes a nonce in the `custom` parameter.

Fix this by making a second call to the entropy function to grab enough entropy for a random nonce (half the entropy length, per SP 800-90A).

To maximize backward compatibility, do not make a second call to the entropy function if the entropy length is sufficient to cover a nonce. Thus the only combination of boolean compile-time option that leads to making a second call to the entropy function is:
```
!MBEDTLS_CTR_DRBG_USE_AES_128 && (!MBEDTLS_SHA512_C || MBEDTLS_ENTROPY_FORCE_SHA256)
```
An alternative design choice would be to always grab a nonce whose length is half the entropy length, like the HMAC_DRBG module does. This would be a simpler design but I preferred this solution which only adds a few bytes of code and which has less risk of breaking someone's workflow.

Test run on a merge of this PR and #291 to get coverage of relevant compilation settings that are currently not covered: https://jenkins-internal.mbed.com/job/mbedtls-psa-release-new/592/ → passed except for some Mbed OS jobs which are also failing in this PR.

Some commits in this PR are minor bugs which I've fixed in the backports of #299 (https://github.com/ARMmbed/mbedtls/pull/2894 and https://github.com/ARMmbed/mbedtls/pull/2895):
* The fixup commits for “xxx_DRBG: support set_entropy_len() before seed()” in this PR are documentation updates that I missed in the last commit of #299. The ones in `ctr_drbg.h` and `hmac_drbg.h` needs to be backported, and I did so in the backport PRs  which have not been merged yet. The one in `ctr_drbg.c` was already present in a suitably modified form in the 2.16 backport of #299 and is not applicable to 2.7.
* "Note that mbedtls_ctr_drbg_seed() must not be called twice" is a follow-up to #287 which I added to the backports of #299.
* "Fix CTR_DRBG benchmark" fixes a bug that this PR revealed. I backported this fix in the backports of #299 and side-ported it in the Mbed OS benchmark in https://github.com/ARMmbed/mbed-os-example-tls/pull/262.

The rest of the PR is a behavior change that should not be backported, and some very minor test improvements that do not need to be backported.

Internal ref: IOTCRYPT-180

### DIFF
--- a/library/ctr_drbg.c
+++ b/library/ctr_drbg.c
@@ -732,6 +732,7 @@ int mbedtls_ctr_drbg_self_test( int verbose )
 
     test_offset = 0;
     mbedtls_ctr_drbg_set_entropy_len( &ctx, 32 );
+    mbedtls_ctr_drbg_set_nonce_len( &ctx, 0 );
     CHK( mbedtls_ctr_drbg_seed( &ctx,
                                 ctr_drbg_self_test_entropy,
                                 (void *) entropy_source_pr,
@@ -756,6 +757,7 @@ int mbedtls_ctr_drbg_self_test( int verbose )
 
     test_offset = 0;
     mbedtls_ctr_drbg_set_entropy_len( &ctx, 32 );
+    mbedtls_ctr_drbg_set_nonce_len( &ctx, 0 );
     CHK( mbedtls_ctr_drbg_seed( &ctx,
                                 ctr_drbg_self_test_entropy,
                                 (void *) entropy_source_nopr,
--- a/library/ctr_drbg.c
+++ b/library/ctr_drbg.c
@@ -56,6 +56,9 @@
 void mbedtls_ctr_drbg_init( mbedtls_ctr_drbg_context *ctx )
 {
     memset( ctx, 0, sizeof( mbedtls_ctr_drbg_context ) );
+    /* Indicate that the entropy nonce length is not set explicitly.
+     * See mbedtls_ctr_drbg_set_nonce_len(). */
+    ctx->reseed_counter = -1;
 
 #if defined(MBEDTLS_THREADING_C)
     mbedtls_mutex_init( &ctx->mutex );
@@ -419,6 +422,19 @@ int mbedtls_ctr_drbg_reseed( mbedtls_ctr_drbg_context *ctx,
     return( mbedtls_ctr_drbg_reseed_internal( ctx, additional, len, 0 ) );
 }
 
+/* Return a "good" nonce length for CTR_DRBG. The chosen nonce length
+ * is sufficient to achieve the maximum security strength given the key
+ * size and entropy length. If there is enough entropy in the initial
+ * call to the entropy function to serve as both the entropy input and
+ * the nonce, don't make a second call to get a nonce. */
+static size_t good_nonce_len( size_t entropy_len )
+{
+    if( entropy_len >= MBEDTLS_CTR_DRBG_KEYSIZE * 3 / 2 )
+        return( 0 );
+    else
+        return( ( entropy_len + 1 ) / 2 );
+}
+
 /* CTR_DRBG_Instantiate with derivation function (SP 800-90A &sect;10.2.1.3.2)
  * mbedtls_ctr_drbg_seed(ctx, f_entropy, p_entropy, custom, len)
  * implements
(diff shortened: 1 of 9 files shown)


tests/suites/test_suite_ctr_drbg.function
```diff
@@ -187,17 +188,47 @@ void ctr_drbg_validate_reseed_first( data_t * add_init, data_t * entropy,
...
+    test_offset_idx = 0;
+    test_max_idx = sizeof( entropy );
+    memset( entropy, 0, sizeof( entropy ) );
 
+    /* The initial seeding must grab at least byte_strength bytes of entropy
+     * for the entropy input and byte_strength/2 bytes for a nonce. */
+    TEST_ASSERT( mbedtls_ctr_drbg_seed( &ctx,
+                                        mbedtls_test_entropy_func, entropy,
+                                        NULL, 0 ) == 0 );
+    TEST_ASSERT( test_offset_idx >= ( byte_strength * 3 + 1 ) / 2 );
```
k-stachowiak: `test_offset_idx` doesn't seem to be updated anywhere in this test case. Therefore, neither this, nor the last assertion operate on anything depending on the runtime other than the function argument.

library/ctr_drbg.c
```diff
@@ -357,10 +357,10 @@ void mbedtls_ctr_drbg_update( mbedtls_ctr_drbg_context *ctx,
  * and with output
  *   ctx contains new_working_state
  */
-int mbedtls_ctr_drbg_reseed_internal( mbedtls_ctr_drbg_context *ctx,
-                                      const unsigned char *additional,
-                                      size_t len,
-                                      size_t nonce_len )
+static int mbedtls_ctr_drbg_reseed_internal( mbedtls_ctr_drbg_context *ctx,
```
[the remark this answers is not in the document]
gilles-peskine-arm: It's an internal function, so it doesn't _need_ an `mbedtls_xxx` prefix, but I think using the prefix makes the code easier to read. For example you can search for `mbedtls_ctr_drbg_reseed` (in the file, or in `git grep`) and it'll find `mbedtls_ctr_drbg_reseed_internal`.

tests/suites/test_suite_ctr_drbg.function
gilles-peskine-arm: It's a global variable. `mbedtls_test_entropy_func` updates it.

gilles-peskine-arm:
Looking at the test results, many Mbed OS jobs are failing. This isn't just known bugs or ordinary flakiness of some boards on the CI: all the benchmark jobs are failing on K64F with an error that's directly related to this PR (https://jenkins-internal.mbed.com/job/mbed-crypto-pr/job/PR-305-merge/4/execution/node/3923/log/), and all the CY8CKIT jobs are failing (not just the known issue with IAR) for a reason I don't understand.

gilles-peskine-arm:
CI failures:

* The benchmark jobs are failing on Mbed OS. This is because the Mbed OS benchmark application has the same bug that I fixed in this PR. https://github.com/ARMmbed/mbed-os-example-tls/pull/262 fixes it (CI run: https://jenkins-internal.mbed.com/job/mbedtls-psa-release-new/599/ → benchmark passes everywhere, but some runs of hashing and authcrypt are failing on CY8CKIT_062_WIFI_BT).
* `getting-started-CY8CKIT_062_WIFI_BT-IAR` fails. This is a known issue that existed before this PR.
* Some other `CY8CKIT_062_WIFI_BT` jobs are failing, but only on [-merge](https://jenkins-internal.mbed.com/blue/organizations/jenkins/mbed-crypto-pr/detail/PR-305-merge/6/pipeline), not on [-head](https://jenkins-internal.mbed.com/blue/organizations/jenkins/mbed-crypto-pr/detail/PR-305-head/5/pipeline). I don't understand what's going on there.

gilles-peskine-arm:
The non-reproducible CY8CKIT_062_WIFI_BT failures look like one board is non-responsive and other boards are working. There are enough passes to show that this PR is ok.

yanesca:
@gilles-peskine-arm What is the meaning of the `fixup!` prefix in the commit messages? (I just find it confusing, because I usually mark commits like that if I later want to rebase and fixup some other commits with them.)

gilles-peskine-arm:
@yanesca It's a git convention and feature indicating that the commit is an amendment to the previous commit with the same title minus the `fixup! ` prefix. This convention is built into the [`--autosquash` option of `git rebase`](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash).

Normally a fixup is only in work in progress and shouldn't be pushed: you should rebase and squash the fixup before pushing. But here the defective commit was already present in the history, so the amendment can't be squashed.

PR #307, opened 2019-10-29
Add ASN.1 ENUMERATED tag support

msopiha-linaro:
Add support for ASN.1 ENUMERATED tag.

### DIFF
--- a/include/mbedtls/asn1.h
+++ b/include/mbedtls/asn1.h
@@ -75,6 +75,7 @@
 #define MBEDTLS_ASN1_OCTET_STRING            0x04
 #define MBEDTLS_ASN1_NULL                    0x05
 #define MBEDTLS_ASN1_OID                     0x06
+#define MBEDTLS_ASN1_ENUMERATED              0x0A
 #define MBEDTLS_ASN1_UTF8_STRING             0x0C
 #define MBEDTLS_ASN1_SEQUENCE                0x10
 #define MBEDTLS_ASN1_SET                     0x11
@@ -254,12 +255,31 @@ int mbedtls_asn1_get_bool( unsigned char **p,
  *              a valid ASN.1 INTEGER.
  * \return      #MBEDTLS_ERR_ASN1_INVALID_LENGTH if the parsed value does
  *              not fit in an \c int.
- * \return      An ASN.1 error code if the input does not start with
- *              a valid ASN.1 INTEGER.
  */
 int mbedtls_asn1_get_int( unsigned char **p,
                           const unsigned char *end,
                           int *val );
+
+/**
+ * \brief       Retrieve an enumerated ASN.1 tag and its value.
+ *              Updates the pointer to immediately behind the full tag.
+ *
+ * \param p     On entry, \c *p points to the start of the ASN.1 element.
+ *              On successful completion, \c *p points to the first byte
+ *              beyond the ASN.1 element.
+ *              On error, the value of \c *p is undefined.
+ * \param end   End of data.
+ * \param val   On success, the parsed value.
+ *
+ * \return      0 if successful.
+ * \return      An ASN.1 error code if the input does not start with
+ *              a valid ASN.1 ENUMERATED.
+ * \return      #MBEDTLS_ERR_ASN1_INVALID_LENGTH if the parsed value does
+ *              not fit in an \c int.
+ */
+int mbedtls_asn1_get_enum( unsigned char **p,
+                           const unsigned char *end,
+                           int *val );
 
 /**
  * \brief       Retrieve a bitstring ASN.1 tag and its value.
@@ -367,8 +387,6 @@ int mbedtls_asn1_get_sequence_of( unsigned char **p,
  * \return      #MBEDTLS_ERR_ASN1_INVALID_LENGTH if the parsed value does
  *              not fit in an \c int.
  * \return      An MPI error code if the parsed value is too large.
- * \return      An ASN.1 error code if the input does not start with
- *              a valid ASN.1 INTEGER.
  */
 int mbedtls_asn1_get_mpi( unsigned char **p,
                           const unsigned char *end,
--- a/library/asn1parse.c
+++ b/library/asn1parse.c
@@ -139,17 +139,20 @@ int mbedtls_asn1_get_bool( unsigned char **p,
     return( 0 );
 }
 
-int mbedtls_asn1_get_int( unsigned char **p,
-                  const unsigned char *end,
-                  int *val )
+static int asn1_get_tagged_int( unsigned char **p,
+                                const unsigned char *end,
+                                int tag, int *val )
 {
     int ret;
     size_t len;
 
-    if( ( ret = mbedtls_asn1_get_tag( p, end, &len, MBEDTLS_ASN1_INTEGER ) ) != 0 )
+    if( ( ret = mbedtls_asn1_get_tag( p, end, &len, tag ) ) != 0 )
         return( ret );
 
-    /* len==0 is malformed (0 must be represented as 020100). */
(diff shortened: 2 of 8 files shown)


gilles-peskine-arm:
Thank you for your contribution! Even if the code is very simple, can you please add a few test cases in `tests/suites/test_suite_asn1parse`? We do have a test coverage backlog there (ASN.1 used to be mostly tested via X.509, but nowadays X.509 lives in a separate library), but we prefer to at least test new features.

There's no code to parse ENUMERATED. Would you be able to add that as well? That's not a blocker (we don't require the feature set of asn1parse and asn1write to be identical), but it would be nice.

msopiha-linaro:
I have extended first commit by adding parser for ENUMERATED tag value.
I have also added test cases for writing/parsing enumerated values, based in ASN1_INTEGER test cases.

include/mbedtls/asn1.h
```diff
@@ -261,6 +262,29 @@ int mbedtls_asn1_get_int( unsigned char **p,
                           const unsigned char *end,
                           int *val );
 
+/**
+ * \brief       Retrieve an integer ASN.1 tag and its value.
```
gilles-peskine-arm: Copypasta:
```suggestion
 * \brief       Retrieve an enumerated ASN.1 tag and its value.
```

include/mbedtls/asn1.h
```diff
@@ -261,6 +262,29 @@ int mbedtls_asn1_get_int( unsigned char **p,
...
+ *              On error, the value of \c *p is undefined.
+ * \param end   End of data.
+ * \param val   On success, the parsed value.
+ *
+ * \return      0 if successful.
+ * \return      An ASN.1 error code if the input does not start with
+ *              a valid ASN.1 ENUMERATED.
+ * \return      #MBEDTLS_ERR_ASN1_INVALID_LENGTH if the parsed value does
+ *              not fit in an \c int.
+ * \return      An ASN.1 error code if the input does not start with
```
gilles-peskine-arm: This clause is present twice. That's a preexisting bug in the description of `mbedtls_asn1_get_int`. Can you please remove the second occurrence from the documentation of both functions?

library/asn1parse.c
```diff
@@ -139,17 +139,18 @@ int mbedtls_asn1_get_bool( unsigned char **p,
...
     int ret;
     size_t len;
 
-    if( ( ret = mbedtls_asn1_get_tag( p, end, &len, MBEDTLS_ASN1_INTEGER ) ) != 0 )
+    if( ( ret = mbedtls_asn1_get_tag( p, end, &len, tag ) ) != 0 )
         return( ret );
 
-    /* len==0 is malformed (0 must be represented as 020100). */
+    /* len==0 is malformed (0 must be represented as 020100 for INTEGER,
+       or 0A0100 for ENUMERATED tags */
```
gilles-peskine-arm: Style: comment continuation lines must start with `*` (aligned with the `*` of the opening `/*`), otherwise some of our sanity check scripts that don't perfectly understand C syntax can fail.

library/asn1parse.c
```diff
@@ -180,6 +181,20 @@ int mbedtls_asn1_get_int( unsigned char **p,
     return( 0 );
 }
 
+int mbedtls_asn1_get_int( unsigned char **p,
+                    const unsigned char *end,
+                    int *val )
+{
+    return asn1_get_tagged_int (p, end, MBEDTLS_ASN1_INTEGER, val);
```
gilles-peskine-arm: Style:
```suggestion
    return( asn1_get_tagged_int (p, end, MBEDTLS_ASN1_INTEGER, val) );
```
and likewise for `get_enum` and for the write functions

library/asn1parse.c
```diff
@@ -139,17 +139,18 @@ int mbedtls_asn1_get_bool( unsigned char **p,
     return( 0 );
 }
 
-int mbedtls_asn1_get_int( unsigned char **p,
-                  const unsigned char *end,
-                  int *val )
+static int asn1_get_tagged_int( unsigned char **p,
+                        const unsigned char *end,
```
gilles-peskine-arm: Style: please align the parameters, i.e.
```
static int asn1_get_tagged_int( unsigned char **p,
                                const unsigned char *end,
                                int tag, int *val )
```
We generally don't fix alignment errors after the fact because it's noise in the history, but since you're modifying that line anyway, please fix the indentation.

library/asn1write.c
```diff
@@ -231,7 +231,7 @@ int mbedtls_asn1_write_bool( unsigned char **p, unsigned char *start, int boolea
     return( (int) len );
 }
 
-int mbedtls_asn1_write_int( unsigned char **p, unsigned char *start, int val )
+static int asn1_write_tagged_int (unsigned char **p, unsigned char *start, int val, int tag)
```
gilles-peskine-arm: Style:
```suggestion
static int asn1_write_tagged_int( unsigned char **p, unsigned char *start, int val, int tag )
```

include/mbedtls/asn1.h
msopiha-linaro: I can see this is present not only in mbedtls_asn1_get_int. Should I remove it everywhere in asn1.h?

gilles-peskine-arm: It's present for many functions, but it should only be present once per function.

msopiha-linaro: I understand this. But if you look at mbedtls_asn1_get_bitstring_null(...), mbedtls_asn1_get_sequence_of(...), mbedtls_asn1_get_bool(...), etc. - you will also find multiple includes of return statements. If we are talking about fixing docs for mbedtls_asn1_get_int() - should be all multiple returns fixed as part of this PR, or just leave it as is?

gilles-peskine-arm: It's ok to have multiple `\return` directives for the same function. That's our house style.

The problem with `mbedtls_asn1_get_int`, `mbedtls_asn1_get_mpi` and now `mbedtls_asn1_get_enum` is that they have multiple `\return` directives _with the same content_. That's noise.

(And probably my fault because I revised `asn1.h` recently. But since that's already merged and you've revealed it, please fix it.)

msopiha-linaro: Ah, got it. Was a bit confused because quoted code was not including repeating values. Ok, I will fix this.

include/mbedtls/asn1.h
msopiha-linaro: Done

library/asn1parse.c
msopiha-linaro: Done

library/asn1parse.c
msopiha-linaro: Done.

library/asn1parse.c
msopiha-linaro: Done.

library/asn1write.c
msopiha-linaro: Done.

msopiha-linaro:
I see that some tests are failing, but I don't have access. Is there any info about that?

gilles-peskine-arm:
All the test failures are due to a defective board on our CI. The rest passes and there's enough working coverage to validate this PR.

gilles-peskine-arm:
New full CI run now that the boards are working again (internal link): https://jenkins-internal.mbed.com/job/mbed-crypto-pr/job/PR-307-merge/9/

msopiha-linaro:
Is anything required from me on this PR?

gilles-peskine-arm:
You don't need to do anything else. All pull requests need to be approved by two team members, so the next action on this PR is a second review. If that's an approval, we'll merge the PR.

PR #314, opened 2019-11-05
Fix pk_write with EC key to use a constant size for the private value

gilles-peskine-arm:
When writing a private EC key, use a constant size for the private value, as specified in RFC 5915. Previously, the value was written as an ASN.1 INTEGER, which caused the size of the key to leak about 1 bit of information on average and could cause the value to be 1 byte too large for the output buffer.

These changes were previously applied to Mbed TLS 2.7, 2.16 and development, but got lost during the split of Mbed TLS and Mbed Crypto. This is a straightforward side-port of https://github.com/ARMmbed/mbedtls-restricted/pull/503 and https://github.com/ARMmbed/mbedtls-restricted/pull/508 (private links: these were pull requests for a security fix which was quarantined until release).

Fix ARMmbed/mbedtls#1028

Internal ref: IOTCRYPT-969


gilles-peskine-arm:
New CI run with repaired CI: https://jenkins-internal.mbed.com/job/mbed-crypto-pr/job/PR-314-merge/2/

PR #317, opened 2019-11-07
getting_started: Make it clear that keys are passed in

Patater:
It was not obvious before that `AES_KEY` and `RSA_KEY` were shorthand
for key material. A user copy pasting the code snippet would run into a
compilation error if they didn't realize this. Make it more obvious that
key material must come from somewhere external by making the snippets
which use global keys into functions that take a key as a parameter.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -72,9 +72,10 @@ with other function calls.
 
 This example shows how to import a key:
 ```C
+void import_a_key(const uint8_t *key, size_t key_len)
+{
     psa_status_t status;
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
-    uint8_t data[] = AES_KEY;
     psa_key_handle_t handle;
 
     printf("Import an AES key...\t");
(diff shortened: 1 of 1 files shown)


Patater:
This is a portability enhancement, not a workaround.

Yes, we should use ECDSA here and also in the Mbed OS example. It's not this PR's aim to change this. Another PR should do so.

This PR's aim is to help ensure we test the code snippets in our docs. The only testing our getting started doc currently has is the Mbed OS example. For ease of maintenance, as much as is possible, it's preferred to have parity between what we test and run in the Mbed OS example and what we list as code snippets in our docs. The ideal would be we run our getting started guide as a sort of literate program ala Knuth (which is out of scope for this PR), rather than copy pasting to an Mbed OS example for testing.

gilles-peskine-arm:
No. Adding extra code to declare a non-standard scope variables has no place in example code. It's an additional cognitive burden. The example code needs to be easy to understand, not maximally portable.

Patater:
We need to do both readable and portable. New patch set changes snippets to avoid magic all-caps keys which always cause a compilation error when a user copy pastes. Now, it is more obvious that the key material must be supplied from elsewhere.

Patater:
> Adding extra code to declare a non-standard scope variables has no place in example code

Agreed. Removed `static` from `signature`, which was noise. Key would be OK as static const as that's idiomatic C for constants, but new tact is even better: function parameter for the key to show obviously that your key material should come from elsewhere.

gilles-peskine-arm:
The [corresponding change to the source code of the example](https://github.com/ARMmbed/mbed-os-example-mbed-crypto/pull/49) has been merged. This pull request has been approved, and it only modifies a file that isn't used by CI so CI is irrelevant. Merging.

PR #322, opened 2019-11-15
Update Mbed Crypto with latest Mbed TLS changes as of 2019-11-15

gilles-peskine-arm:
The previous update was https://github.com/ARMmbed/mbed-crypto/pull/223. Salient changes:

* Reconcile `config.py`, in particular with `MBEDTLS_MEMORY_BUFFER_ALLOC_C` disabled in `config.py full`.
* Bring in the test outcome collection code.
* This update encompasses the removal of crypto files from mbedtls. This will make future merges easier.

This is a merge commit (9afbfdc833726a31e5806494c661d6b83dc3a926) and its follow-up.

CI passed and produced the expected outcome file.

Internal ref: IOTCRYPT-974


Patater:
Given approving reviews and no blocking issues, merging. We can further align the TLS and Crypto branches as desired in follow up PRs.

PR #323, opened 2019-11-22
Initialise return values to an error

yanesca:
Initialising the return values to and error is best practice and makes the library more robust.
    
No new error was introduced because we are running out of error codes and we shouldn't be wasting them. 

We already had an error code reserved for generic errors. This PR makes it explicit and uses this error code to initialise return values.

### DIFF
--- a/include/mbedtls/error.h
+++ b/include/mbedtls/error.h
@@ -52,7 +52,7 @@
  * For historical reasons, low-level error codes are divided in even and odd,
  * even codes were assigned first, and -1 is reserved for other errors.
  *
- * Low-level module errors (0x0002-0x007E, 0x0003-0x007F)
+ * Low-level module errors (0x0002-0x007E, 0x0001-0x007F)
  *
  * Module   Nr  Codes assigned
  * ERROR     2         0x006E   0x0001
--- a/include/mbedtls/error.h
+++ b/include/mbedtls/error.h
@@ -86,7 +86,7 @@
  * CHACHA20  3                  0x0051-0x0055
  * POLY1305  3                  0x0057-0x005B
  * CHACHAPOLY 2 0x0054-0x0056
- * PLATFORM  1  0x0070-0x0072
+ * PLATFORM  2  0x0070-0x0072
  *
  * High-level module nr (3 bits - 0x0...-0x7...)
  * Name      ID  Nr of Errors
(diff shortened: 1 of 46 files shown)


yanesca:
CI fails because Mbed OS uses Mbed TLS's `error.h` and should pass once Mbed TLS has been updated.

library/psa_crypto.c
```diff
@@ -351,6 +351,14 @@ static psa_status_t mbedtls_to_psa_error( int ret )
             return( PSA_ERROR_INSUFFICIENT_MEMORY );
         case MBEDTLS_ERR_ECP_HW_ACCEL_FAILED:
             return( PSA_ERROR_HARDWARE_FAILURE );
+        /*
+         * Mbed Crypto return values are initialised to
+         * MBEDTLS_ERR_GENERIC_ERROR.  PSA return valuse are initialised to
+         * PSA_ERROR_CORRUPTION_DETECTED.  We are mapping these to each other
+         * to align this behaviour.
```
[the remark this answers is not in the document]
yanesca: Ok, I have added a new error code.

yanesca:
I have added a new error code for this purpose and simplified the approach as well. The PR is ready for review again.

gilles-peskine-arm:
There are many CI failures. I don't think this PR is ready.

yanesca:
CI fails because Mbed OS uses Mbed TLS's error.h and should pass once Mbed TLS has been updated.

yanesca:
In retrospect I probably should have started with the Mbed TLS PR, but now I would prefer to continue with the review of this one and raise the Mbed TLS counterpart when this is approved. (I would like to do them sequentially to minimise duplicating work during review.)

gilles-peskine-arm:
Ok. Reviewing.

include/mbedtls/error.h
```diff
@@ -55,6 +55,7 @@
  * Low-level module errors (0x0002-0x007E, 0x0003-0x007F)
```
AndrzejKurek: Please update this information.

include/mbedtls/error.h
```diff
@@ -52,9 +52,10 @@
  * For historical reasons, low-level error codes are divided in even and odd,
  * even codes were assigned first, and -1 is reserved for other errors.
  *
- * Low-level module errors (0x0002-0x007E, 0x0003-0x007F)
+ * Low-level module errors (0x0001-0x007E, 0x0003-0x007F)
```
AndrzejKurek: Shouldn't it rather be (even, odd)?
```suggestion
 * Low-level module errors (0x0002-0x007E, 0x0001-0x007F)
```

yanesca: Indeed, my bad.

PR #324, opened 2019-11-22
test_psa_constant_names: support key agreement, better code structure

gilles-peskine-arm:
This pull request consists mostly of changes to `test_psa_constant_names.py`:

* Many small refactorings of `test_psa_constant_names.py`. I initially did them in preparation for reusing major parts of the code in another script. This other script is not stable yet and I'm not sure if I need it after all. Since the refactorings are at worst harmless and at best useful to make the code easier to maintain, I'm submitting them.
* Fix the collection of curves, which I discovered to be broken.
* Error out if a test case uses a macro name which isn't defined in a header covered by this script.
* Add support for key agreement, which was long overdue.
* In `test_suite_psa_crypto_metadata`, add a few more test cases for KDF.

### DIFF
--- a/tests/scripts/test_psa_constant_names.py
+++ b/tests/scripts/test_psa_constant_names.py
@@ -8,6 +8,7 @@ or 1 (with a Python backtrace) if there was an operational error.
 """
 
 import argparse
+from collections import namedtuple
 import itertools
 import os
 import platform
@@ -341,6 +342,9 @@ def collect_values(inputs, type_word, include_path=None, keep_c=False):
 class Tests:
     """An object representing tests and their results."""
 
+    Error = namedtuple('Error',
+                       ['type', 'expression', 'value', 'output'])
+
     def __init__(self, options):
         self.options = options
         self.count = 0
@@ -362,7 +366,10 @@ class Tests:
         self.count += len(expressions)
         for expr, value, output in zip(expressions, values, outputs):
             if normalize(expr) != normalize(output):
-                self.errors.append((type_word, expr, value, output))
+                self.errors.append(self.Error(type=type_word,
+                                              expression=expr,
+                                              value=value,
+                                              output=output))
 
     def run_all(self, inputs):
         """Run psa_constant_names on all the gathered inputs."""
@@ -376,9 +383,10 @@ class Tests:
         Write the errors to ``out``.
         Also write a total.
         """
-        for type_word, name, value, output in self.errors:
+        for error in self.errors:
             out.write('For {} "{}", got "{}" (value: {})\n'
-                      .format(type_word, name, output, value))
+                      .format(error.type, error.expression,
+                              error.output, error.value))
         out.write('{} test cases'.format(self.count))
         if self.errors:
             out.write(', {} FAIL\n'.format(len(self.errors)))
--- a/tests/scripts/test_psa_constant_names.py
+++ b/tests/scripts/test_psa_constant_names.py
@@ -61,6 +61,7 @@ class read_file_lines:
 
 class Inputs:
     """Accumulate information about macros to test.
+
     This includes macro names as well as information about their arguments
     when applicable.
     """
@@ -101,6 +102,7 @@ class Inputs:
 
     def gather_arguments(self):
         """Populate the list of values for macro arguments.
+
         Call this after parsing all the inputs.
         """
         self.arguments_for['hash_alg'] = sorted(self.hash_algorithms)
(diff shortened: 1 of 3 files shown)


tests/scripts/test_psa_constant_names.py
```diff
@@ -348,13 +359,9 @@ def run_tests(options, inputs):
...
-    for type_word, names in [('status', inputs.statuses),
-                             ('algorithm', inputs.algorithms),
-                             ('ecc_curve', inputs.ecc_curves),
-                             ('dh_group', inputs.dh_groups),
-                             ('key_type', inputs.key_types),
-                             ('key_usage', inputs.key_usage_flags)]:
-        c, e = do_test(options, inputs, type_word, names)
+    for type_word in ['status', 'algorithm', 'ecc_curve', 'dh_group',
+                      'key_type', 'key_usage']:
+        c, e = do_test(options, inputs, type_word)
```
Patater: In commit "Move the type_word->name_set mapping into its own method", which says it has no change on behavior, we don't need to call `get_names()` anywhere? For example, while passing a names parameter to `do_test()`

tests/scripts/test_psa_constant_names.py
```diff
@@ -324,7 +324,18 @@ def normalize(expr):
...
+
+    Return a list of pairs of (expr, value) where expr is an expression and
+    value is a string representation of its integer value.
+    """
+    names = inputs.get_names(type_word)
+    expressions = sorted(inputs.generate_expressions(names))
+    values = run_c(options, type_word, expressions)
+    return expressions, values
+
+def do_test(options, inputs, type_word):
```
Patater: In commit "Move value collection into its own function", which says it makes no change to behavior, don't we need to update all callers of `do_test()`? We are changing the parameters `do_test()` makes in a commit that doesn't change the callers, so I don't see how this wouldn't change behavior.

gilles-peskine-arm:
I rewrote the history in two places:

* I fixed the split between "Move the type_word->name_set mapping into its own method" and "Move value collection into its own function" to make each commit correct as pointed out in the review.
* I modified the commit "add_test_case_line: data-driven dispatch" significantly to better separate data and code and to avoid reaching pylint's threshold for too many branches when the kdf/ka cases are added.

tests/scripts/test_psa_constant_names.py
```diff
@@ -200,26 +229,35 @@ def parse_header(self, filename):
...
-            self.dh_groups.add(argument)
+            sets.append(self.algorithms)
+            if function == 'key_agreement_algorithm' and \
+               argument.startswith('PSA_ALG_KEY_AGREEMENT('):
+                # We only want *raw* key agreement algorithms as such, so
+                # exclude ones that are already chained with a KDF.
+                # Keep the expression as one to test as an algorithm.
+                function = 'other_algorithm'
+        if function in self.table_by_test_function:
+            sets.append(self.table_by_test_function[function])
```
Patater: Do we want to call `accept_test_case_line()` if `function` was not in `self.table_by_test_function`? I see `accept_test_case_line()` throws an exception instead of returning `False` for not accepted test cases.

tests/scripts/test_psa_constant_names.py
```diff
@@ -200,26 +229,35 @@ def parse_header(self, filename):
...
+    def generate_undeclared_names(self, expr):
+        for name in re.findall(self._macro_identifier_re, expr):
+            if name not in self.all_declared:
+                yield name
+
+    def accept_test_case_line(self, function, argument):
+        #pylint: disable=unused-argument
+        undeclared = list(self.generate_undeclared_names(argument))
+        if undeclared:
+            raise Exception('Undeclared names in test case', undeclared)
```
Patater: Should we return `False` instead of throwing an exception?

gilles-peskine-arm: The exception here is intended. This only comes up if there's a test case with data that is not declared in a header. If this ever happens, it probably indicates that the test script need to be updated to add a missing header, or that there's a mistake in the test script logic, or that the intent of the test suite has changed in a way that doesn't agree with the test script. Whatever the reason, it's best that we know that something weird is going on, hence the test failure.

tests/scripts/test_psa_constant_names.py
gilles-peskine-arm: Good question. If we add a function to `test_suite_psa_crypto_metadata`, what do we want to happen here? I don't think the current behavior is correct: a new function whose name does not end in `_algorithm` would undergo the argument validity check without contributing to the data tested by this script. I'll change the code to insist on the function being known: better fail with an exception than silently not test something that we might reasonably believe is tested.

gilles-peskine-arm: I've changed the code to insist on knowing the function name. This uncovered a lack of coverage for some expressions involving a wildcard.

PR #325, opened 2019-11-26
Rename psa_asymmetric_{sign_verify} to psa_{sign,verify}_hash

gilles-peskine-arm:
Rename `psa_asymmetric_sign` to `psa_sign_hash` to make it clearer that this function takes a hash as its input (or something that serves as a hash). Likewise for `psa_asymmetric_verify` to `psa_verify_hash`. Rename some macros accordingly. A follow-up will add a new pair of functions `psa_sign_message` and `psa_verify_message`.

Keep deprecated aliases. Improve the testing of deprecated features: do test some deprecated features when `MBEDTLS_DEPRECATED_REMOVED` and `MBEDTLS_DEPRECATED_WARNING` are both disabled, and add a test component that builds and runs with those tests and `MBEDTLS_DEPRECATED_WARNING` enabled and with `-Werror=deprecated-declarations` disabled.

This enacts a change in the PSA specification (private link: https://github.com/ARMmbed/psa-crypto/pull/301). Internal ref: IOTCRYPT-933


gilles-peskine-arm:
CI passed except for expected failures: an ABI change (two external symbols have been renamed) and generated files in TLS (due to the new header).

include/psa/crypto_compat.h
```diff
@@ -0,0 +1,102 @@
+/**
+ * \file psa/crypto_compat.h
+ *
+ * \brief PSA cryptography module: Backward compatibility aliases
```
AndrzejKurek: I would add a sentence about this file also having deprecation information.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -3700,8 +3730,14 @@ void signature_size( int type_arg,
 {
     psa_key_type_t type = type_arg;
     psa_algorithm_t alg = alg_arg;
-    size_t actual_size = PSA_ASYMMETRIC_SIGN_OUTPUT_SIZE( type, bits, alg );
+    size_t actual_size = PSA_SIGN_OUTPUT_SIZE( type, bits, alg );
+
     TEST_EQUAL( actual_size, (size_t) expected_size_arg );
+#if defined(MBEDTLS_TEST_DEPRECATED)
+    TEST_EQUAL( actual_size,
+                PSA_ASYMMETRIC_SIGN_OUTPUT_SIZE( type, bits, alg ) );
```
gilles-peskine-arm: This is conditioned by `MBEDTLS_TEST_DEPRECATED`. Using a deprecated macro outside of a `MBEDTLS_TEST_DEPRECATED` guard should result in a CI failure in the job with `MBEDTLS_DEPRECATED_WARNING` set.

include/psa/crypto_compat.h
gilles-peskine-arm: What do you mean? Something like “This header declares alternative names for macros and functions that may be removed in a future version of Mbed Crypto”?

tests/suites/test_suite_psa_crypto.function
```diff
@@ -3769,6 +3769,15 @@ void sign_deterministic( int key_type_arg, data_t *key_data,
     ASSERT_COMPARE( output_data->x, output_data->len,
                     signature, signature_length );
 
+#if defined(MBEDTLS_TEST_DEPRECATED)
+    PSA_ASSERT( psa_asymmetric_sign( handle, alg,
```
AndrzejKurek: I was thinking about a potential gap in testing deprecated stuff here:
If `psa_asymmetric_sign` somehow happens to return `PSA_SUCCESS` on error, but doesn't modify `signature` in the process, the next `ASSERT_COMPARE` will succeed anyway, since there's old data in the buffer. 
Not sure if it's worth to add buffer zeroization here, since `psa_asymmetric_sign` should not behave badly.

gilles-peskine-arm: I don't think it's useful to test deprecated identifiers as black boxes. My intent was a gray-box test, where we assume that the deprecated identifiers are defined as aliases but they may be defined as the wrong alias, or pass the wrong parameter, or have syntax errors in a macro expansion. Zeroing `signature` here is cheap and useful so I'll do it.

include/psa/crypto_compat.h
AndrzejKurek: Or just `Backward compatibility aliases and deprecated entities`

gilles-peskine-arm:
Again CI passed except for expected failures: an ABI change (two external symbols have been renamed) and generated files in TLS (due to the new header).

PR #326, opened 2019-11-28
Buffer Alloc: allow allocating zero length buffers

yanesca:
The allocator returned NULL pointer when trying to allocate zero length
buffers. Most memory allocators allow it and applications might expect a
valid pointer even when allocating zero bytes.

### DIFF
--- a/library/memory_buffer_alloc.c
+++ b/library/memory_buffer_alloc.c
@@ -241,9 +241,15 @@ static void *buffer_alloc_calloc( size_t n, size_t size )
     if( heap.buf == NULL || heap.first == NULL )
         return( NULL );
 
+    if( n == 0 || size == 0 )
+    {
+        n = 1;
+        size = 1;
+    }
+
     original_len = len = n * size;
 
-    if( n == 0 || size == 0 || len / n != size )
+    if( len / n != size )
         return( NULL );
     else if( len > (size_t)-MBEDTLS_MEMORY_ALIGN_MULTIPLE )
         return( NULL );
--- a/tests/suites/test_suite_memory_buffer_alloc.function
+++ b/tests/suites/test_suite_memory_buffer_alloc.function
@@ -30,6 +30,46 @@ void mbedtls_memory_buffer_alloc_self_test(  )
 /* END_CASE */
 
 /* BEGIN_CASE */
+void memory_buffer_alloc_free_zero( )
+{
+    unsigned char buf[1024];
+    unsigned char *ptr_a = NULL, *ptr_b = NULL, *ptr_c = NULL;
+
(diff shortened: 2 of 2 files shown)


Patater:
We depend on our allocator returning NULL for zero-length buffers in order to get test coverage on allocators that can return NULL for zero-length buffers. glibc doesn't usually do this, but the man page on my system indicates that getting a NULL back is allowed. Without our allocator returning NULL for zero-length buffers, we lose test coverage with that type of allocator.

On my Linux system, I have this in `man malloc`:
```
If size is  0,  then malloc() returns either NULL, or a unique pointer value that can later be successfully passed to free().
```

yanesca:
On, linux I get the same, I made the mistake of checking the man on OS X.

gilles-peskine-arm:
The library is supposed to support `malloc(0)` returning `NULL`. Currently, in Mbed TLS, we only test it via `MEMORY_BUFFER_ALLOC`, and that makes us miss some issues. In Mbed Crypto we also have [an `all.sh` component](https://github.com/ARMmbed/mbed-crypto/pull/264/files#diff-d36612fab4653d3f01f967a3c1bfc170) that catches more issues thanks to ASan+UBSan.

athoelke:
If the platform's implementation of `malloc()` is also used to provide storage for C++ `operator new()` then it will return distinct non-null pointers when asked to allocate zero bytes (as that is a C++ spec requirement). It is certainly not something C programs should expect from `malloc()`.

gilles-peskine-arm:
@athoelke C allows both (unchanged between C89 and C11):

> If the size of the space requested is zero, the behavior is implementation-defined: either a null pointer is returned, or the behavior is as if the size were some nonzero value, except that the returned pointer shall not be used to access an object.

Most implementations choose to handle `malloc(0)` as `malloc(1)`, partly because it's better for `realloc(p, 0)` to return non-NULL since otherwise you can't tell whether `p` has been freed. And `malloc(0)` returning NULL also makes it hard to tell whether `malloc` has succeeded. But that's not an obligation. Our own allocator only has `calloc`, not `realloc`. C++ is not a compelling argument either since if you have a C++ compiler, you presumably also have a C++ runtime, and you presumably also have a C runtime that includes malloc.

Platforms where `malloc(0)` returns `NULL` include some old Unix systems (mentioned in the [Cerberus survey](https://www.cl.cam.ac.uk/~pes20/cerberus/analysis-2016-02-05-anon.txt)) and IIRC CompCert.

Patater:
https://wiki.sei.cmu.edu/confluence/display/c/MEM04-C.+Beware+of+zero-length+allocations

PR #329, opened 2019-11-29
Add support for building MSVC shared libraries

Xaymar:
Adds support for building and importing MSVC dynamic libraries via generated .lib files. This is part 1 of the fix for (https://github.com/ARMmbed/mbedtls/issues/470). Due to the mixing of headers and source files with no real differentiation between projects this turned out to be much harder than expected, and may not work perfectly yet.

Needs testing across many compilers, especially embedded device compilers.

### DIFF
--- a/3rdparty/everest/include/everest/everest.h
+++ b/3rdparty/everest/include/everest/everest.h
@@ -22,6 +22,8 @@
 #ifndef MBEDTLS_EVEREST_H
 #define MBEDTLS_EVEREST_H
 
+#include "mbedtls/export.h"
+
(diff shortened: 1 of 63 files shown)


gilles-peskine-arm:
Thank you for working on this! Given that [Mbed Crypto has been merged back into Mbed TLS](https://github.com/ARMmbed/mbed-crypto/issues/374), we are no longer making changes to build scripts in Mbed Crypto, so this pull request needs to be wrapped into https://github.com/ARMmbed/mbedtls/pull/2943.

PR #334, opened 2019-12-19
Fix some pylint warnings

dgreen-arm:
This PR fixes a few warnings that the check_python_files component of all.sh was complaining about

### DIFF
--- a/tests/scripts/check-test-cases.py
+++ b/tests/scripts/check-test-cases.py
@@ -26,6 +26,7 @@ import re
 import sys
 
 class Results:
+    """Store file and line information about errors or warnings in test suites."""
     def __init__(self):
         self.errors = 0
         self.warnings = 0
@@ -41,6 +42,7 @@ class Results:
         self.warnings += 1
 
 def collect_test_directories():
+    """Get the relative path for the TLS and Crypto test directories."""
     if os.path.isdir('tests'):
         tests_dir = 'tests'
     elif os.path.isdir('suites'):
@@ -55,6 +57,7 @@ def collect_test_directories():
     return directories
 
 def check_description(results, seen, file_name, line_number, description):
+    """Check test case descriptions for errors."""
     if description in seen:
         results.error(file_name, line_number,
                       'Duplicate description (also line {})',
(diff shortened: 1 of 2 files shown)


tests/scripts/check-test-cases.py
```diff
@@ -26,6 +26,8 @@
 import sys
 
 class Results:
+    """Stores file and line information about errors
```
gilles-peskine-arm: Nitpick: [PEP 257](https://www.python.org/dev/peps/pep-0257) says to write ”Store …“, not “Stores …”, and to end with a period. PEP 257 also says a one-line docstring “should really fit on one line”, and a multi-line docstring should start with a one-line summary.

tests/scripts/mbedtls_test.py
```diff
@@ -310,7 +310,10 @@ def run_test(self, name, function_id, dependencies, args):
 
         param_bytes, length = self.test_vector_to_bytes(function_id,
                                                         dependencies, args)
-        self.send_kv(''.join('{:02x}'.format(x) for x in length), ''.join('{:02x}'.format(x) for x in param_bytes))
+        self.send_kv(
+            ''.join('{:02x}'.format(x) for x in length),
+            ''.join('{:02x}'.format(x) for x in param_bytes)
+        )
```
gilles-peskine-arm: How about
```
params = length + param_bytes
self.send_kv('.join('{:02x}'.format(x) for x in params))
```

dgreen-arm: Wouldn't that send one string, rather than two separate strings?

gilles-peskine-arm: Oh right. Never mind.

tests/scripts/check-test-cases.py
dgreen-arm: I've updated this.

tests/scripts/check-test-cases.py
```diff
@@ -41,6 +42,7 @@ def warning(self, file_name, line_number, fmt, *args):
         self.warnings += 1
 
 def collect_test_directories():
+    """Gets the relative path for the TLS and Crypto test directories."""
```
Patater: Get

tests/scripts/check-test-cases.py
```diff
@@ -55,6 +57,7 @@ def collect_test_directories():
     return directories
 
 def check_description(results, seen, file_name, line_number, description):
+    """Checks test case descriptions for errors."""
```
Patater: Check

Patater:
Mbed OS tests expected to be failing until https://github.com/ARMmbed/mbed-os/pull/12150 is merged

PR #344, opened 2020-01-16
Avoid overwriting tls headers in submodule mode

Ionic:
##### Summary of change (*What the change is for and why*)

When crypto is embedded as a submodule and the cmake build system is used, it would previously overwrite some header files installed by tls.

Most of them are harmless (since they should be identical), but config.h is a special case.

tls's and crypto's config.h files differ widely in scope and overwriting the more general, bigger config.h file from tls with crypto's smaller one will make a lot of features unavailable in programs using tls.

Let's just avoid overwriting any tls header in submodule mode.

Note that this will not fix the potential issue that crypto might be using a different configuration than tls in the submodule case.

##### Documentation (*Details of any document updates required*)

N/A

----------------------------------------------------------------------------------------------------------------
### Pull request type (*required*)

    [X] Patch update (Bug fix / Target update / Docs update / Test update / Refactor)
    [ ] Feature update (New feature / Functionality change / New API)
    [ ] Major update (Breaking change E.g. Return code change / API behaviour change)

----------------------------------------------------------------------------------------------------------------
### Test results (*required*)

    [ ] No Tests required for this change (E.g docs only update)
    [ ] Covered by existing mbed-os tests (Greentea or Unittest)
    [X] Tests / results supplied as part of this PR
    
    
----------------------------------------------------------------------------------------------------------------
### Reviewers (*optional*)

TBD

----------------------------------------------------------------------------------------------------------------
##### Summary of changes

N/A

##### Impact of changes

N/A

##### Migration actions required

N/A

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -5,6 +5,19 @@ if(INSTALL_MBEDTLS_HEADERS)
     file(GLOB headers "mbedtls/*.h")
     file(GLOB psa_headers "psa/*.h")
 
+    if(USE_CRYPTO_SUBMODULE)
+        # Don't overwrite mbedtls's header files!
+        # config.h is supposed to be automatically checked for compatibility
+        # in automatic builds, while the other files should not just be
+        # compatible, but also identical in theory.
+        # Practically, we don't check that in crypto but just assume that the
+        # submodule configuration is sane and trust tls's headers.
+        list(REMOVE_ITEM headers    "${CMAKE_CURRENT_SOURCE_DIR}/mbedtls/compat-1.3.h"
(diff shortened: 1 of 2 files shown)


Ionic:
Needs test case.

Ionic:
Let's see if that works. I'm a bit unsure about the test case and whether it'll do what we need, or whether this stuff should better be part of tls instead of crypto, but oh well.

gilles-peskine-arm:
[Mbed Crypto has been merged back into Mbed TLS](https://github.com/ARMmbed/mbed-crypto/issues/374), therefore this pull request is no longer relevant.

PR #345, opened 2020-01-20
Fix possible error code mangling in psa_mac_verify_finish

gilles-peskine-arm:
If psa_mac_finish_internal fails (which can only happen due to bad
parameters or hardware problem), the error code was converted to
PSA_ERROR_INVALID_SIGNATURE if the uninitialized stack variable
actual_mac happened to contain the expected MAC. This is a minor bug
but it may be possible to leverage it as part of a longer attack path
in some scenarios.

Reported externally. Found by static analysis.


mpg:
Looking at CI results in the pr-merge job, the only failures are with Mbed OS testing, which is expected as Mbed OS needs updating. All other tests pass, so this is as good as a pass.

PR #348, opened 2020-01-21
Bump version to Mbed TLS 2.20.0

yanesca:
Increase the version number to reflect the latest release.

Remarks:
- Although Mbed Crypto has its own version number it is not reported in the code yet. The in-source version still follows the main Mbed TLS version.
- The SO version needed updating as well because of an API break.


yanesca:
CI only fails on Mbed OS and only because a known issue independent from this PR.

PR #349, opened 2020-01-21
Fix minor defects found by Coverity

gilles-peskine-arm:
* One instance of dead code in the product. IOTCRYPT-475
* Some missing return value checks in test code.
* Several cases where test code could cause a resource leak or crash if the test fails in certain ways.
* (Found while I was looking at the code) Make `ASSERT_ALLOC_WEAK` skip the test rather than pass it if it fails, as was intended all along.

The missing return value checks should be backported where applicable. The other bugs are so minor that they don't need backporting (but can be). `ASSERT_ALLOC_WEAK` is not in the LTS branches.

### Backports

- 2.16 https://github.com/ARMmbed/mbedtls/pull/3001
- 2.7 https://github.com/ARMmbed/mbedtls/pull/3002

### DIFF
--- a/library/cipher.c
+++ b/library/cipher.c
@@ -527,6 +527,10 @@ int mbedtls_cipher_update( mbedtls_cipher_context_t *ctx, const unsigned char *i
 
     *olen = 0;
     block_size = mbedtls_cipher_get_block_size( ctx );
+    if ( 0 == block_size )
+    {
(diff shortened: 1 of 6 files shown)


mpg:
The CI passed all tests except the Mbed OS one, which are expected to fail until Mbed OS is updated. So that's as good as a pass.

PR #350, opened 2020-01-21
test_suite_asn1parse: improve testing of trailing garbage in parse_prefixes

gilles-peskine-arm:
Do it in a way that's more systematic and clear. Add comments to document what's going on.


yanesca:
The CI failures are due to a known issue that is unrelated to this PR.

Issue #347, opened 2020-01-20
Plan to address RSA attacks of eprint.iacr.org/2020/055?


jack-fortanix:
https://eprint.iacr.org/2020/055.pdf describes side channel attacks affecting mbedtls ECDSA signing and RSA key loading. It appears the ECDSA signing was already resolved in 247c4d3c8876d but as best I can determine the RSA side channel has not yet been resolved since `mbedtls_rsa_deduce_crt` (https://github.com/ARMmbed/mbed-crypto/blob/development/library/rsa_internal.c#L480) invokes `mbedtls_mpi_inv_mod` (https://github.com/ARMmbed/mbed-crypto/blob/development/library/bignum.c#L2345) which is implemented using an extended binary GCD which leaks information due to conditional branching and which additional invokes binary GCD which is also leaky (https://github.com/ARMmbed/mbed-crypto/blob/development/library/bignum.c#L2252), the second being the attack implemented in 2020/055 though they note that the BEEA implementation can also be used as a source of information.

I wanted to know if you have plans to address these issues and/or would be willing to accept patches addressing them.

For computing Q^-1 mod P following the papers suggested remediation of using a side-channel secured mod-exp and computing Q^(P-2) mod P seems good to me, albeit with some probable runtime overhead.

The paper https://gcd.cr.yp.to/safegcd-20190413.pdf proposes various constant-time algorithms for GCD and extended GCD. In particular the plain GCD algorithm (Figure 1.2 in the paper) is quite easy to implement.

Another option for inversion is the algorithm given in the appendix of https://hal.inria.fr/hal-01506572/document (Algorithm 5) which works for any odd modulus and is quite easily implemented in const time.

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/IOTCRYPT-1040*

yanesca:
@jack-fortanix thank you for reporting this issue!

> I wanted to know if you have plans to address these issues and/or would be willing to accept patches addressing them.

Yes, we would like to fix this issue as soon as possible. We would be very grateful for contributions addressing this issue! If you would like to submit a PR, please let us know in advance, so that we can avoid double work!

Regarding the nature of the fix, I think we should take the performance penalty, the issue only happens on private key import, which usually does not happen very frequently.

jack-fortanix:
@yanesca Glad to hear.

[...]

PR #352, opened 2020-01-22
Parse RSA parameters DP, DQ and QP from PKCS1 private keys

jack-fortanix:
Otherwise these values are recomputed in `mbedtls_rsa_deduce_crt`, which currently suffers from side channel issues in the computation of QP (see https://eprint.iacr.org/2020/055). By loading the pre-computed values not only is the side channel avoided, but runtime overhead of loading RSA keys is reduced.

Discussion in https://github.com/ARMmbed/mbed-crypto/issues/347

### DIFF
--- a/library/pkparse.c
+++ b/library/pkparse.c
@@ -782,25 +782,17 @@ static int pk_parse_key_pkcs1_der( mbedtls_rsa_context *rsa,
     */
 
     /* Import DP */
-    if( ( ret = mbedtls_asn1_get_tag( &p, end, &len,
-                                      MBEDTLS_ASN1_INTEGER ) ) != 0 ||
-        ( ret = mbedtls_mpi_read_binary( &rsa->DP, p, len ) ) != 0 )
-        goto cleanup;
-    p += len;
+    if( ( ret = mbedtls_asn1_get_mpi( &p, end, &rsa->DP ) ) != 0)
+       goto cleanup;
 
     /* Import DQ */
-    if( ( ret = mbedtls_asn1_get_tag( &p, end, &len,
-                                      MBEDTLS_ASN1_INTEGER ) ) != 0 ||
-        ( ret = mbedtls_mpi_read_binary( &rsa->DQ, p, len ) ) != 0 )
-        goto cleanup;
-    p += len;
+    if( ( ret = mbedtls_asn1_get_mpi( &p, end, &rsa->DQ ) ) != 0)
+       goto cleanup;
 
     /* Import QP */
-    if( ( ret = mbedtls_asn1_get_tag( &p, end, &len,
-                                      MBEDTLS_ASN1_INTEGER ) ) != 0 ||
-        ( ret = mbedtls_mpi_read_binary( &rsa->QP, p, len ) ) != 0 )
-        goto cleanup;
-    p += len;
+    if( ( ret = mbedtls_asn1_get_mpi( &p, end, &rsa->QP ) ) != 0)
+       goto cleanup;
+
 #else
     /* Verify existance of the CRT params */
     if( ( ret = mbedtls_asn1_get_mpi( &p, end, &T ) ) != 0 ||
--- a/library/pkparse.c
+++ b/library/pkparse.c
@@ -769,14 +769,29 @@ static int pk_parse_key_pkcs1_der( mbedtls_rsa_context *rsa,
         goto cleanup;
     p += len;
 
+    /* Import DP */
+    if( ( ret = mbedtls_asn1_get_tag( &p, end, &len,
+                                      MBEDTLS_ASN1_INTEGER ) ) != 0 ||
+        ( ret = mbedtls_mpi_read_binary( &rsa->DP, p, len ) ) != 0 )
+        goto cleanup;
+    p += len;
+
+    /* Import DQ */
+    if( ( ret = mbedtls_asn1_get_tag( &p, end, &len,
+                                      MBEDTLS_ASN1_INTEGER ) ) != 0 ||
+        ( ret = mbedtls_mpi_read_binary( &rsa->DQ, p, len ) ) != 0 )
+        goto cleanup;
+    p += len;
+
+    /* Import QP */
+    if( ( ret = mbedtls_asn1_get_tag( &p, end, &len,
+                                      MBEDTLS_ASN1_INTEGER ) ) != 0 ||
+        ( ret = mbedtls_mpi_read_binary( &rsa->QP, p, len ) ) != 0 )
+        goto cleanup;
+    p += len;
+
     /* Complete the RSA private key */
     if( ( ret = mbedtls_rsa_complete( rsa ) ) != 0 )
-        goto cleanup;
-
-    /* Check optional parameters */
-    if( ( ret = mbedtls_asn1_get_mpi( &p, end, &T ) ) != 0 ||
-        ( ret = mbedtls_asn1_get_mpi( &p, end, &T ) ) != 0 ||
-        ( ret = mbedtls_asn1_get_mpi( &p, end, &T ) ) != 0 )
         goto cleanup;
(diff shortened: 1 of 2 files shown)


gilles-peskine-arm:
Thanks for the patch! Unfortunately, the CI points out that it doesn't build when `MBEDTLS_RSA_NO_CRT` is defined.

```
var/lib/build/library/rsa.c: In function 'mbedtls_rsa_complete':
/var/lib/build/library/rsa.c:252:67: error: variable 'have_QP' set but not used [-Werror=unused-but-set-variable]
     int have_N, have_P, have_Q, have_D, have_E, have_DP, have_DQ, have_QP;
                                                                   ^
/var/lib/build/library/rsa.c:252:58: error: variable 'have_DQ' set but not used [-Werror=unused-but-set-variable]
     int have_N, have_P, have_Q, have_D, have_E, have_DP, have_DQ, have_QP;
                                                          ^
/var/lib/build/library/rsa.c:252:49: error: variable 'have_DP' set but not used [-Werror=unused-but-set-variable]
     int have_N, have_P, have_Q, have_D, have_E, have_DP, have_DQ, have_QP;
                                                 ^
cc1: all warnings being treated as errors
library/CMakeFiles/mbedcrypto.dir/build.make:1358: recipe for target 'library/CMakeFiles/mbedcrypto.dir/rsa.c.o' failed
make[2]: *** [library/CMakeFiles/mbedcrypto.dir/rsa.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
CMakeFiles/Makefile2:268: recipe for target 'library/CMakeFiles/mbedcrypto.dir/all' failed
make[1]: *** [library/CMakeFiles/mbedcrypto.dir/all] Error 2
make: *** [all] Error 2
Makefile:138: recipe for target 'all' failed
^^^^test_rsa_no_crt: build: Default + RSA_NO_CRT (ASan build): command make -> 2^^^^
```

To reproduce:
```
make clean
scripts/config.py set MBEDTLS_RSA_NO_CRT
make all test
```

To run all the tests we run with CRT usage disabled:
```
tests/scripts/all.sh test_rsa_no_crt
```

library/pkparse.c
```diff
@@ -769,14 +769,29 @@ static int pk_parse_key_pkcs1_der( mbedtls_rsa_context *rsa,
         goto cleanup;
     p += len;
 
-    /* Complete the RSA private key */
-    if( ( ret = mbedtls_rsa_complete( rsa ) ) != 0 )
+    /* Import DP */
```
yanesca: Could we please add a comment here as a reminder for the future to ensure that nobody "optimises" this code out?

Basically explaining that although we could compute these values, we choose not to because this way it is faster and reduces the attack surface exposed to side channel attacks.

jack-fortanix:
@gilles-peskine-arm I addressed the build problem but it looks like Jenkins is still failing, can you forward the issue here?

`tests/scripts/all.sh test_rsa_no_crt` was able to reproduce the error you were seeing, but after fixing it a later build fails with

```
/home/jack/sw/mbed-crypto/tests/suites/test_suite_psa_crypto.function: In function ‘test_generate_random’:
suites/helpers.function:154:25: error: argument 2 value ‘18446744073709551597’ exceeds maximum object size 9223372036854775807 [-Werror=alloc-size-larger-than=]
             ( pointer ) = mbedtls_calloc( sizeof( *( pointer ) ), \
             ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                           ( length ) );           \
                                           ~~~~~~~~~~~~
/home/jack/sw/mbed-crypto/tests/suites/test_suite_psa_crypto.function:5161:5: note: in expansion of macro ‘ASSERT_ALLOC’
     ASSERT_ALLOC( changed, bytes );
     ^~~~~~~~~~~~
In file included from suites/helpers.function:5:0:
/usr/include/stdlib.h:541:14: note: in a call to allocation function ‘calloc’ declared here
 extern void *calloc (size_t __nmemb, size_t __size)
              ^~~~~~
cc1: all warnings being treated as errors
```

This appears to be a GCM test so I don't think has anything to do with my change - and the same error appears with `development` branch.

yanesca:
Unfortunately `all.sh` is very sensitive to the OS and tools used and can have such problems. As a workaround you could delete the `generate_random` function from `tests/suites/test_suite_psa_crypto.function` and the entries starting with `generate_random` from `tests/suites/test_suite_psa_crypto.data` then `tests/scripts/all.sh test_rsa_no_crt` should be able to reproduce the failure.

For more detailed test output you can run the individual test functions with `-v` option. For example: `cd tests/ && ./test_suite_pk -v`.

jack-fortanix:
@yanesca Thanks, your suggestion resolved the build problem in the random tests. Unfortunately? now `all.sh` passes for me so I still don't know why the Jenkins builds are failing.

yanesca:
I copied the CI output here:
https://gist.github.com/yanesca/f250b5faa8e390b76eede7520e4ec84c

In the first couple of lines it prints out the OS and tools versions. Also here is my local environment (slightly more recent than the CI), I could reproduce with these too (I encountered the same compilation problem though):
```
* uname: /bin/uname: Linux mbedTLS-devPC1 4.15.0-74-generic #84-Ubuntu SMP Thu Dec 19 08:06:28 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
                                                                      
* gcc: /usr/bin/gcc: gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0

* ldd: /usr/bin/ldd: ldd (Ubuntu GLIBC 2.27-3ubuntu1) 2.27

* asan: Version: 5.5.0-12ubuntu1
```

The PR should be ready after fixing this test. This won't make the CI green though, because we have an unrelated failure in the CI.

jack-fortanix:
I was not able to reproduce until I edited `all.sh` to add `-fsanitize=leak` - for whatever reason it doesn't show up with just ASan. Anyway should be resolved now.

library/pkparse.c
```diff
@@ -769,14 +769,48 @@ static int pk_parse_key_pkcs1_der( mbedtls_rsa_context *rsa,
...
+    * are secret. See https://eprint.iacr.org/2020/055 for a
+    * description of one such attack.
+    */
+
+    /* Import DP */
+    if( ( ret = mbedtls_asn1_get_tag( &p, end, &len,
+                                      MBEDTLS_ASN1_INTEGER ) ) != 0 ||
+        ( ret = mbedtls_mpi_read_binary( &rsa->DP, p, len ) ) != 0 )
+        goto cleanup;
+    p += len;
```
yanesca: These line seem to be equivalent with
```C
if( ( ret = mbedtls_asn1_get_mpi( &p, end, &rsa->DP ) ) != 0 )
       goto cleanup;
```
Is there any reason why we shouldn't use `mbedtls_asn1_get_mpi()` here?

yanesca:
Yes, it is resolved, the CI is only failing on known issues that are unrelated to this PR.

Aren't you on OS X by any chance? Leak analyser is turned off by default in OS X ASan:
https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer

library/pkparse.c
jack-fortanix: I was just following the pattern in the rest of the function which uses asn1_get_tag +mpi_read_binary for this, but agree asn1_get_mpi is cleaner, will fix thanks

jack-fortanix:
Updated

yanesca:
The only thing left on this PR is to do the backports. @jack-fortanix Could you please raise PRs with the same content in the Mbed TLS repository targeting the `mbedtls-2.7` and `mbedtls-2.16` branches?

yanesca:
Thank you for submitting the backports!

PR #354, opened 2020-01-24
Fix incrementing pointer instead of value

mpg:
This was introduced by a hasty search-and-replace that embarrassingly didn't account for C's
operator precedence when changing those variables to pointer types.

## Backports

- [x] Mbed TLS 2.16: https://github.com/ARMmbed/mbedtls/pull/3007
- [x] Mbed TLS 2.7 is not affected (restartable ECC not introduced yet, so these variables are not pointers).

## Impact assessment

See the ChangeLog entry in the backport to 2.16.

### DIFF
--- a/library/ecdsa.c
+++ b/library/ecdsa.c
@@ -298,7 +298,7 @@ static int ecdsa_sign_restartable( mbedtls_ecp_group *grp,
     *p_sign_tries = 0;
     do
     {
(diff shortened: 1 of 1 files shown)


mpg:
The CI passes all tests except the ones involving Mbed OS, which is a known issue independent from this PR, so it's as good as a pass as far as merging this PR is concerned.

PR #357, opened 2020-02-03
Update Mbed Crypto with latest Mbed TLS changes as of 2020-02-03

gilles-peskine-arm:
The previous update was #322. Not much going on this time: a few bug fixes; `ASSERT_ALLOC` in test suites.

This is just one merge commit.


mpg:
I checked that the CI only fails the Mbed OS tests, which are a known issue independent from this PR.

yanesca:
Double checked that the CI only fails on Mbed OS tests.

PR #358, opened 2020-02-05
Define some structure for lifetime values

gilles-peskine-arm:
After some internal discussion ([private link](https://github.com/ARMmbed/psa-crypto-api/issues/207#issuecomment-582474856)), here's a proposal for how to assign lifetimes.

This proposal is backward-compatible with the current values for volatile keys (0) and persistent keys (1). Platforms with a secure element (i.e. a cryptoprocessor that can process keys which are not accessible to the PSA crypto core) should use the lifetime 0x1ff (whether the key material is stored inside the secure element, or in generic PSA storage in a form wrapped by the secure element).

This pull request defines the lifetime structure, but does not yet implement any new feature. A future pull request will adapt the (currently experimental) secure element support code accordingly.

We intend to include the types and values defined in this pull request with the same semantics in an upcoming release of the PSA Cryptography API specification.

@athoelke @MarcusJGStreets

### DIFF
--- a/include/psa/crypto_types.h
+++ b/include/psa/crypto_types.h
@@ -108,17 +108,114 @@ typedef uint32_t psa_algorithm_t;
  * The lifetime of a key indicates where it is stored and what system actions
  * may create and destroy it.
  *
- * Keys with the lifetime #PSA_KEY_LIFETIME_VOLATILE are automatically
- * destroyed when the application terminates or on a power reset.
+ * Lifetime values have the following structure:
+ * - Bits 0-7 (#PSA_KEY_LIFETIME_GET_PERSISTENCE(\c lifetime)):
+ *   persistence level. This value indicates what device management
+ *   actions can cause it to be destroyed. In particular, it indicates
+ *   whether the key is _volatile_ or _persistent_.
+ *   See ::psa_key_lifetime_persistence_t for more information.
+ * - Bits 8-31 (#PSA_KEY_LIFETIME_GET_LOCATION(\c lifetime)):
+ *   location indicator. This value indicates where the key is stored
+ *   and where operations on the key are performed.
+ *   See ::psa_key_lifetime_location_t for more information.
  *
- * Keys with a lifetime other than #PSA_KEY_LIFETIME_VOLATILE are said
- * to be _persistent_.
- * Persistent keys are preserved if the application or the system restarts.
+ * Volatile keys are automatically destroyed when the application instance
+ * terminates or on a power reset of the device. Persistent keys are
+ * preserved until the application explicitly destroys them or until an
+ * implementation-specific device management event occurs (for example,
+ * a factory reset).
+ *
  * Persistent keys have a key identifier of type #psa_key_id_t.
+ * This identifier remains valid throughout the lifetime of the key,
+ * even if the application instance that created the key terminates.
  * The application can call psa_open_key() to open a persistent key that
  * it created previously.
+ *
+ * This specification defines two basic lifetime values:
+ * - Keys with the lifetime #PSA_KEY_LIFETIME_VOLATILE are volatile.
+ *   All implementations should support this lifetime.
+ * - Keys with the lifetime #PSA_KEY_LIFETIME_PERSISTENT are persistent.
+ *   All implementations that have access to persistent storage with
+ *   appropriate security guarantees should support this lifetime.
  */
 typedef uint32_t psa_key_lifetime_t;
+
+/** Encoding of key persistence levels.
+ *
+ * What distinguishes different persistence levels is what device management
+ * events may cause keys to be destroyed. _Volatile_ keys are destroyed
+ * by a power reset. Persistent keys may be destroyed by events such as
+ * a transfer of ownership or a factory reset. What management events
+ * actually affect persistent keys at different levels is outside the
+ * scope of the PSA Cryptography specification.
+ *
+ * This specification defines the following values of persistence levels:
+ * - #PSA_KEY_LIFETIME_PERSISTENCE_VOLATILE (\c 0): volatile key.
+ *   A volatile key is automatically destroyed by the implementation when
+ *   the application instance terminates. In particular, a volatile key
+ *   is automatically on a power reset of the device.
+ * - \c 1: persistent key with a default lifetime.
+ *   Implementations should support this value if they support persistent
+ *   keys at all.
+ *   Applications should use this value if they have no specific needs that
+ *   are only met by implementation-specific features.
+ * - \c 2-127: persistent key with a PSA-specified lifetime.
+ *   The PSA Cryptography specification does not define the meaning of these
+ *   values, but other PSA specifications may do so.
+ * - \c 128-254: persistent key with a vendor-specified lifetime.
+ *   No PSA specification will define the meaning of these values, so
+ *   implementations may choose the meaning freely.
+ *   As a guideline, higher persistence levels should cause a key to survive
+ *   more management events than lower levels.
+ * - \c 255: read-only or write-once key.
+ *   A key with this persistence level cannot be destroyed.
+ *   Implementations that support such keys may either allow their creation
+ *   through the PSA Cryptography API, preferably only to applications with
+ *   the appropriate privilege, or only expose keys created through
+ *   implementation-specific means such as a factory ROM engraving process.
+ *   Note that keys that are read-only due to policy restrictions
+ *   rather than due to physical limitations should not have this
+ *   persistence levels.
+ *
+ * \note Key persistence levels are 8-bit values. Key management
+ *       interfaces operate on lifetimes (type ::psa_key_lifetime_t) which
+ *       encode the persistence as the lower 8 bits of a 32-bit value.
+ */
+typedef uint8_t psa_key_lifetime_persistence_t;
+
+/** Encoding of key location indicators.
+ *
+ * If an implementation of this API can make calls to external
+ * cryptoprocessors such as secure elements, the location of a key
+ * indicates which secure element performs the operations on the key.
+ * If an implementation offers multiple physical locations for persistent
+ * storage, the location indicator reflects at which physical location
+ * the key is stored.
+ *
+ * This specification defines the following values of location indicators:
+ * - \c 0: default location.
+ *   All implementations should support this value.
+ *   The default location typically indicates that the key material is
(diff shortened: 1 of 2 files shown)


include/psa/crypto_values.h
```diff
@@ -1555,9 +1563,43 @@
  * This lifetime value is the default storage area for the calling
  * application. Implementations may offer other storage areas designated
  * by other lifetime values as implementation-specific extensions.
+ * See ::psa_key_lifetime_location_t for more information.
  */
 #define PSA_KEY_LIFETIME_PERSISTENT             ((psa_key_lifetime_t)0x00000001)
 
+#define PSA_KEY_LIFETIME_PERSISTENCE_VOLATILE ((psa_key_lifetime_persistence_t)0x00)
+#define PSA_KEY_LIFETIME_PERSISTENCE_PRIMARY ((psa_key_lifetime_persistence_t)0x01)
```
yanesca: We have a location with the same value that we describe as the "primary secure element". This made me confused at first about the purpose of this constant.

```suggestion
#define PSA_KEY_LIFETIME_PERSISTENCE_PERSISTENT ((psa_key_lifetime_persistence_t)0x01)
```

include/psa/crypto_types.h
```diff
@@ -108,18 +108,115 @@ typedef uint32_t psa_algorithm_t;
...
+ *   are only met by implementation-specific features.
+ * - \c 2-127: persistent key with a PSA-specified lifetime.
+ *   The PSA Cryptography specification does not define the meaning of these
+ *   values, but other PSA specifications may do so.
+ * - \c 128-254: persistent key with a vendor-specified lifetime.
+ *   No PSA specification will define the meaning of these values, so
+ *   implementations may choose the meaning freely.
+ *   As a guideline, higher persistence levels should cause a key to survive
+ *   more management events than lower levels.
+ * - \c 255: read-only or write-once key.
```
yanesca:
```suggestion
 * - #PSA_KEY_LIFETIME_PERSISTENCE_READ_ONLY (\c 255): read-only or write-once key.
```

include/psa/crypto_values.h
```diff
@@ -1555,9 +1563,43 @@
...
+    ((psa_key_lifetime_persistence_t)((lifetime) & 0x000000ff)
+
+#define PSA_KEY_LIFETIME_GET_LOCATION(lifetime)      \
+    ((psa_key_lifetime_location_t)((lifetime) >> 8)
+
+/** Whether a key lifetime indicates that the key is volatile.
+ *
+ * A volatile key is automatically destroyed by the implementation when
+ * the application instance terminates. In particular, a volatile key
+ * is automatically on a power reset of the device.
```
yanesca:
```suggestion
 * is automatically destroyed on a power reset of the device.
```

include/psa/crypto_values.h
```diff
@@ -1555,9 +1563,43 @@
...
+ *
+ * A key that is not volatile is persistent. Persistent keys are
```
yanesca: This means that there can't be a lifetime for keys that survives the termination of the application but not a power reset. I can hardly imagine a use case where that would be needed, I just would like to confirm that this is what we want?

MarcusJGStreets: Seem sensible changes.
I am still thinking about "volatile" keys that last beyond the life of the calling partition.

include/psa/crypto_types.h
```diff
@@ -108,18 +108,115 @@ typedef uint32_t psa_algorithm_t;
...
+ *   A key with this persistence level cannot be destroyed.
+ *   Implementations that support such keys may either allow their creation
+ *   through the PSA Cryptography API, preferably only to applications with
+ *   the appropriate privilege, or only expose keys created through
+ *   implementation-specific means such as a factory ROM engraving process.
+ *   Note that keys that are read-only due to policy restrictions
+ *   rather than due to physical limitations should not have this
+ *   persistence levels.
```
athoelke: Perhap just for internal clarity, if not for the public documentation: is this lifetime appropriate for keys that have the same lifetime as the device identifier? - Some devices might be designed to support a factory process that erases the device reprograms it as a new device identity with new "read-only" keys.

include/psa/crypto_values.h
```diff
@@ -1555,9 +1563,43 @@
...
+ * \param lifetime      The lifetime value to query (value of type
+ *                      ::psa_key_lifetime_t).
+ *
+ * \return \c 1 if the key is volatile, otherwise \c 0.
+ */
+#define PSA_KEY_LIFETIME_IS_VOLATILE(lifetime)  \
+    (PSA_KEY_LIFETIME_GET_PERSISTENCE(lifetime) == \
+     PSA_KEY_LIFETIME_PERSISTENCE_VOLATILE)
+
+#define PSA_KEY_LOCATION_BUILT_IN    ((psa_key_lifetime_location_t)0x000000)
```
athoelke: Do we also want a PSA defined symbol for the specification allocated location 0x000001? Maybe `PSA_KEY_LOCATION_SECURE_ELEMENT`?

athoelke: These definitions do not have `LIFETIME` scoping term in the identifier. This should be added for consistency, or the lifetime/LIFETIME scope term should be removed from all of them (as described above).

include/psa/crypto_values.h
```diff
@@ -1555,9 +1563,43 @@
...
+#define PSA_KEY_LOCATION_VENDOR_FLAG ((psa_key_lifetime_location_t)0x800000)
```
athoelke: The other ranged values no longer have `VENDOR_FLAG` defintions - this will not appear in the PSA specification.

include/psa/crypto_values.h
gilles-peskine-arm: Yeah, I don't like `PERSISTENCE_PRIMARY` either. But `PERSISTENCE_PERSISTENT` is weird and misleading: other `PERSISTENT_xxx` values are also persistent. And I don't like `PERSISTENCE_DEFAULT` either (for example, when you create a key, the persistence defaults to `VOLATILE`, not to `DEFAULT`). Does anyone have a better suggestion?

include/psa/crypto_values.h
gilles-peskine-arm: A lifetime that survives the instance of the calling application but not a power reset could exist, but:

* This is an imprecise definition. If the application is running in a virtual machine, is the scope of the power reset the VM or its physical host?
* How does another instance of the application refer to the same key? It would need to use a key identifier, so this would be more like a persistent key from the application's perspective.
* Yes, it's possible. I've worked with a system that basically had this third intermediate scope (volatile-to-application-instance, volatile-to-device, persistent). But AFAIK this intermediate scope was never used outside of the test suite. So I'm with “can hardly imagine a use case”.
* If an implementation really wants this feature, it can provide it through the vendor range for persistence levels.

include/psa/crypto_types.h
```diff
@@ -108,18 +108,115 @@ typedef uint32_t psa_algorithm_t;
...
+ *
+ * \note Key persistence levels are 8-bit values. Key management
+ *       interfaces operate on lifetimes (type ::psa_key_lifetime_t) which
+ *       encode the persistence as the lower 8 bits of a 32-bit value.
+ */
+typedef uint8_t psa_key_lifetime_persistence_t;
```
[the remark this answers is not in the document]
gilles-peskine-arm: I prefer to make the connection with lifetimes explicit.

include/psa/crypto_types.h
gilles-peskine-arm: If the device identifier is set in stone, that's the corresponding lifetime. If the device identifier is rewritten on a factory reset, this corresponds to a different lifetime (which may be 1 or something else depending on what levels of “factory reset” exist).

include/psa/crypto_values.h
gilles-peskine-arm: Right. Like the other helper macros, these macros have no Doxygen, so they aren't intended to be included in the PSA specification.

include/psa/crypto_values.h
gilles-peskine-arm: So far I consider this an implementation-specific definition, but I don't have a strong opinion about it.

include/psa/crypto_values.h
MarcusJGStreets: Yes that covers what I meant by "thinking".
In general PSA  has been writtien for M-class with a fixed set of secure partitions all of which active - so there really is no concept of application life time except reboot. 

However, when we extend to Cortex A we will need to deal with other life times - but I believe that can be left to the next release.

include/psa/crypto_values.h
athoelke: So, unlike other imp-specific parts of the API, this one has a specified value, but not a specified symbol? A bit unusual in the documentation, but we don't provide handy spec-defined symbols for all of the [vendor] ranges in other value types either.

It's a bit like `persistence == 1`, which we can't find an ideal name for either, but really have to name because it is effectively used by `PSA_KEY_LIFETIME_PERSISTENT`.

include/psa/crypto_values.h
yanesca: Well it might be a little bit overkill, but how about `PSA_KEY_LIFETIME_PERSISTENCE_PERSISTENT_DEFAULT`?

athoelke: Another candidate perhaps: `PSA_KEY_LIFETIME_PERSISTENCE_STANDARD`?

MarcusJGStreets: Standard works for me

gilles-peskine-arm:
Since Mbed Crypto is no longer updated, I've rebased this pull request on top of Mbed TLS. To be continued at https://github.com/ARMmbed/mbedtls/pull/3302

include/psa/crypto_values.h
gilles-peskine-arm: I don't like “standard”: which standard? I [went with “default”](https://github.com/ARMmbed/mbedtls/pull/3302/commits/87977450e154398e5f0b0422d2d93515dc70973e).

include/psa/crypto_types.h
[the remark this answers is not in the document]
gilles-peskine-arm: On second thoughts [I shortened the names](https://github.com/ARMmbed/mbedtls/pull/3302/commits/27bf930f4dcf37133a39e5284d0c44ac45945e1d).

PR #364, opened 2020-02-10
Key wrapping API proposal

gilles-peskine-arm:
This is a first draft of the key wrapping API for PSA Cryptography 1.x.

This pull request only provides function prototypes and associated macro definitions. Tests and implementations will come later.

I implemented the proposal discussed [in private](https://github.com/ARMmbed/psa-crypto-api/issues/77#issuecomment-582506652). I'm not attached to function/macro names if you have better suggestions. The API comprises:

* Functions to back up and restore a key, wrapped with another key, using a vendor-chosen algorithm.
* A reserved identifier value for a wrapping key. There is currently no way to create a wrapping key that can wrap keys with metadata.
* Functions to wrap key material with another key, using standard algorithms.

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -1029,6 +1029,117 @@ psa_status_t psa_get_wrapped_key_attributes(const uint8_t *data,
                                             size_t data_length,
                                             psa_key_attributes_t *attributes);
 
+/**
+ * \brief Export key material in wrapped form.
+ *
+ * A wrapped form of the key material preserves the confidentiality
+ * and authenticity of the key material. In practical terms, the key
+ * material is encrypted and authenticated.
+ *
+ * The policy on the key must have the usage flag
+ * #PSA_KEY_USAGE_EXPORT_WRAPPED set.
+ *
+ * \param wrapping_key      Handle to the key to wrap with.
+ * \param alg               The key wrapping algorithm to compute
+ *                          (\c PSA_ALG_XXX value such that
+ *                          #PSA_ALG_IS_KEY_WRAP(\p alg) is true).
+ * \param handle            Handle to the key whose material is to be
+ *                          exported in wrapped form.
+ * \param[out] data         Buffer where the wrapped key data is to be written.
+ * \param data_size         Size of the \p data buffer in bytes.
+ * \param[out] data_length  On success, the number of bytes
+ *                          that make up the wrapped key data.
+ *
+ * \retval #PSA_SUCCESS
+ * \retval #PSA_ERROR_INVALID_HANDLE
+ *         One or both of \p handle and \p wrapping_key is not a valid
+ *         handle to a key.
+ * \retval #PSA_ERROR_NOT_PERMITTED
+ *         The key \p handle does not have the #PSA_KEY_USAGE_BACKUP flag.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         \p wrapping_key does not support wrapping key material.
+ * \retval #PSA_ERROR_NOT_SUPPORTED
+ *         \p wrapping_key does not support wrapping the key designated
+ *         by \p handle.
+ * \retval #PSA_ERROR_BUFFER_TOO_SMALL
+ *         The size of the \p data buffer is too small. You can determine a
+ *         sufficient buffer size by calling
+ *         #PSA_WRAP_KEY_MATERIAL_OUTPUT_SIZE(\p alg, \c type, \c bits)
+ *         where \c type is the key type of \p handle
+ *         and \c bits is the key size of \p handle in bits.
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_HARDWARE_FAILURE
+ * \retval #PSA_ERROR_CORRUPTION_DETECTED
+ * \retval #PSA_ERROR_STORAGE_FAILURE
+ * \retval #PSA_ERROR_INSUFFICIENT_MEMORY
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+psa_status_t psa_wrap_key_material(psa_key_handle_t wrapping_key,
+                                   psa_algorithm_t alg,
+                                   psa_key_handle_t handle,
+                                   uint8_t *data,
+                                   size_t data_size,
+                                   size_t *data_length);
+
+/**
+ * \brief Import wrapped key material.
+ *
+ * \param wrapping_key      Handle to the key to unwrap with.
+ * \param alg               The key unwrapping algorithm to compute
+ *                          (\c PSA_ALG_XXX value such that
+ *                          #PSA_ALG_IS_KEY_WRAP(\p alg) is true).
+ * \param[in] attributes    The attributes for the new key.
+ *                          They are used in the same way as by
+ *                          psa_import_key().
+ * \param[in] data          Buffer containing the key data. The expected
+ *                          format of this buffer depends on the wrapping
+ *                          key and on the type declared in \p attributes.
+ * \param data_length       Size of the \p data buffer in bytes.
+ * \param[out] handle       On success, a handle to the newly created key.
+ *                          \c 0 on failure.
+ *
+ * \retval #PSA_SUCCESS
+ *         Success.
+ *         If the unwrapped key is persistent, the key material and the
+ *         key's metadata have been saved to persistent storage.
+ * \retval #PSA_ERROR_ALREADY_EXISTS
+ *         This is an attempt to create a persistent key, and there is
+ *         already a persistent key with the given identifier.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         The key attributes, as a whole, are invalid.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         The key data is not correctly formatted.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         The size in \p attributes is nonzero and does not match the size
+ *         of the key data.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         \p wrapping_key does not support unwrapping keys with metadata.
+ * \retval #PSA_ERROR_INVALID_SIGNATURE
+ *         \p data is not a valid wrapped key for \p wrapping_key.
+ * \retval #PSA_ERROR_NOT_SUPPORTED
+ *         Some of the metadata in either \p attributes or \p data is
+ *         not supported.
+ * \retval #PSA_ERROR_INSUFFICIENT_MEMORY
+ * \retval #PSA_ERROR_INSUFFICIENT_STORAGE
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_STORAGE_FAILURE
+ * \retval #PSA_ERROR_HARDWARE_FAILURE
+ * \retval #PSA_ERROR_CORRUPTION_DETECTED
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+psa_status_t psa_unwrap_key_material(psa_key_handle_t wrapping_key,
+                                     psa_algorithm_t alg,
+                                     const psa_key_attributes_t *attributes,
+                                     const uint8_t *data,
+                                     size_t data_length,
+                                     psa_key_handle_t *handle);
+
 
 
 /**@}*/
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -854,6 +854,134 @@ psa_status_t psa_export_public_key(psa_key_handle_t handle,
                                    size_t data_size,
                                    size_t *data_length);
 
+/**
+ * \brief Export a key and its metadata in wrapped form.
+ *
+ * A wrapped form of the key object preserves the confidentiality and
+ * authenticity of the key material and the authenticity of the key
+ * policy. In practical terms, the key material is encrypted, and
+ * the key data and metadata are authenticated together.
+ *
+ * The format of the wrapped data is implementation-dependent. It may depend
+ * both on the choice of wrapping key and on the type of key to wrap.
+ *
+ * The policy on the key must have the usage flag #PSA_KEY_USAGE_BACKUP set.
+ *
+ * \param wrapping_key      Handle to the key to wrap with.
+ * \param handle            Handle to the key to export in wrapped form.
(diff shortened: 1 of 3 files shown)


include/psa/crypto.h
```diff
@@ -852,6 +854,292 @@ psa_status_t psa_export_public_key(psa_key_handle_t handle,
...
+ * used to wrap. For asymmetric wrapping where the wrapping operation uses
+ * a public key and the unwrapping operation uses the corresponding private
+ * key, you must unwrap with the corresponding unwrapping key.
+ *
+ * \param wrapping_key      Handle to the key to unwrap with.
+ * \param[in] attributes    The attributes for the new key.
+ *                          They are used as follows:
+ *                          - The key type and size may be 0. If either is
+ *                            nonzero, it must match the corresponding
+ *                            attribute of the source key.
```
MarcusJGStreets: You have used both source key and \p data which I believe both refer to the key to unwrap. 
I would prefer consistency as it is quite hard to understand.

include/psa/crypto.h
```diff
@@ -852,6 +854,292 @@ psa_status_t psa_export_public_key(psa_key_handle_t handle,
...
+ *                          - The key location (the lifetime and, for
+ *                            persistent keys, the key identifier) is
+ *                            used directly.
+ *                            If the wrapped key does not have the usage
+ *                            flag #PSA_KEY_USAGE_COPY, then the location
+ *                            must match the location embedded in \p data.
+ *                            If the wrapped key has the usage
+ *                            flag #PSA_KEY_USAGE_COPY, then the location
+ *                            embedded in \p data is ignored.
```
MarcusJGStreets: I believe that from this description the caller must set attributes -> location.
So the caller needs to call psa_get_wrapped_key_attributes

Is there a reason why they should not be able to leave it unset and use the one in the wrapped key?
And potentiall not offer psa_get_wrapped_key_attributes?

include/psa/crypto.h
```diff
@@ -852,6 +854,292 @@ psa_status_t psa_export_public_key(psa_key_handle_t handle,
...
+ * // Here attributes contains the policy, location and other attributes
+ * // of the key.
+ * // To unwrap the key to a different location,
+ * // call psa_set_key_lifetime() and psa_set_key_id() here.
+ * psa_unwrap_key_with_policy(wrapping_key, &attributes,
+ *                            wrapped_data, wrapped_data_length,
+ *                            &handle)
+ * psa_reset_key_attributes(&attributes);
+ * \endcode
+ *
```
MarcusJGStreets: Is it worth adding a line checking if the copy permission is set.

include/psa/crypto.h
gilles-peskine-arm: fixed

include/psa/crypto.h
gilles-peskine-arm: I'd prefer not to have `psa_get_wrapped_key_attributes`, but I added it because I couldn't find a better way.

“Leave unset” works for the key type and the key size because for those 0 is not a valid value, so it's interpreted as unset. But for lifetimes, 0 is a valid value, indicating a volatile key: keys are volatile by default, you only need to specify a lifetime explicitly if you want the key to live longer. If a wrapped key is non-copyable and persistent, attempting to unwrap it as volatile should fail.

Likewise, for the policy, 0 means “you can't do anything”. That's not the same thing as “use whatever is in the wrapped data”.

include/psa/crypto.h
gilles-peskine-arm: Why would the calling code check the copy permission? I'd expect typical application code to either expect a copyable key and specify a location here, or not expect a copyable key and not specify the location. Neither case calls for checking the copy permission in the application code.

include/psa/crypto.h
MarcusJGStreets: Ok - good answer.

include/psa/crypto.h
MarcusJGStreets: Thanks

include/psa/crypto.h
MarcusJGStreets: Yes you are probably right for IOT.

include/psa/crypto.h
athoelke: Good use cases. How about one or moe of these options as alternatives:

* Permit `attributes == NULL` - this indicates that the key should be unwrapped exactly as it was wrapped. No override attributes.
* Define a special lifetime value to mean `ANY` - i.e. "as wrapped"
* Permit usage flags of MAX_UINT32 - and rely on the 'constrain to wrapped key policy' behaviour to replicate the wrapped key usage flags in the unwrapped key.

include/psa/crypto_values.h
```diff
@@ -1571,6 +1601,22 @@
...
+ * #PSA_KEY_USAGE_BACKUP usage flag and later restore it on the same device,
+ * if the implementation supports it.
+ *
+ * The existence of this key is optional. Implementations that support
+ * key wrapping should offer a key with this identifier if having a
+ * single wrapping key makes sense in the scenarios where this implementation
+ * is used. If this key exists, it must be private to the device, i.e. it
+ * must not be shared with any other device or entity.
+ */
+#define PSA_KEY_ID_WRAP_BOUND                   ((psa_app_key_id_t)0x80000062)
```
athoelke: Is there any particular significance to the value chosen? - why MIN_UINT32 + 0x62?

athoelke:
Do we also need a usage flag for WRAP and UNWRAP, as well as the defined flags for a key being exporting in wrapped form? Keys used as wrapping keys in `psa_wrap_key_*()` or `psa_unwrap_key_*()` would then require the applicable usage flag.

include/psa/crypto.h
```diff
@@ -852,6 +854,292 @@ psa_status_t psa_export_public_key(psa_key_handle_t handle,
...
+ * \param data_size         Size of the \p data buffer in bytes.
+ * \param[out] data_length  On success, the number of bytes
+ *                          that make up the wrapped key data.
+ *
+ * \retval #PSA_SUCCESS
+ * \retval #PSA_ERROR_INVALID_HANDLE
+ *         One or both of \p handle and \p wrapping_key is not a valid
+ *         handle to a key.
+ * \retval #PSA_ERROR_NOT_PERMITTED
+ *         The key \p handle does not have the #PSA_KEY_USAGE_BACKUP flag.
```
athoelke: copy-pasta: presume `PSA_KEY_USAGE_EXPORT_WRAPPED`

include/psa/crypto.h
```diff
@@ -852,6 +854,292 @@ psa_status_t psa_export_public_key(psa_key_handle_t handle,
...
+ *         The key data is not correctly formatted.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         The size in \p attributes is nonzero and does not match the size
+ *         of the key data.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         \p wrapping_key does not support unwrapping keys with metadata.
+ * \retval #PSA_ERROR_INVALID_SIGNATURE
+ *         \p data is not a valid wrapped key for \p wrapping_key.
+ * \retval #PSA_ERROR_NOT_SUPPORTED
+ *         Some of the metadata in either \p attributes or \p data is
```
athoelke: There is no metadata in `data` for this API.

include/psa/crypto_values.h
```diff
@@ -1082,6 +1094,24 @@
...
     ref :
 
+/** The KW algorithm from NIST SP 800-38F, also described in RFC 3394.
+ *
+ * This algorithm may be used with any key type for a block cipher with
+ * a 128-bit block size, not just AES. The calculations are the same as
+ * described in SP 800-38F, but replacing uses of AES with the 128-bit
+ * block cipher that the key is for.
+ */
+#define PSA_ALG_NIST_KW                         ((psa_algorithm_t)0x0a401003)
```
athoelke: This does not satisfy `PSA_ALG_IS_KEY_WRAP()` - the category should be `0x0e000000`?

include/psa/crypto_values.h
```diff
@@ -1082,6 +1094,24 @@
...
+
+/** The KWP algorithm from NIST SP 800-38F.
+ *
+ * This algorithm may be used with any key type for a block cipher with
+ * a 128-bit block size, not just AES. The calculations are the same as
+ * described in SP 800-38F, but replacing uses of AES with the 128-bit
+ * block cipher that the key is for.
+ */
+#define PSA_ALG_NIST_KWP                        ((psa_algorithm_t)0x0a401005)
```
athoelke: This does not satisfy `PSA_ALG_IS_KEY_WRAP()` - the category should be `0x0e000000`?

include/psa/crypto_values.h
athoelke: Not enamoured with the name, as we haven't used "bound" anywhere else in the API or documentation. Is `PSA_KEY_ID_WRAP_BUILT_IN` any better, or worse?

include/psa/crypto_values.h
```diff
@@ -1606,6 +1652,42 @@
...
+ *
+ * This flag allows the use of psa_wrap_key_with_policy() on the given key,
+ * with any suitable wrapping key.
+ *
+ * A wrapped form of the key object preserves the confidentiality and
+ * authenticity of the key material and the authenticity of the key
+ * policy. In practical terms, the key material is encrypted, and
+ * the key data and metadata are authenticated together.
+ */
+#define PSA_KEY_USAGE_BACKUP                    ((psa_key_usage_t)0x00000020)
```
athoelke: If we used `PSA_KEY_ID_WRAP_BOUND` then I would be inclined to make this `PSA_KEY_USAGE_EXPORT_BOUND` or perhaps `PSA_KEY_USAGE_WRAP_BOUND`.

`BACKUP` is quite cute, but then we should use 'backup' in the wrap-with-policy API name and the built-in wrap-with-policy binding key to strengthen the association of these API elements.

include/psa/crypto.h
gilles-peskine-arm: > Permit attributes == NULL

Doesn't help if you want to use the wrapped policy but specify a different lifetime or id, which I think is a very common use case.

> Permit usage flags of MAX_UINT32

That does in fact work for the usage flags, but not for the algorithm. It could be added as a feature for the algorithm.

include/psa/crypto_values.h
gilles-peskine-arm: 0x80000000 is the start of the range reserved for ids defined by PSA.

`0x80000000 + 'b'` (for “bound”) because why not.

I'm very open to naming suggestions. Once a few people have made theirs I'll do a pass of renaming stuff with an eye to consistency and the other eye to comprehensibility.

include/psa/crypto.h
athoelke: The second bullet implies that we define (within the new structured framework) a special lifetime value, e.g.

    #define PSA_KEY_LIFETIME_UNSPECIFIED ((psa_key_lifetime_t) 0x0000007f)

This value is only valid in attributes passed to `psa_unwrap_key_with_policy()` to indicate that the key should be unwrapped into the same lifetime that it was wrapped from.

You could also support all of the above mechanisms:

1. If you want to restore a backed up key, just pass `attributes = NULL`. (Volatile keys might have a different key handle/id after unwrapping)
2. If you want to move the key then initialise the attributes, set the wildcard usage flags (`PSA_KEY_USAGE_UNSPECIFIED` - all bits 1, not all known bits one: only valid for unwrapping/copying), and set the lifetime/id for the desired location
3. If you you want to limit the policy, but maintain the lifetime, then initialise the attributes, set the usage flags, and set the lifetime to `PSA_KEY_LIFETIME_UNSPECIFIED` (only valid for unwrap).

For (1), it would be possible to just use `UNSPECIFIED` for both the usage flags and the lifetime, but I suspect that the compact `attributes = NULL` form might be a common usage?

athoelke: Instead of `*_UNSPECIFIED`, other terms could be `*_AS_SOURCE` or `*_ORIGINAL`.

gilles-peskine-arm:
@MarcusJGStreets @athoelke I pushed an update with a rewritten history that should resolve all your comments and cover what we discussed orally today, other than the naming. The original version is in https://github.com/gilles-peskine-arm/mbed-crypto/tree/psa-wrap-api-1

* I fixed copypasta by amending the corresponding commits.
* I removed the commit that introduced `psa_get_wrapped_key_attributes`, and instead made a new commit introducing `psa_unwrap_key_to_alternate_lifetime` and removing the `attributes` parameter to `psa_unwrap_key_with_policy`.
* I added usage flags for wrapping keys.

I'll subsequently make another update to fix the terminology issues (which includes finding a decent name for `psa_unwrap_key_to_alternate_lifetime`; I couldn't resolve myself to `psa_unwrap_key_with_policy_to_alternate_lifetime`). I'm more and more warming up to backup/restore for wrap-with-policy. It isn't the only use case for this function (there's also sharing, when another entity has the unwrapping key), but it's a common one. I think anyone would understand backup/restore, even if it isn't what crypto architects would use, unlike “binding”, which even crypto architects don't fully agree on.

include/psa/crypto.h
```diff
@@ -854,6 +854,135 @@ psa_status_t psa_export_public_key(psa_key_handle_t handle,
...
+ *
+ * A wrapped form of the key object preserves the confidentiality and
+ * authenticity of the key material and the authenticity of the key
+ * policy. In practical terms, the key material is encrypted, and
+ * the key data and metadata are authenticated together.
+ *
+ * The format of the wrapped data is implementation-dependent. It may depend
+ * both on the choice of wrapping key and on the type of key to wrap.
+ *
+ * The policy on the key must have the usage flag
```
athoelke: Need to specify which key needs this policy?

include/psa/crypto.h
```diff
@@ -852,7 +854,320 @@ psa_status_t psa_export_public_key(psa_key_handle_t handle,
...
+ * #PSA_KEY_USAGE_EXPORT_WRAPPED set.
```
athoelke: #PSA_KEY_USAGE_BACKUP for this API

include/psa/crypto.h
athoelke: This issue is still present.

athoelke:
Hi @gilles-peskine-arm - thanks for the update.

* I'm quite happy with the idea of backup/restore terminology. Would you drop the `_material` from the other wrapping functions as a result?

* Some ideas for the tricky-to-name API:
   - `psa_restore_key_as_volatile()`
   - `psa_restore_key_elsewhere()` :-)
   - `psa_restore_key_ex()

The usage flag naming gets interesting now that we have "can be wrapped" and "can use to wrap" policy flags. It is clear there are two categories of flags:
1. Management policies - things the implementation/application can do **to** a key object: EXPORT, COPY, CACHE, WRAP, BACKUP
2. Usage policies - things the implementation/application can do **with** the key material: ENCRYPT, DECRYPT, SIGN, VERIFY, DERIVE, WRAP.

I wonder now if we should have named these categories of flag separately to make it easier to distinguish can-wrap-with from can-be-wrapped.

giacomo3000 (approved):

PR #365, opened 2020-02-10
Change the compatibility API to inline functions

soby-mathew:
This patch changes the compatibility API defined in crypto_compat.h
to static inline functions as the previous macro definitions were
causing issues for the C pre-processor when included in projects
which need to redefine the PSA function names. Making it static
inline function solves this problem neatly and also modern compilers
do a good job at inlining the function which makes the need for making
it a macro redundant.

Signed-off-by: Soby Mathew <soby.mathew@arm.com>

### DIFF
--- a/include/psa/crypto_compat.h
+++ b/include/psa/crypto_compat.h
@@ -89,18 +89,28 @@ typedef MBEDTLS_PSA_DEPRECATED psa_dh_group_t mbedtls_deprecated_psa_dh_group_t;
 /*
  * Deprecated PSA Crypto function names (PSA Crypto API  <= 1.0 beta3)
  */
-/* Make these macros and not wrappers so that there is no cost to
- * applications that don't use the deprecated names.
- *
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm:
Thanks Soby. Making the old function names be static inline functions rather than macros is ok. However, you'll need to change several things to pass the CI. I think the functions should return `psa_status_t`: `mbedtls_deprecated_psa_status_t` is a hack to get a deprecation warning via macro usage. Instead, use `MBEDTLS_PSA_DEPRECATED` at the beginning of the function declaration.

Please ensure that the CI passes. In particular, run
```
tests/scripts/check-files.py
tests/scripts/check-names.sh
tests/scripts/all.sh build_deprecated
```

soby-mathew:
Thanks Gilles. I have updated with new patch.

gilles-peskine-arm:
CI is failing only on a known unrelated issue on Mbed OS, so that's as good as a pass.

PR #368, opened 2020-02-15
Implemented inplace base64 encoding aligning it to decode function

mirkoflchtt:
I've noticed that base64 encoding routine was not working properly when src and dst buffers overlaps.
Safe inplace encoding could be a nice to have feature on some very limited memory device, allowing as a sideback src buffer to be overwritten (safely) during the encoding.
Thus I have modify the routine to always encode from back to front considering that output buffer is always larger than src buffer. I've also updated self_test to test for both encoding and decoding that
inplace operations are correctly implemented.


gilles-peskine-arm:
Thank you for your contribution! Before we review it:

* Please make sure that the CI passes. At the moment, `tests/scripts/check-files.py` fails.
* Please make separate commits for the self-test changes, the code changes, and the interface/documentation changes.

include/mbedtls/base64.h
```diff
@@ -42,7 +42,9 @@ extern "C" {
 /**
  * \brief          Encode a buffer into base64 format
  *
- * \param dst      destination buffer
+ * \param dst      destination buffer (can be NULL for checking size, also src
+ *                 and dst buffers may overlap: int this case inplace encoding
+ *                 takes place and src buffer is overwritten)
```
gilles-peskine-arm: Please use full sentences, correct punctuation, and applicable formatting. Older documentation sometimes takes liberties, but we fix those when we touch them.

> Destination buffer. This can be `\c NULL` to obtain the required buffer size.
> The \p src and \p dst buffers may overlap.

gilles-peskine-arm: Note that in case of partial overlap, this function is only allowed to write between `dst` and `dst + dlen`. It is not allowed to overwrite any part of the `src` buffer that does not overlap, since `src` is a `const *` parameter.

mirkoflchtt: Hopefully it should be ok now.

mirkoflchtt: Not sure to understand. The proposed patch is intended to support full buffer overlapping to potentially save memory on some scenario. Former implementation was not allowing this. I don't think partial overlapping is that relevant.
But maybe I missed something there..

mirkoflchtt:
Not sure what's going on there. I follow the instructions to pass dco checks rebasing the repositoru.
That added others 25 commits to pull request.

gilles-peskine-arm:
You appear to have made a mistake during a rebase. Please rebase the pull request to have just your own commits. It might be easier to start from the current state of `development` and `git cherry-pick` your commits one by one.

gilles-peskine-arm:
Having base64 work in place would be nice. However, this pull request appears to have been abandoned, and the Mbed TLS maintenance team does not intend to work on it.

If you're still interested in this feature, please raise a new pull request against Mbed TLS which has reabsorbed Mbed Crypto.

PR #370, opened 2020-02-19
Bump version to Mbed TLS 2.21.0

yanesca:
Not bumping the SO version:

The API/ABI breaking changes only affect internal APIs (changes related to `mbedtls_asn1_traverse_sequence_of()`) and to PSA, which is from the point of view of Mbed TLS is an unstable API.

### DIFF
--- a/doxygen/mbedtls.doxyfile
+++ b/doxygen/mbedtls.doxyfile
@@ -28,7 +28,7 @@ DOXYFILE_ENCODING      = UTF-8
 # identify the project. Note that if you do not use Doxywizard you need
 # to put quotes around the project name if it contains spaces.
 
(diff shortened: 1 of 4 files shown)


gilles-peskine-arm:
Don't we need to update the crypto submodule and the changelog first?

yanesca:
This is mbed-crypto, it does not have it's own version reporting, it reports the Mbed TLS version it is included in. I wanted to update the version here first so that we can do a single submodule update on Mbed TLS instead of two.

yanesca:
The CI failures are mostly an issue in Mbed OS, with one exception, which is an ssl-opt.sh DTLS spurious resend. All of these are known issues unrelated to this PR.

PR #376, opened 2020-02-26
Modify path handling in some build scripts to be closer to mbedtls

gilles-peskine-arm:
Mostly this aligns the Visual Studio templates with mbedtls and makes the handling of directories in `generate_visual_studio.pl` more generic.

There is no completeness goal in this pull request, only a goal of non-harmfulness (no semantic change).


mpg:
I check the CI results and the only failures are in mbed-os tests, which are known and expected to fail at this point, so can be ignored for this PR. Therefore I'm removing "needs: ci" and labeling "ready for merge".

PR #378, opened 2020-02-28
Fix useless calls in PSA crypto metadata tests

ronald-cron-arm:
In ecc_key_family test, the parity of the elliptic curve identifier
was computed but not tested to be even.

In dh_key_family test, the parity of the PSA Diffie-Hellman group
identifier was computed but not tested to be even.

Signed-off-by: Ronald Cron <ronald.cron@arm.com>

### DIFF
--- a/tests/suites/test_suite_psa_crypto_metadata.function
+++ b/tests/suites/test_suite_psa_crypto_metadata.function
@@ -70,8 +70,6 @@ int test_parity( uint32_t value )
     value ^= value >> 4;
     return( 0x9669 & 1 << ( value & 0xf ) );
 }
-#define TEST_PARITY( value )                    \
-    TEST_ASSERT( test_parity( value ) )
 
 void algorithm_classification( psa_algorithm_t alg, unsigned flags )
 {
@@ -130,7 +128,7 @@ void key_type_classification( psa_key_type_t type, unsigned flags )
                 ( PSA_KEY_TYPE_IS_DH( type ) &&
                   PSA_KEY_TYPE_IS_PUBLIC_KEY( type ) ) );
 
-    TEST_PARITY( type );
+    TEST_ASSERT( test_parity( type ) );
 
 exit: ;
 }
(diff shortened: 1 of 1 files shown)


tests/suites/test_suite_psa_crypto_metadata.function
```diff
@@ -56,17 +56,15 @@
 #define TEST_CLASSIFICATION_MACRO( flag, alg, flags )           \
     TEST_ASSERT( PSA_##flag( alg ) == !! ( ( flags ) & flag ) )
 
-/* Check the parity of value.
- * Return 0 if value has even parity and a nonzero value otherwise. */
-int test_parity( uint32_t value )
+/* Check if the parity of value is even.
```
gilles-peskine-arm: Please comment the _why_ in preference to the _what_. The job of this function is to test whether the parity is whichever parity is expected for numerical values. The fact that it's even parity is a detail.

See the discussion in https://github.com/ARMmbed/mbed-crypto/pull/330#discussion_r360305364 . This discussion also highlights that “even parity” is not such good documentation: Jaeden calls it “odd parity”!

My bad for not documenting why we're checking the parity of some values. Here's some proposed documentation.

```
/* Check the parity of value.
 * Numerical encodings of key types and related values such as EC curves and
 * DH groups should all have the same parity. This guarantees that a
 * single-bit error will be detected.
 * The expected parity is even because this makes all-bits-zero valid.
 */
int test_parity( uint32_t value )
```

gilles-peskine-arm: Please don't combine unrelated things into the same commit. Here you have a commit that does four things: adding documentation, renaming a function, changing the return value, and adding the missing `TEST_ASSERT` calls. Adding documentation could go with changing the return value but the other three should be separate.

ronald-cron-arm: I've changed the function documentation taking into account your proposal. Please have a look to the new version.

ronald-cron-arm: I don't rename the function and don't change the return value anymore. Two commits now: the first one fix/improve the documentation of the test_parity() function. The second one fixes the useless calls.

mpg:
Everything passed in the CI except for mbed-os and iar tests which are known issues.

PR #386, opened 2020-03-06
Add support for ASN.1 tags on multiple bytes

frankmorgner:
Replaces https://github.com/ARMmbed/mbedtls/pull/848; I think all formal requirements and change requests posted in the original PR are already considered.


gilles-peskine-arm:
Since [Mbed Crypto has been merged back into Mbed TLS](https://github.com/ARMmbed/mbed-crypto/issues/374), we can continue the review at https://github.com/ARMmbed/mbedtls/pull/848 .

PR #390, opened 2020-03-19
The public exponent is not required for mbedtls_rsa_private()

mvardan:
Signed-off-by: Vardan Mikayelyan <mvardan.dev@gmail.com>

### DIFF
--- a/tests/suites/test_suite_rsa.function
+++ b/tests/suites/test_suite_rsa.function
@@ -870,15 +870,18 @@ void mbedtls_rsa_private( data_t * message_str, int mod, int radix_P,
                           int result )
 {
     unsigned char output[256];
-    mbedtls_rsa_context ctx, ctx2; /* Also test mbedtls_rsa_copy() while at it */
-    mbedtls_mpi N, P, Q, E;
+    mbedtls_rsa_context ctx, ctx2, ctx3; /* Also test mbedtls_rsa_copy() while at it */
+    mbedtls_mpi N, P, Q, D, E;
     rnd_pseudo_info rnd_info;
     int i;
 
     mbedtls_mpi_init( &N ); mbedtls_mpi_init( &P );
     mbedtls_mpi_init( &Q ); mbedtls_mpi_init( &E );
+    mbedtls_mpi_init( &D );
+
     mbedtls_rsa_init( &ctx, MBEDTLS_RSA_PKCS_V15, 0 );
     mbedtls_rsa_init( &ctx2, MBEDTLS_RSA_PKCS_V15, 0 );
+    mbedtls_rsa_init( &ctx3, MBEDTLS_RSA_PKCS_V15, 0 );
 
     memset( &rnd_info, 0, sizeof( rnd_pseudo_info ) );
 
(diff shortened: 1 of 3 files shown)


mvardan:
I have no access to the test results, so if fails are meaningful please provide results to me.

mpg:
Hi @mvardan, and thanks for your contribution!

Unfortunately, I don't think I agree with that change: some private key operations might need to use the public exponent for counter-measures against some attacks. For example, currently our code for PKCS#1 v1.5 signing uses the public exponent in order to verify the signature as a counter-measure against [Lenstra's attack](https://people.redhat.com/~fweimer/rsa-crt-leaks.pdf), via a call to `mbedtls_rsa_public()`: https://github.com/ARMmbed/mbed-crypto/blob/4fca6b1a15fb3575f7f3dfc284edd6ca39f79170/library/rsa.c#L2074

So I don't think it would be appropriate to remove this check for private keys. Is there a particular reason why you wanted to remove this check?

mvardan:
Hi @mpg, thank you for your detailed feedback,

I have source code that currently works with custom HW, I wanted to try mbed_tls as an alternative SW implementation. In my case, I have API which is wrapping mbedtls_rsa_private() and that API doesn't have public exponent.

As a user, in some cases, I would like to have the possibility to perform mbedtls_rsa_*** APIs with minimal input. Is it an option to make this configurable? If so I will be happy to refactor this and add a new configuration.

BTW: In the particular use case which was described by you (PKCS#1 v1.5 signing), mbedtls_rsa_public() will still fail, it will error out from [here](https://github.com/ARMmbed/mbed-crypto/blob/4fca6b1a15fb3575f7f3dfc284edd6ca39f79170/library/rsa.c#L2092).

mvardan:
Unfortunately, I can't provide more details about my environment.

I have added a test-case with a possible use-case. I am not pretending to have this merged.
Currently `mbedtls_rsa_import()` allows filling `mbedtls_rsa_context` having only N and D without E and with this trick, it will correctly perform `mbedtls_rsa_private()`.

I understand that it will be costly to provide a solid solution, but if you'll decide to add support for this use case, I will be happy to spend some time contributing to open-source.

tests/suites/test_suite_rsa.function
```diff
@@ -922,11 +925,24 @@ void mbedtls_rsa_private( data_t * message_str, int mod, int radix_P,
         TEST_ASSERT( hexcmp( output, result_hex_str->x, ctx2.len, result_hex_str->len ) == 0 );
     }
 
+    /* And now one more time only with N and D */
+    TEST_ASSERT( mbedtls_rsa_export( &ctx2, &N, &P, &Q, &D, &E ) == 0 );
+    TEST_ASSERT( mbedtls_rsa_import( &ctx3, &N, NULL, NULL, &D, NULL ) == 0 );
```
gilles-peskine-arm: The documentation of `mbedtls_rsa_import` states that you must call `mbedtls_rsa_complete` afterwards. I disagree with lifting this restriction.

C being C, it isn't currently enforced at the language level. But we don't guarantee that it will keep working in the next version of Mbed TLS. And please note that the next major version (3.0) is likely to remove these low-level functions from the public interface.

gilles-peskine-arm:
This pull request has been abandoned and the maintener team has no intention of pursuing this feature.

Mbed TLS (which has reintegrated Mbed Crypto) is moving away from exposing the RSA API directly: a future major version will remove `rsa.h` from the public API. Cryptography should go through the PSA API. If you need to do something that the PSA Crypto API does not allow, please raise an issue against [Mbed TLS](https://github.com/Mbed-TLS/mbedtls/issues) or against [the PSA API](https://github.com/ARM-software/psa-api/issues).

PR #392, opened 2020-03-24
Have issue and PR templates redirect to Mbed TLS

mpg:
This is a follow-up to https://github.com/ARMmbed/mbedtls/pull/3085

We don't really want to disable issues and PR completely for this repo (for now), as this would make the existing ones invisible, and there are issues we want to track and PRs that are on their way to getting merged, and I don't think github offers and middle ground of keeping the existing ones but forbidding new ones.

So the second best thing is what this PR does: discourage people to create new issues and PRs here, and encourage them to open them in the mbedtls repo instead.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,44 +1,10 @@
-<!--
+We are no longer accepting new issues in this repository
+========================================================
 
-   ************************************** WARNING **************************************
+Mbed Crypto is now maintained as part of Mbed TLS, see
+https://github.com/ARMmbed/mbedtls/issues/3064
 
-   The ciarcom bot parses this header automatically. Any deviation from the 
-   template may cause the bot to automatically correct this header or may result in a 
-   warning message, requesting updates.
+Please open your issue in the mbedtls repository:
+https://github.com/ARMmbed/mbedtls/issues/new
 
-   Please ensure that nothing follows the Issue request type section, all 
-   issue details are within the Description section and no changes are made to the 
-   template format (as detailed below).
-
-   *************************************************************************************
-
--->
-
-### Description
-
-<!--
-    Required
-    Add detailed description of what you are reporting.
-    Good example: https://os.mbed.com/docs/latest/reference/workflow.html
-    Things to consider sharing:
-    - What target does this relate to?
-    - What toolchain (name + version) are you using?
-    - What tools (name + version - is it mbed-cli, online compiler or IDE) are you using?
-    - What is the SHA of Mbed OS (git log -n1 --oneline)?
-    - Steps to reproduce. (Did you publish code or a test case that exhibits the problem?)
--->
-
-
-### Issue request type
-
-<!--
-    Required
-    Please add only one X to one of the following types. Do not fill multiple types (split the issue otherwise.)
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
+We are no longer accepting new pull requrests in this repository
```
ronald-cron-arm: typo, "requrests" instead of requests

.github/issue_template.md
```diff
@@ -1,44 +1,10 @@
...
-    Required
-    Please add only one X to one of the following types. Do not fill multiple types (split the issue otherwise.)
-    Please note this is not a GitHub task list, indenting the boxes or changing the format to add a '.' or '*' in front
-    of them would change the meaning incorrectly. The only changes to be made are to add a description text under the
-    description heading and to add a 'x' to the correct box.
--->
-    [ ] Question
-    [ ] Enhancement
-    [ ] Bug
+Please open your issue in the mbedtls repository:
```
ronald-cron-arm: "Arm Mbed / mbedtls" instead of just "mbedtls" ?

.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
...
+================================================================
+
+Mbed Crypto is now maintained as part of Mbed TLS, see
+https://github.com/ARMmbed/mbedtls/issues/3064
+
+Please open your pull request in the mbedtls repository:
```
ronald-cron-arm: ditto for "mbedtls"

.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
...
+https://github.com/ARMmbed/mbedtls/pulls
+
+Thanks for your interest in Mbed Crypto!
```
ronald-cron-arm: What about to add that in issue_template.md" as well?

.github/issue_template.md
danh-arm: We don't need to worry about trademark compliance in technical docs. Also, we don't want to mention Arm seeing as we want this to be open governance now. So either "mbedtls" or "Mbed TLS" is fine I think.

.github/pull_request_template.md
mpg: Aw. Thanks for catching.

.github/issue_template.md
mpg: I think since there is a link below it's OK to use the short name of the repo (without the organisation prefix) in the sentence.

.github/pull_request_template.md
mpg: I think it's there too, but for some reason the diff hides it at the end. (I'll never understand why we can't seem to come up with diff tools that generate more sensible diffs.)

.github/issue_template.md
ronald-cron-arm: ok fine by me

.github/pull_request_template.md
ronald-cron-arm: ok no issue then

mpg:
This has two approvals and we don't need to wait for the CI to complete as it's only touching files in the `.github` directory.

PR #393, opened 2020-03-25
Use FindPython3 when cmake version >= 3.12.0

### LINKED ISSUE #389: Don't use FindPythonInterp
### Issue request type

    [ ] Question
    [x] Enhancement
    [ ] Bug

see [https://cmake.org/cmake/help/latest/module/FindPythonInterp.html](https://cmake.org/cmake/help/latest/module/FindPythonInterp.html)
FindPythonInterp may failed when python 2 found first

okhowang:
fix #389

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,18 +47,27 @@ set(CTR_DRBG_128_BIT_KEY_WARNING "${WARNING_BORDER}"
                          "${WARNING_BORDER}")
 
 # Python 3 is only needed here to check for configuration warnings.
-find_package(PythonInterp 3)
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm:
Thanks! But [we've now merged Mbed Crypto back into Mbed TLS](https://github.com/ARMmbed/mbed-crypto/issues/374). We'll still process existing pull requests that only change crypto files, but we can't do this for a pull request that changes the build system because it might need extra changes to handle x509/tls files. [Please raise a new pull request in Mbed TLS](https://github.com/ARMmbed/mbedtls/pulls).

PR #395, opened 2020-04-22
Deprecate Mbed Crypto

gilles-peskine-arm:
Mbed Crypto is no longer being updated, but this is not visible on https://github.com/ARMmbed/mbed-crypto. In this pull request, I update the readme to let people know to use Mbed TLS instead.

To ease the transition, I propose to have a short transition period during which we fix the security issues in Mbed Crypto, and building the library is still possible but shows a warning. After this transition period, we'll make a last patch to cause the library to fail to build.

To verify what security fixes needed to be backported:
```
git log --decorate --name-status upstream-public/pr/3085..mbedtls-2.22.0 -- library ':!library/ssl_*' ':!library/x509*'
```
I saw only one relevant commit.


yanesca:
I have confirmed that all currently available security fixes have been backported, checked spelling and I am happy with the transition period and how it is implemented and documented.

I have only one minor question: In `README.md` we are asking for feedback on the API to be mailed to mbed-crypto@arm.com. Is this address still live and being monitored? We also mention that opening an issue on this repository is an other way to give feedback. Will these issues continue to be monitored in the future?

README.md
```diff
@@ -16,85 +31,4 @@ The design goals of the PSA cryptography API include:
...
-
-Unless specifically indicated otherwise in a file, Mbed TLS files are provided under the [Apache-2.0](https://spdx.org/licenses/Apache-2.0.html) license. See the [LICENSE](LICENSE) file for the full text of this license. Contributors must accept that their contributions are made under both the Apache-2.0 AND [GPL-2.0-or-later](https://spdx.org/licenses/GPL-2.0-or-later.html) licenses. This enables LTS (Long Term Support) branches of the software to be provided under either the Apache-2.0 OR GPL-2.0-or-later licenses.
-
-## Contributing
-
-We gratefully accept bug reports and contributions from the community. Please see the [contributing guidelines](CONTRIBUTING.md) for details on how to do this.
-
-## Feedback welcome
-
 Arm welcomes feedback on the design of the API. If you think something could be improved, please open an issue on our Github repository. Alternatively, if you prefer to provide your feedback privately, please email us at [`mbed-crypto@arm.com`](mailto:mbed-crypto@arm.com). All feedback received by email is treated confidentially.
```
danh-arm, line 34: As @yanesca said, we should not say "open an issue on our Github repository". We should refer to the psa-crypto mailing list. Actually, we have the same issue on development.

The reference mbed-crypto@arm.com is fine though (that's still active)

gilles-peskine-arm:
CI is passing except for the known breakage in Mbed OS.

PR #398, opened 2021-01-18
Update https://armmbed.github.io/mbed-crypto/psa/ to version 1.0.1 + remove cruft

gilles-peskine-arm:
* Remove files that aren't served through GitHub Pages. The sole purpose of the `master` branch is to serve https://armmbed.github.io/mbed-crypto/.
* Update https://armmbed.github.io/mbed-crypto/psa/ to version 1.0.1.
* Remove obsolete text about the old driver interface and mention the new driver interface.


athoelke:
The updates to the PSA specification look sound - (it has built the same version of the document as the published spec). I'm happy with the updates to the landing page (index.md) - no need to mention the API status as it is now Final/Release quality.

bensze01 (approved):
Looks good to me. The only change I didn't see explicitly mentioned, was that I think removing the source files disables the "show source" button on the html docs - but that is a very niche functionality I don't think we mind disabling.

gilles-peskine-arm:
Thanks for the review! I wasn't even aware of “show source”, and I don't see why we would publish some rst files.

PR #401, opened 2022-04-07
Update PSA API document to version 1.1

gilles-peskine-arm:
Update the copy of the PSA Cryptography API (HTML and PDF) served at https://armmbed.github.io/mbed-crypto/psa to version 1.1.0 (released a few days ago).

The previous version is still available and referenced from the `/psa` page. All versions will now be available in a versioned URL.

I put up a live copy on my fork at https://gilles-peskine-arm.github.io/mbed-crypto/psa/ if you want to see the rendering.


bensze01 (approved):
Looks good to me.

Non blocker: Should we maybe update the URLs to point to the url for the Mbed TLS repo? (It'll work / get redirected as is as well.)

gilles-peskine-arm:
I've re-uploaded with signoff lines in the commit messagex.

Let's not scope creep unrelated changes into this. We'll keep the ARMmbed redirection for a while, there's no urgency.

bensze01 (approved):
Sure, that makes sense. Approved.

minosgalanakis (approved):
LGTM

PR #402, opened 2022-04-12
Pages api add pake

gilles-peskine-arm:
Update the copy of the PSA Cryptography API (HTML and PDF) served at https://armmbed.github.io/mbed-crypto/psa to add the PAKE extension. I forgot this in #401.

I put up a live copy on my fork at https://gilles-peskine-arm.github.io/mbed-crypto/psa/ if you want to see the rendering.


minosgalanakis (approved):
LGTM

